PR #95, opened 2015-01-10
make mupdf work for multiple instances of Vim with same main TeX file

Konfekt:
For explanations see https://github.com/lervag/vim-latex/issues/94


lervag:
Thanks!

PR #99, opened 2015-01-18
Update README.md

thor:
A small, small edit to update the commands to be in line with Vundle.vims documentation, due to how the *Bundle*-namespace is deprecated.


lervag:
Thanks!

PR #102, opened 2015-01-21
use v:progname instead of progpath because exists for all Vim version > 6

Konfekt:
Closes https://github.com/lervag/vim-latex/issues/101


lervag:
Thanks. I am considering to test for the correct vim patch and use `v:progpath` if it is available. In the meantime this seems like a good enough solution if it works for everyone.

Issue #105, opened 2015-01-29
Not compiling on Windows with `shellslash`


salviati6:
Environment: Windows 7 SP1 64-bit, VIM 7.4.589 64-bit
Symptom: On Windows with `shellslash` set, compiling with vim-latex (`\lv`) reports success but nothing is actually done.

This is much like Issue #40, unfortunately the fix to that issue does not seem to work on this one.

From `:h shellescape()`:

> On MS-Windows and MS-DOS, when 'shellslash' is NOT set, it will enclose {string} in double quotes and double all double quotes within {string}.

On Windows with `shellslash` set, `shellescape()` will enclose its returned string in single quotes to produce a `cmd` like

> cmd /s /c "cd /D 'C:\PATH' ...

which cannot be recognized by `cmd` and consequent operations are automatically aborted (the compilation never actually starts in this case).

Maybe a change to `autoload\latex\latexmk.vim:89` is necessary, or just inform Windows users to turn off shellslash in the docs. However, some plugins (e.g., `vim-startify`) are not functioning without `shellslash`.

lervag:
Interesting. As I think I mention in the docs, I am not a windows user myself and I therefore rely on windows users to help me fix problems such as these.

Do I understand correctly that the issue may be "fixed" by turning of `shellslash`? One possibility would then be to turn off shellslash temporarily in the functions where necessary. However, I am not really sure I want that, because I want the code to be as clean as possible.

salviati6:
Thanks for the prompt reply!

I work cross-platform on OS X and Windows, and I wish to keep `shellslash` on Windows because it gives me a consistent environment (always forward slashes in paths). I totally agree with the idea to keep the code succinct, and I think only minor changes are necessary. I'm new to VIM and I'll try to fix the problem.

lervag:
I understand. I will be very happy to consider pull requests!

Let me know if I can assist somehow. A couple of tips: I've created some utility functions that are mapped to `<localleader>li` and `<localleader>lo` that may be helpful in debugging. The first shows the important information that is stored by vim-latex, and the latter shows that output from the `latexmk` command itself.

salviati6:
The helper functions are extremely useful! I like the way this plugin is designed, and I'm currently working on my LaTeX docs solely with it.

I am looking into the code and I believe a small hack to `shellescape()` will be enough. I don't want to alter `shellslash` even temporarily because it's global any may affect other plugins.

lervag:
[...]

PR #106, opened 2015-01-30
Fix for issue #105

salviati6:
This fix is expected to resolve Issue #105 in which `latex#latexmk#compile()` gives a success message but actually does nothing on Windows with `shellslash` set. The cause of the problem is that `shellescape()` on Windows returns a string enclosed by single quotes with `shellslash`, while Windows `cmd` recognizes only double quotes.

The fix should be in effect for `vim-latex` on Windows only, except a small typo fix as a bonus. :)

Tested on Windows 7 SP1 64-bit, VIM 7.4.589 64-bit w/ and w/o `shellslash`.

### DIFF
--- a/autoload/latex/latexmk.vim
+++ b/autoload/latex/latexmk.vim
@@ -110,7 +110,7 @@ function! latex#latexmk#clean(full) " {{{1
   " Run latexmk clean process
   "
   if has('win32')
-    let cmd = 'cd /D ' . shellescape(data.root) . ' & '
+	let cmd = 'cd /D "' . data.root . '" & '
   else
     let cmd = 'cd ' . shellescape(data.root) . '; '
   endif
@@ -120,7 +120,7 @@ function! latex#latexmk#clean(full) " {{{1
   else
     let cmd .= ' -c '
   endif
-  let cmd .= shellescape(data.base)
+  let cmd .= latex#util#fnameescape(data.base)
   let g:latex#data[b:latex.id].cmds.clean = cmd
   let exe = {
         \ 'cmd' : cmd,
@@ -331,7 +331,7 @@ function! s:latexmk_build_cmd(data) " {{{1
   let tmp = tempname()
 
   if has('win32')
-    let cmd  = 'cd /D ' . shellescape(a:data.root)
+    let cmd  = 'cd /D "' . a:data.root . '"'
     let cmd .= ' && set max_print_line=2000 & latexmk'
   else
     let cmd  = 'cd ' . shellescape(a:data.root)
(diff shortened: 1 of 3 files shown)


autoload/latex/latexmk.vim
```diff
@@ -110,7 +110,7 @@ function! latex#latexmk#clean(full) " {{{1
   " Run latexmk clean process
   "
   if has('win32')
-    let cmd = 'cd /D ' . shellescape(data.root) . ' & '
+	let cmd = 'cd /D "' . data.root . '" & '
```
salviati6: Sorry for the 3 inconsistent indents in my commits...

salviati6:
Sorry for the 3 inconsistent indents in my commits...

lervag:
First, thanks for investigating and proposing a fix! I have a lot on my platter these days, and there is not that much time to work on vim-latex. I have some comments to your commits, and I hope that you could make some minor changes. Note that you can fix the inconsistent indents on your end. When you push new changes to your own version, this pull request will be automatically updated.

autoload/latex/latexmk.vim
lervag: In addition to the inconsistent indent: Is it safe to remove the `shellescape` here? What happens if `data.root` has whitespaces? Is the added quotes enough?

lervag:
Oh, and one more thing: Could you please check that your change will work on windows both with `shellslash` and `noshellslash`?

salviati6:
For some Windows commands (like `cd` in this case), a path with space may be supplied as a parameter only if it is surrounded by double quotes. Single quotes are not recognized and no escape is required inside the path. For example, to enter a directory `hello world`, one should use
```
cd "hello world"
```
The following commands are illegal and will fail with the error "The system can not find the path specified":
```
cd 'hello world'
cd "hello\ world"
cd 'hello\ world'
```

In my fix, for `shellescape()`s that already reside in a conditional block for `win32`, I just enclosed the target path with double quotes manually like `'"' . path . '"'`. There are hardcoded double quotes in the modified `autoload/latex/latexmk.vim:113` and `autoload/latex/latexmk.vim:334`, but it's easy to miss them. Other `shellescape()`s should be valid on both Windows and *NIX, and they have been replaced by a new util function `latex#util#fnameescape()`.

I have tested my fix on Windows 7 SP1 64-bit, VIM 7.4.589 64-bit w/ both `shellslash` and `noshellslash` (default). I didn't perform a test on *NIX as the fix should be in effect only for Windows.

lervag:
Great, thanks!

lervag:
I've fixed the indentations myself.

salviati6:
Sorry for the inconvenience. Next time I'll surely remember to follow the coding conventions. :)

lervag:
No problem! Thanks for contributing!

PR #120, opened 2015-02-18
document necessity of --servername in Unix terminal Vim

Konfekt:


lervag:
Thanks!

lervag:
I hope you don't mind that I updated the text: I removed the part about how to create the aliases or scripts, and instead referred to the mentioned website. I think the important part was simply that one needs the `--servername` option.

Konfekt:
Ok, I wanted to make the point to someone uninformed that she preferably chooses the aliasing option or option (B), and why so. 

The website itself leaves the reader at loss and option (C) is rather harmful.

PR #127, opened 2015-03-02
Return an empty string in callback function

micbou:
On Windows, when executing the command `gvim --servername servername --remote-expr "latex#latexmk#callback(1)"` in a terminal, gvim opens the following window:

callback-window

By returning an empty string in `latex#latexmk#callback` function, this window disappears.


lervag:
Thanks!

PR #133, opened 2015-03-12
Add autoref syntax and completion

zegervdv:
Adds support for `hyperref`'s `\autoref`with completion like `\ref`.


lervag:
Thanks!

PR #144, opened 2015-04-03
silence mappings and stiffen operator mappings  

Konfekt:


lervag:
The `<silent>` is obvious, but could you explain how the operator maps become more robust through these changes? Do I understand correctly that this prevents conflicts with custom remappings of for instance `v`?

Konfekt:
More robust because it makes the operator mappings work even if not 

- `v` enters character-wise or block-wise visual mode, and
- the visual mode map was mapped to the default key sequence

For example, `omap ie` is `:normal vie<cr>` and thus requires that `v` enters character-wise visual mode and that `ie` selects the inner environment text object.

lervag:
Thanks! I appreciate the contribution!

PR #146, opened 2015-04-06
Fix util#execute on Windows

micbou:
When executing a command, `!` and `cmd` should be concatenated to avoid a space between them. This is critical on Windows because the command `! start` has not the same behavior as `!start`.

### DIFF
--- a/autoload/vimtex/util.vim
+++ b/autoload/vimtex/util.vim
@@ -154,9 +154,9 @@ function! vimtex#util#execute(exe) " {{{1
     endif
   else
     if silent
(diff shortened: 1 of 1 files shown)


lervag:
I'm accepting, but I just want to make sure I understand: Does this mean that on windows, preceding spaces matter when you execute a command in the shell (dos shell or whatever)?

micbou:
Only with the keyword `start`. It is explained in `:help !start`.

lervag:
I see, thanks!

PR #147, opened 2015-04-06
Youcompleteme

micbou:
`YouCompleteMe` always expects a return value from the `omnifunc` function hence the first commit.

The second commit is about the documentation. There are several mistakes. 
Firstly, it is not necessary to change the option `g:ycm_filetype_whitelist` because `YouCompleteMe` is enabled for all files by default. 
Secondly, identifier-based completion can be useful for some users so the `g:ycm_min_num_of_chars_for_completion` option should not be modified. Moreover, this option is global.
Finally, the pattern for the semantic completion must be written in `Python` regex, not `Vim`. The part to be completed (the labels and the citations) must not be present in the pattern.

With these changes, the automatic completion of `YouCompleteMe` works relatively well. There are some problems with labels and citations containing special characters (`:`for example) or spaces. The completion menu disappears when typing these characters and must be reopened with the `<C-space>` command.

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -65,6 +65,7 @@ function! vimtex#complete#omnifunc(findstart, base) " {{{1
         return -2
       endif
     endfor
+    return -3
   else
     "
     " Second call:  Find list of matches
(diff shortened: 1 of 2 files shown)


lervag:
Thanks!

PR #152, opened 2015-04-14
Add a space after the '-c' flag in latexmk#clean.

simichaels:
Previously the lack of a space after '-c' would cause VimtexClean
to silently fail when not performing a full clean.


lervag:
Thanks! Nice catcht!

PR #162, opened 2015-05-05
Fix view command when output directory is an absolute path

kyessenov:
The original code is buggy (e.g. <Leader>lv shows "can not read PDF file" error message).
It appears the candidate file has the form  .//tmp/latex if my output_directory is set to /tmp/latex.


lervag:
Thanks for the pull request. I've made some updates to how the output directory option is considered in the code, so that it should better handle both absolute and relative paths. When the user does not set this option, the output directory should now correctly never be relevant.

PR #165, opened 2015-05-07
Fix typo in documentation

tommcdo:
A couple places in the documentation mention a `csd` mapping, which I think should actually be `csc`.


lervag:
Thanks! Nice catch! I chose to fix it slightly differently in order to be consistent with the surrounding text.

tommcdo:
Considered that, but didn't want to assume there wasn't a good reason for it already :)

lervag:
Nah, I was probably just too sloppy... :p

Issue #191, opened 2015-07-01
Modeline foldmethod option overwritten


cledoux:
When writing latex documents, I keep my common preamble in a header.tex file that is included in the main latex file.

I prefer to use manual folding for this header document as it is essentially a config file and I want full control over what sections of the config are folded together at a time.

Thus, I utilize the following modeline in my header.tex:
`% vim: set foldtext=foldtext() foldmarker={{{,}}} foldmethod=marker foldlevel=0 spell:`

The issue is that when I open header.tex, foldmethod is set to manual instead of marker. Manually setting foldmethod=marker brings back my desired folding behavior.

lervag:
Interesting. Do you want to have the vimtex folding feature enabled in other tex files? If not, then you can simply `:let g:vimtex_fold_enabled = 0`.

If you do want the fold feature enabled, but disabled in this particular case, then we have an interesting issue. I do not see directly how to solve it, but I can try to find if there is any way to detect if a setting comes from modeline or not, and let modeline settings override vimtex.

cledoux:
I want vimtex folding on by default with the ability to disable it on a per file basis.

I tried setting the referenced variable in modeline via let-modeline plugin, but this didn't seem to work. I'm guessing this is because vimtex is loaded before the modeline variables, but I haven't verified.

cledoux:
I just noticed that the foldtext option that's set via modeline is not overwritten by vimtex.  Without putting the foldtext into the modeline, foldtext is `vimtex#fold#text()`. With the modeline setting, the value is as expected.

It's only the foldmethod that isn't set as expected.

cledoux:
Another piece of debugging information:

This is only the case when initially opening the file. Reloading the buffer sets the values as expected.

In other words, if from the command line, I type `vim header.tex` foldmethod will be set to `manual`. However, typing `:edit` from vim once open will reset foldmethod to `marker` as desired.

lervag:
Ah, I see. Then the settings are actually parsed correctly. What is happening is that vimtex by default uses a trick where the manual and expression foldmethods are mixed (to speed things up). An autocommand is run after the cursor is moved in order to ensure that folds are updated and the manual foldmethod is entered.

Could you please open the header file again and immediately check what the foldmethod is. IMPORTANT: Don't move the cursor before you do this. Could you also please try to comment out/delete the lines 72 to 77 in `VIMTEX-PATH/autoload/vimtex/fold.vim` and see if this resolves the issue?

[...]

PR #192, opened 2015-07-01
Edited fix used for bug #191

cledoux:
The previous method used to determine if the foldmethod option was set from
a modeline wasn't working for me.

The function vimtex#fold#s:check_modeline was always returning false.

I changed the function to use a simple regex search on the actual file
itself searching for a modeline in the proper line ranges.


lervag:
Thanks, this looks good.

PR #194, opened 2015-07-02
Fix compilation when Vim path contains spaces

micbou:
On Windows, `v:progpath` may contain spaces and so `g:vimtex_latexmk_progname` too. If this is the case, `$success_cmd` and `$failure_cmd` latexmk options will not work properly. This is fixed by surrounding `g:vimtex_latexmk_progname` with double quotes in these options.


lervag:
Thanks. It seems this works as expected also on Linux, and I guess it should work also on MacOS. If not, then I guess we'll know in not too long... :)

PR #195, opened 2015-07-02
Fix general viewer when paths contain backslashes

micbou:
Backslashes in paths must be escaped to be correctly parsed by the substitute function. This fixes the command `:VimtexView` with SumatraPDF on Windows.

### DIFF
--- a/autoload/vimtex/util.vim
+++ b/autoload/vimtex/util.vim
@@ -121,14 +121,18 @@ endfunction
 
 " }}}1
 function! vimtex#util#fnameescape(path) " {{{1
+  " Blackslashes in path must be escaped to be correctly parsed by the
+  " substitute() function.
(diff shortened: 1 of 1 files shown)


lervag:
Hi. The point of `vimtex#util#fnameescape` is to do what is needed to escape file names properly. I'm wondering if it is not a better solution to instead do the escaping you want inside this function. That is, add the `escape(...)` in `fnameescape` in `util.vim` line 131. Since I don't use Windows, I can't test this. Could you please test if that works and verify that it does not break anything else?

micbou:
I tested it and it is working. I updated the PR with the new approach. Escaping is done for all platforms since Unix paths may also contain backslashes.

lervag:
Thanks!

PR #196, opened 2015-07-04
escape path in MuPDF reverse search function

lahwaacz:
Fixed just a small bug I found when the path contained spaces...

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -257,7 +257,7 @@ function! s:mupdf.reverse_search() dict " {{{2
   let self.line = system(self.cmd_getline)
 
   " Go to file and line
-  silent exec 'edit ' . self.file
+  silent exec 'edit ' . fnameescape(self.file)
   if self.line > 0
     silent exec ':' . self.line
     " Unfold, move to top line to correspond to top pdf line, and go to end of


lervag:
Thanks! I didn't accept the pull request because I wanted to use the vimtex utility function `vimtex#util#fnameescape` instead, in order to also support windows.

lahwaacz:
In that case, although I don't fully understand the difference, `fnameescape` is still used at multiple places. No idea if `vimtex#util#fnameescape` should be used everywhere or not...
```
$ grep -F ". fnameescape(" **/*.vim
autoload/vimtex/complete.vim:  execute 'lcd ' . fnameescape(b:vimtex.root)
autoload/vimtex/complete.vim:  execute 'lcd ' . fnameescape(l:save_pwd)
autoload/vimtex/index.vim:    execute 'badd ' . fnameescape(entry.file)
autoload/vimtex/latexmk.vim:    execute 'cfile ' . fnameescape(log)
autoload/vimtex/latexmk.vim:    execute 'cgetfile ' . fnameescape(log)
autoload/vimtex/util.vim:    execute 'lcd ' . fnameescape(a:exe.wd)
autoload/vimtex/util.vim:    execute 'lcd ' . fnameescape(pwd)
```

micbou:
No, `vimtex#util#fnameescape` cannot be used for Vim commands. Commit a5531c7 breaks things for MuPDF users. The issue is that paths in Vim command cannot be escaped the same way as paths in the shell.

I think the function `vimtex#util#fnameescape` should be renamed to `vimtex#util#shellescape` to avoid confusion with `fnameescape` and should only be used for shell command. Always use `fnameescape` for Vim commands.

lervag:
@micbou You are right! I don't know why I've mixed these functions (i.e. `fnameescape` and `shellescape`), but that is what I've done.

I'm updating as you recommend, changing `...fnameescape` to `...shellescape`.

If you find examples where I've mixed the `shellescape` with `fnameescape`, please let me know. I've looked through the code, but I couldn't find any more errors. I wouldn't be surprised if I've missed anything, though.

micbou:
I didn't find any misuse of `vimtex#util#shellescape` and `fnameescape` but I found 3 `shellescape` that could certainly be replaced by `vimtex#util#shellescape`.

lervag:
Thanks!

PR #202, opened 2015-07-13
No console window when executing vimtex#latexmk#clean and vimtex#latexmk#kill on Windows

micbou:
This PR sets the background option for the `vimtex#latexmk#clean` and `vimtex#latexmk#kill` functions to prevent the opening of a console window on Windows.

I think this change is safe on Unix platforms.


micbou:
Those lines could also be removed since this is the default in `vimtex#latexmk#execute`.

lervag:
I think the reason I've chosen to keep these in the foreground was that they should always be fast. But I don't really see a good reason why it can't stay in the background, so I'll accept your pull request. That is, you are right that the lines can be removed, so I'll rather push the change myself where I remove and simplify the code.

In any case, thanks!

PR #210, opened 2015-08-07
Do not treat '(' as delimiter for indentation

andrewmwhite:
Can't think of a good reason for an open-paren to increase indentation level in LaTeX sources, but it does get in the way of typing text with parenthetical expressions when hard-wrapping.

### DIFF
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -138,7 +138,6 @@ let s:envs_lists = 'itemize\|description\|enumerate\|thebibliography'
 let s:envs_noindent = 'document\|verbatim\|lstlisting'
 let s:delimiters_open = '\(' . join([
         \ '{',
-        \ '(',
         \ '\[',
         \ '\\{',
         \ '\\(',
@@ -149,7 +148,6 @@ let s:delimiters_open = '\(' . join([
       \ ], '\|') . '\)'
 let s:delimiters_close = '\(' . join([
         \ '}',
-        \ ')',
         \ '\]',
         \ '\\}',
         \ '\\)',


lervag:
Hi @andrewmwhite, and thanks for the suggestion! However, I do not want to merge this directly. Instead, I think I will implement the open/close delimiters as an option.

Note that I understand that you think the open-paren should not increase the indentation level. Personally I find it quite nice, but I agree it is unusual. Thus as a compromise, I will remove the open-parens from the default setting, and instead add them personally as a custom setting.

I will make the implementation as soon as possible, but unfortunately I don't have that much time right now, so it might take some time. Feel free to adjust the pull request if you feel up for the task. Note, in that case, that this requires an explanation in the docs.

andrewmwhite:
Fair enough -- I thought that might end up being the case. If I have
time, I'll update the pull request to make it an option, but I'm also
pretty busy right now.

Out of curiosity, in what scenario do you find it nice to have those
delimiters trigger an indent? I couldn't think of a use-case. For me, it
gets in the way of the "flow" of writing when I have a sentence with a
parenthetical expression (since I use hard line wraps).

Karl Yngve Lervåg wrote:

lervag:
I also use hard wraps (although I am reconsidering that in lieu of collaborative projects where a sentence per line makes sense for version control). I personally like the indentation of paranthesized sentences, simply because it makes me aware of the parantheses. I understand that this is probably strange to most people. But the most important use case is mathematical expressions, where I like that expressions in parantheses are indented if the expressions don't fit on one line, e.g.

```tex
\begin{equation}
  f(x) = a_1(b_1 + b_2 + b_3 + b_4
    + b_5 + b_6)
\end{equation}
```

I think the latter case should be a more common preference.

One alternative would be to only add the indentations on parantheses in math mode. What do you think about that? Personally I could live with a compromise where I only indent on (most) delimiters, including parens, inside math mode. And in this case to not introduce more options.

andrewmwhite:
I can see where that would be useful in math mode. I think that
compromise makes sense, though I'm guessing the implementation would be
rather more onerous.

Karl Yngve Lervåg wrote:

lervag:
I think it is doable. I'll look into it and see if I can solve the issue in not too long. In the meantime, I hope you can live with the current state (or your personal fork).

lervag:
It seems to work as expected on my end. Please verify that this solves your issue.

andrewmwhite:
Looks good to me. Thanks!

Karl Yngve Lervåg wrote:

PR #213, opened 2015-08-11
Fixes #212. Fix pdflatex substitution to work with Latexmk's 'internal' keyword

### LINKED ISSUE #212: Latexmk `internal` keyword (sh: 1: internal not found)
I have the following in my `.latexmkrc`:

    pdflatex = 'internal musixlatex %R %O %S';
    push @generated_exts,"mx2";
    push @generated_exts,"mx1";
 
    sub musixlatex {
       my $root = shift;
       # Define root filename for the mx1 and mx2 files
       # Note that $aux_dir1 includes directory separator
       #      and it equals $out_dir1 if only the out_dir is set.
       my $mx = "$aux_dir1$root";
       my @args = @_; 
       if ( $pass{$rule} <= 1 ) { 
          print "======= FIRST PASS: Deleting '$mx.mx2' =======\n";
          unlink "$mx.mx2"; 
          print "====   Generating '$mx.mx1' then '$mx.mx2' ===\n";
          my $ret = system( "pdflatex @args" );
          if ($ret) { return $ret; }
          $ret = system( "musixflx $mx" );
          if ($ret) { return $ret; }
       }
       return system( "pdflatex @args" );
    }

Running `latexmk -pdf ...` from the command line works perfectly. However, `VimtexCompile` gives me the following error:

    Latexmk: Need to remake files.
      File changes, etc:
       Changed files, or newly in use since previous run(s):
          'songbook.tex'
    Rule 'pdflatex': File changes, etc:
       Changed files, or newly in use since previous run(s):
          'songbook.tex'
    ------------
    ------------
    ------------
    Running 'internal -file-line-error musixlatex "songbook"  -recorder  "songbook.tex"'
    ------------
    Latexmk: All targets (songbook.pdf) are up-to-date

    === Watching for updated files. Use ctrl/C to stop ... 
    Latexmk: applying rule 'pdflatex'...
    sh: 1: internal: not found
    Latexmk: fls file doesn't appear to have been made
    Latexmk: Log file says output to 'songbook.pdf'
    Latexmk: Failure to make the files correctly
        ==> You will need to change a source file before I do another run <== 
    Collected error summary (may duplicate other messages):
      pdflatex: Command for 'pdflatex' gave return code 32512

I didn't change any of VimTeX's default options. What could be causing Latexmk to fail when run by VimTeX?

camilstaps:
This regex substitution won't touch the `pdflatex` command if `internal` is at the beginning of the command. This is needed for Latexmk subroutines.

Testing:

    pdflatex somefile           --> pdflatex -file-line-errors somefile
    pdflatex internal somefile  --> pdflatex -file-line-errors internal somefile
    internal routine file       --> internal routine file
    pdfinternal somefile        --> pdfinternal -file-line-errors somefile
    internalpdf somefile        --> internalpdf -file-line-errors somefile
    pdfinternalpdf somefile     --> pdfinternalpdf -file-line-errors somefile

I'm not particularly a regex wizard, so this may be possible in an easier way. Unfortunately, Perl doesn't support variable length lookbehind.

### DIFF
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -383,8 +383,8 @@ function! s:latexmk_build_cmd() " {{{1
   endif
 
   let cmd .= ' ' . g:vimtex_latexmk_options
-  let cmd .= ' -e ' . shellescape(
-        \ '$pdflatex =~ s/^((.(?<!^internal(?=\s)))*?) /$1 -file-line-error /', 1)
+  let cmd .= ' -e ' . vimtex#util#shellescape(
+        \ '$pdflatex =~ s/^((.(?<!^internal(?=\s)))*?) /$1 -file-line-error /')
   if g:vimtex_latexmk_build_dir !=# ''
     let cmd .= ' -outdir=' . g:vimtex_latexmk_build_dir
   endif
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -384,7 +384,7 @@ function! s:latexmk_build_cmd() " {{{1
 
   let cmd .= ' ' . g:vimtex_latexmk_options
   let cmd .= ' -e ' . vimtex#util#shellescape(
-        \ '$pdflatex =~ s/ / -file-line-error /')
+        \ '$pdflatex =~ s/^((.(?<!^internal(?=\s)))*?) /$1 -file-line-error /')
   if g:vimtex_latexmk_build_dir !=# ''
     let cmd .= ' -outdir=' . g:vimtex_latexmk_build_dir
   endif
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -383,8 +383,8 @@ function! s:latexmk_build_cmd() " {{{1
   endif
 
   let cmd .= ' ' . g:vimtex_latexmk_options
-  let cmd .= ' -e ' . vimtex#util#shellescape(
-        \ '$pdflatex =~ s/^((.(?<!^internal(?=\s)))*?) /$1 -file-line-error /')
+  let cmd .= ' -e ' . shellescape(
+        \ '$pdflatex =~ s/^((.(?<!^internal(?=\s)))*?) /$1 -file-line-error /', 1)
   if g:vimtex_latexmk_build_dir !=# ''
     let cmd .= ' -outdir=' . g:vimtex_latexmk_build_dir
   endif
--- a/autoload/vimtex/util.vim
+++ b/autoload/vimtex/util.vim
@@ -123,7 +123,7 @@ endfunction
 function! vimtex#util#shellescape(path) " {{{1
   " Blackslashes in path must be escaped to be correctly parsed by the
   " substitute() function.
-  let l:path = escape(a:path, '\')
+  let l:path = has('win32') ? escape(a:path, '\') : a:path
 
   "
   " In a Windows environment, a path used in "cmd" only needs to be enclosed by
@@ -132,7 +132,7 @@ function! vimtex#util#shellescape(path) " {{{1
   " reports an error.  Any path that goes into vimtex#util#execute() should be
   " processed through this function.
   "
-  return has('win32') ? '"' . l:path . '"' : shellescape(l:path)
+  return has('win32') ? '"' . l:path . '"' : shellescape(l:path, 1)
 endfunction
 
 " }}}1


camilstaps:
I needed to escape the exclamation mark in the regex, and didn't get it working with your `shellescape` function. That's why I changed it to the built-in.

lervag:
Again, sorry for taking so long to answer!

I would be very happy if you had a minimal example for a case where this is necessary. Then I could create a simple test with a `.latexmkrc` file and a `tex` file for future refence and in case of future issues.

I would also want to understand why the `vimtex#util#shellescape` does not work. Are you on Windows? If not, then I think the problem might be [this part](https://github.com/lervag/vimtex/blob/master/autoload/vimtex/util.vim#L126). Could you verify this by commenting that line out?

Further, I am wondering: If you use `internal`, does the command still accept options like `-file-line-error`?

camilstaps:
No problem at all.

This was for a somewhat bigger project using MusiXTeX which needs a workflow of

* deleting some files (*.mx1, *.mx2)
* running pdflatex
* running musixflx
* running pdflatex again

This is not easily done in latexmk without `internal`.

Long story short, I cannot provide that project as minimal example due to it not being minimal for one, but I will make something in the coming days, and then I will also do the tests you suggested (I'm on Debian 8.1). My knowledge of vimscript is poor so I may have overlooked some issues with Windows.

Anyway, more soon!

camilstaps:
I have just merged your latest commits into my fork. The following is done using that version.

Here is an minimal example, the TeX:

    \documentclass[11pt,a4paper]{article}
    
    \begin{document}
    
    This is just a simple example. Compiling it with the following \texttt{.latexmkrc} should print ``This line should be printed'' before starting \texttt{pdflatex}.
    
    \begin{verbatim}
    $pdflatex = 'internal internalexample %R %O %S';
    
    sub internalexample {
        my @args = @_;
        print "This line should be printed\n";
        return system("pdflatex @args");
    }
    \end{verbatim}
    
    \end{document}

The .latexmkrc is given in the verbatim block.

Compiling this with `:VimtexCompile`, `:VimtexCompileOutput` gives me:

    Rule 'pdflatex': Rules & subrules not known to be previously run:
       pdflatex
          'pdflatex'
    ------------
    ------------
    ------------
    Running 'internal internalexample "vimtexinternal"  -recorder  "vimtexinternal.tex"'
    ------------
    Latexmk: calling internalexample( vimtexinternal -recorder vimtexinternal.tex )
    Latexmk: applying rule 'pdflatex'...
    This line should be printed
    This is pdfTeX, Version 3.14159265-2.6-1.40.15 (TeX Live 2015/dev/Debian) (preloaded format=pdflatex)

The beforelast line in the output shows that the `internal` keyword has been processed correctly.

Now, this is not an example where this is *necessary*. It is merely an example to test with. An example where it is necessary (the `internal` keyword, that is) could be use with MusiXTeX. This package uses a three-pass system (see http://texdoc.net/texmf-dist/doc/generic/musixtex/musixdoc.pdf#page=13, section 1.3) which suggests the workflow:

> Delete `jobname.mx2` &rarr; TeX &rarr; `musixflx` &rarr; TeX

This workflow can be accomplished with the .latexmkrc found [here](http://tex.stackexchange.com/a/260766/23992).

Would you like me to create a minimal example with MusiXTeX?

---

You were right that the problem was escaping the backslashes. In 41f0eccdc0b9f3cb80e5246dfe07f646d3aa0768, I changed this to do that only on Windows (I gather that was your idea?). I don't have a Windows machine here so I can't test it, unfortunately. 

However, I also had to change [this](https://github.com/lervag/vimtex/blob/master/autoload/vimtex/util.vim#L135) line to

    return has('win32') ? '"' . l:path . '"' : shellescape(l:path, 1)

i.e. add `1` as second argument, it *does* work, also with line 126 uncommented. The second argument `1` is needed [to escape special characters](http://stackoverflow.com/a/21225044/1544337), in this case `!` (and maybe others?). The [regex I wrote](https://github.com/camilstaps/vimtex/blob/master/autoload/vimtex/latexmk.vim#L387) contains a `!` for negative lookbehind.

I don't see any side effects of this yet - but you can better tell if this could cause unwanted behaviour. Another option would be to add this second argument also as an argument to `vimtex#util#shellescape`. 

---

Things like `-file-line-errors` work properly as long as you don't use `internal`. See the top of this pull request for some tests with the regex I wrote. As you can see, command line arguments are passed correctly to every test case except when using `internal`. I did this, because the `internal` subroutine may or may not pass its options to `pdflatex`. The example I gave above,

    sub internalexample {
        my @args = @_;
        print "This line should be printed\n";
        return system("pdflatex @args");
    }

simply calls `pdflatex` with the arguments that were passed to the subroutine. In this case, it would be useful to substitute `internal internalexample` with `internal internalexample -file-line-errors`, which would then mean a system call `pdflatex -file-line-errors ...`. But, this subroutine does not *have* to pass its options to `pdflatex`. It may also do something else with them.

Because the subroutine would typically be written by the end user, I wouldn't consider it a good idea to pass `-file-line-errors` to it. After all, that is a `pdflatex` option, not an option of the subroutine. The user may add it to his subroutine (simply calling `system("pdflatex -file-line-errors ...")`) if he wishes to do so. 

If you really want to pass `-file-line-errors` though, you could use this regex:

    $pdflatex =~ s/(^internal\s\w+?)?( |$)/$1 -file-line-errors$2/;

This gives the test cases:

    pdflatex somefile           --> pdflatex -file-line-errors somefile
    pdflatex internal somefile  --> pdflatex -file-line-errors internal somefile
    internal routine file       --> internal routine -file-line-errors file       <- different
    pdfinternal somefile        --> pdfinternal -file-line-errors somefile
    internalpdf somefile        --> internalpdf -file-line-errors somefile
    pdfinternalpdf somefile     --> pdfinternalpdf -file-line-errors somefile

lervag:
Hi! Sorry for taking so long!

I think the latest version looks good. The shellescape-stuff is probably something I should consider more carefully and redo, once I get the time.

I think the use of `internal` is quite rare, and in those few cases, I think the user should be responsible for ensuring that he/she is using the necessary options, e.g. `-file-line-error`. Thus I think the current version should be suitable.

camilstaps:
That's great, thank you!

lervag:
No problem :)

lervag:
Hi @camilstaps. It seems this work broke vimtex on windows. I was not able to find a good fix, so I reverted the pull request. Instead, I added an option `g:vimtex_latexmk_file_line_error` that you can turn off. If it is off, then I think things should work well, even without the regexp. Could you please test?

Else we must find a way to fix this that also works on windows.

camilstaps:
@lervag oh, I'm so sorry! 

This does work for me, but with a small disadvantage:

The regex solution (at least on linux) worked both with and without the internal keyword: by editing the .latexmkrc file you could choose yourself whether to use such a subroutine or not. *If not, the -file-line-errors options was still passed to pdflatex.*

With this change, if you want to use internal in one project, you have to set the global variable and not have -file-line-errors in all other projects. You could set the variable depending on what file you're editing, but it won't be as clean a solution as simply editing the .latexmkrc you need anyway.

I understand that this is less important than Windows support. Unfortunately, I have rather limited access to Windows machines, but if I can find one and some time I might have a look as to why it breaks vimtex.

Once again apologies for breaking vimtex!

lervag:
No problem, @camilstaps. I hope we might find a better solution that also works on windows, but I don't have the time to investigate now. If you every do get the time to work with a windows machine to find a working solution, I would be happy to reconsider a new pull request!

PR #240, opened 2015-10-05
Avoid clearing global listchars option

tokoro10g:
`listchars` is a global option, so it cannot be set locally to a certain buffer.


lervag:
Agreed! Thanks!

Issue #245, opened 2015-10-08
change surrounding command confused in footnotes


thomasahle:
If I place my cursor on `for` in the following:

    \footnote{We use $\oplus$ for concattenation}

And write `csc`, I get the option of changing `oplus` into something else.

I would prefer to get the option of changing `footnote`, as I don't really consider myself being 'inside' the `oplus`.

Konfekt:
Confirmed, that behavior of `csc` is inconsistent with that of of `dsc` that deletes `\footnote{...}`

Konfekt:
The 
```vim
normal! F\lcea
```
command is too simplistic.

Konfekt:
So how about `normal! F{F\lcea`?

Konfekt:
Better `vaB<esc>` instead of F{ as this skips escaped braces, as does dsc.

lervag:
I've pushed a fix to this. I updated both `csc` and `dsc` so they should both be more general.

I think there might be a problem with `dsc` in some cases. It would be nice to get a report whether you agree, or to hear if it works as expected.

Konfekt:
Hello, 


True that the previous commands worked fine but very invasive, as jump list and previous visual selection were overridden.

Here `dsc` halts the computer. Supposedly the culprit is the additional `while getline('.')[l:col-1 :] =~# '^\s*{'` loop at https://github.com/lervag/vimtex/commit/c45f9b4f94a0bba992ba5c8ba7a133f93cf13ff4#diff-69d39002f65b6ef8058c06e787b5a9d7R139  

Why ought `dsc` to delete all surrounding braces? Also `normal! f{` could jump to an escaped brace, which `vaB` prevents. Finally where's the part that deleted the command `\...` itself?

lervag:
Good point about jump list and visual selection. I'll work on it later. I just pushed a minor update to remove the loop, I'll work on the other things when I get the time.

Konfekt:
Ok, now `dsc` does not halt the computer anymore, but forgets deleting the braces `{}`.

Regarding the jump list, the `g`` command is meant to leave it intact though.

lervag:
This is an overview of remaining tasks and issues:

[]: Previous visual selection is overridden
[]: Jump list is overriden
[]: Use `vaB` instead of `f{` for improved robustness
[]: Ensure the command `\...` is deleted
[]: Ensure that (possibly) following braces are deleted

lervag:
[...]

PR #247, opened 2015-10-11
csc changes command before closest unescaped brace

Konfekt:
resolves https://github.com/lervag/vimtex/issues/245


lervag:
Thanks! I decided to improve this with a slightly more general approach that conserves the cursor position and similar in a better way. See #245 for continued discussion.

PR #255, opened 2015-10-20
Fix imaps mappings.

leomao:
Remove ultisnips dependency for imaps which don't use it.
Check if wrapper is `s:wrap_math_ultisnips` instead of matching `$1`.


lervag:
Thanks! This is mostly good, but the check for UltiSnips is not quite there. The reason is that there also exists another wrapper, `s:wrap_ultisnips`. This was the reason I matched against `$1`, since I thought it was a reasonable assumption. However, I agree it might be more suitable to check for an ultisnips wrapper, but then we should use a regexp, something like:

```vimL
if l:wrapper =~? 'ultisnips' && !s:has_ultisnips | return | endif
```

Here I propose `=~?` to ignore case, so that we also match different variants of the word Ultisnips.

leomao:
You're right!
Change to use `=~?` to check the wrapper.
Thanks!

lervag:
Thanks!

PR #276, opened 2015-11-12
imap_mappings: do not swallow leader outside math

Konfekt:
For example `4.` outside math environments should type `4.` not `.`.


lervag:
Thanks!

PR #278, opened 2015-11-12
allow for sending special keys in imaps

Konfekt:
for example, now
```vim
{ 'lhs_rhs' : ['__',    "_{}\\<lt>Left>"], 'leader' : '',   'wrapper' : 's:wrap_math'}
```
achieves what previously required UltiSnips#Anon()

### DIFF
--- a/autoload/vimtex/imaps.vim
+++ b/autoload/vimtex/imaps.vim
@@ -62,8 +62,8 @@ endfunction
 " Wrappers
 "
 function! s:wrap_math(lhs, rhs) " {{{1
-  return '<c-r>=<sid>is_math() ? ' . string(a:rhs)
-        \ . ' : ' . string(a:lhs) . '<cr>'
+  return '<c-r>=<sid>is_math() ? "' . a:rhs
+        \ . '" : "' .  a:lhs . '"<cr>'
 endfunction
 
 " }}}1


Konfekt:
Somehow for these kind of tricks double quotes are needed. Saw this somewhere on the vim mailing list pointed out by @ZyX-I

ZyX-I:
This suggestion looks strange. What is logical to need here is `"_{}\<Left>"` and that’s *all* (and I guess this variant works). Wanting to have *double* escaping (`<` to `<lt>` and then `\\<lt>`) here is crazy.

lervag:
I am curious: Why would one want this instead of the anonymous snippets? Snippets add the functionality to jump out, which is not possible with this sort of map.

And, I agree with @ZyX-I, it seems the required `RHS` needs escaping, which seems very inconvenient. I would definately consider the update if one could write it as `"_{}\<left>"`, but if one has to escape both the `<` and the `\\`, then it becomes too complicated.

Konfekt:
I came up with it because the `UltiSnips#Anon()` function does not work over here, and thus a solution  built-into Vim is to be preferred. 

Note that this commit does not change anything but replacing `'` by `"` on one line.

@ZyX-I I tried every possible combination, but only 
```vim
inoremap __ <C-R>=empty('') ? "_{}\<lt>Left>" : '__'<CR>
```
with double quote and escaping works (and the corresponding cnoremap fails and shows `<80>kl` instead of moving left!). The double escaping `\\` instead of `\` apparently is necessary here because it is a string to be called by `exe`.

@lervag can perhaps verify that all combinations (single quotes, single escape and `<` instead of `<lt>`) do not work.

Konfekt:
The most obvious advantage is that it adds ^^ and __ for people not using UltiSnips or NeoSnippet.
Regarding jumping out of `{}`, for example `<c-g>` in delimitmate does that. Somehow an autodelimiter completion plugin is tacitly assumed to be installed when using these imaps, since they only expand in math enviornments and thus require closing delimiter before-hand.

lervag:
Ok, I'll accept it. I agree the change is very small, and I believe it does not break anything.

Konfekt:
Now one could check  if ultisnips or neosnippet installed, and otherwise use the built-in mappings.


By the way, a correction, the check for the expansion of the imaps  does not need the closing delimiter, at least at the end of a line.

lervag:
I replied to the [comment](https://github.com/lervag/vimtex/commit/7892c988a5dc7b6222c66587bd6eca04d738cb93) on the commit, but I want to continue the discussion here.

My reply there was:

> Ah, yes. I didn't notice this. As I was already a little bit skeptical of this change, I find I want to revert it. I understand that this breaks things for you, and for that I'm sorry. You could consider to install neosnippet, and simply keep it unloaded except for a backend for these semi-snippets?

As an addendum to the comment there: An alternative is to implement the snippet functionality into vimtex. It actually does not need to be very complex, I think. If I used the same principle as was used by LaTeX-Suite (aka vim-latex), i.e. things like `\frac{<++>}{<++>}`, where `<++>` is a placeholder, then it should not be too difficult to implement. However, if I did this, then I would no longer need to keep the coupling with neosnippet and UltiSnips either...

Konfekt:
I'm using Ulti Snips but somehow doesn't do anything. As adding two braces and cursor left is essentially all there is to it, I thought a built in solution is around the corner. But somehow the handling if spécial characters turns out inscrutable.

The map that jumps out of the braces is determined by Ulti Snips so better for such a functionality insist on that dependency. But as a fallback just what was proposed will be just fine for most users.

lervag:
I've just opened an issue to discuss the future of the automatic snippets. I hope you might join the discussion with your opinions.

With that, I think I'll consider this discussion closed. I hope that's fine with you.

Konfekt:
OK !

Issue #283, opened 2015-11-20
untangle fold setup


Konfekt:
To use vimtex expression folding ideally
```vim
setlocal fdm=expr
setlocal foldexpr=vimtex#fold#level(v:lnum)
```
should suffice. 

However`vimtex#fold#level(v:lnum)` needs `g:vimtex_fold_enabled=1` since only then
```vim
    g:vimtex_fold_preamble, b:vimtex_fold_parts, g:vimtex_fold_envs.
```
are defined. However, `g:vimtex_fold_enabled=1` also overwrites `foldtext` and sets up `autocmd`s.

So how about defining 
```vim
    g:vimtex_fold_preamble, b:vimtex_fold_parts, g:vimtex_fold_envs.
```
always (and documenting the value of `foldexpr`)? 

Then `g:vimtex_fold_enabled` sets up
```vim
  setlocal foldmethod=expr
  setlocal foldexpr=vimtex#fold#level(v:lnum)
  setlocal foldtext=vimtex#fold#text()
```
Also, instead of the many autocmd's, even on `CursorMove` how about simply using https://github.com/Konfekt/FastFold using less frequent autocmd's ?

lervag:
I agree the folding options are a little bit too complex. Personally, I use the default options, which I think is similar to how FastFold functions. With `g:vimtex_fold_automatic`, we keep the `foldmethod=expr`, but toggles it when entering insert mode.

The manual (and default) mode uses a self destructive autocommand in order to update the folds when a file is loaded. Else there are no autocommands involved.

I'm a bit reluctant to make vimtex fully dependent on FastFold. However, I think I could consider to remove the automatic mode, and instead make this a "clean" `expr` mode.

Regarding the other part, I agree: the options should be defined also when the folding is disabled.

Konfekt:
Yeah, that's a fine solution, The automatic mode should do what proposed on top of the inital post; then FastFold can tap into it.

The non-automatic mode can continue working as before. 

Note that defering the fold update to the leave of insert mode caused a pause over here too, but that's probably due to concealing and  not so noticeable otherwise. 

It's fine as it is, but might give it a try with, that is,compare to automatic mode with FastFold.

Konfekt:
To clarify, automatic mode means clean mode.

lervag:
TODOs:

[...]

PR #284, opened 2015-11-22
document automatic fold setup

Konfekt:


lervag:
The automatic option is not yet removed, so I still need to explain the option. However, I am considering to remove the option altogether. In any case, I think we can continue this discussion in #283. I agree that the fold initialization may still be improved and made cleaner than it is.

Btw: Thanks for removing end-of-line spaces.

Konfekt:
Yeah, the same misunderstanding as in

https://github.com/lervag/vimtex/commit/eed762774e3174531897989d31146caa979bd2dd

I  erroneously thought the commit achieved what was mulled about at

#283

namely, " remove the automatic mode".

PR #300, opened 2015-12-15
Add support for (t)csh

wmvanvliet:
The syntax for shell commands for Linux/OSX assumed either bash or fish.
This PR adds a few if-statements to also support (t)csh.


lervag:
Thanks, this looks good.

Issue #204, opened 2015-07-17
Changing PDF viewer to Skim (Part II)


cmarshak:
Dear Dr. Lervag and VimTex,

This is a continuation of https://github.com/lervag/vimtex/issues/175.  I am using MacVim on Yosemite.

I am trying to get Skim to work with vimtex, and have not succeeded.  The documentation seems contradictory to some of the bug fix above, at least it appears that way to a new vim-user like myself.

In the docs, it suggests:

(1) To paste the code in the .vimrc:

    let g:vimtex_view_general_viewer = '/Applications/Skim.app/Contents/SharedSupport/displayline'
    let g:vimtex_view_general_options = '@line @pdf @tex'`

(2) And add the following to the .latexmkrc:

    $pdflatex = 'pdflatex -synctex=1 %O %S';

When I do this, the command `<leader>ll` does not open anything (normally preview is opened).

When I remove the above and include the following as suggested in #175:

    let g:vimtex_view_general_viewer = 'open' 
    let g:vimtex_view_general_options = '-a Skim'

Preview opens upon hitting `<leader>ll`.  I have been trying a bunch of different combinations of the above without any luck.  I like this vim-tex extension because of the exceptional documentation.  The site hosting vim-latex-suite was down yesterday, too.

I got this answer for the latex-suite:

    map ,ls :w<CR>:silent !make <CR>:silent !/Applications/Skim.app/Contents/SharedSupport/displayline -r <C-r>=line('.')<CR> %<.pdf %<CR><CR>

and this always works.  The above is courtesy of:

https://skyuuka.wordpress.com/2014/07/11/setup-synctex-with-vim-pdflatex-and-skim/
http://molecularclouds.blogspot.sg/2012/01/latex-vim-skim.html

I guess I could map `<leader>ll` to a combination of the above and it's current status, but thought I would let you know about the obstacle.

Any suggestions on how to fix it would be appreciated.  Even tangential comments about the framework of this bundle would be awesome because I want to learn more about vim and how it works.

lervag:
Dear @cmarshak,

Since #175, I've made some updates and simplifications to the viewer code. In particular, I've implemented a placeholder scheme where `@pdf`, `@tex`, and `@line` indicates that the viewer should insert the current pdf file name, tex file name and line number (respectively).

I would like you to try a couple of things:

1. Try to set

     let g:vimtex_view_general_options = '-r @line @pdf @tex'`

[...]

PR #306, opened 2015-12-30
Escaping '%' in latexmk command

hijack111:
- In my version of Vim (7.4) both 'if system' and 'if silent' cannot correctly detect existence of these function/command.
- When adding rules to g:vimtex_latexmk_options, the wildcard used by latexmk (%S, %O, etc.) should be correctly escaped.

### DIFF
--- a/autoload/vimtex/util.vim
+++ b/autoload/vimtex/util.vim
@@ -94,17 +94,17 @@ function! vimtex#util#execute(exe) " {{{1
     set shellquote& shellpipe& shellredir& shellslash&
   endif
 
-  if system
-    if silent
+  if exists('*system')
+    if exists(':silent')
       silent call system(cmd)
     else
       call system(cmd)
     endif
   else
-    if silent
-      silent execute '!' . cmd
+    if exists(':silent')
+      silent execute '! sh -c ' . shellescape(cmd, 1)
     else
-      execute '!' . cmd
+      execute '! sh -c ' . shellescape(cmd, 1)
     endif
 
     if !has('gui_running')


lervag:
`if system` and `if silent` does not check if the functions/commands exist, but if they should be used. `silent` and `system` are variables.

Escaping the wild cards seems necessary, but require more testing before I can accept. The point is: I want this to work on all systems. It would therefore be good if you could create a minimal working example where `g:vimtex_latexmk_options` is set so that wild cards must be escaped.

hijack111:
That makes perfect sense. I discovered this "bug" while I am trying to use
Skim as my previewer (I work on Mac), for which I have to provide %S to my
$pdf_update_command. Please check the configuration I feed to vimtex (as
a development on #204 <https://github.com/lervag/vimtex/issues/204>):

" Use Skim application to update & forward searching in PDF
let g:vimtex_latexmk_options = "$pdflatex =~ s/ / -synctex=1 /;"
let g:vimtex_latexmk_options .= "$pdf_update_method = 4;"
let g:vimtex_latexmk_options .= "$pdf_update_command =
\"/Applications/Skim.app/Contents/SharedSupport/displayline -g -r 999999999
%S\";"
let g:vimtex_latexmk_options = "-e '" . g:vimtex_latexmk_options . "'"

let g:vimtex_view_general_viewer =
'/Applications/Skim.app/Contents/SharedSupport/displayline'
let g:vimtex_view_general_options_latexmk = '-g -r 999999999'
let g:vimtex_view_general_options = '-g -r @line @pdf @tex'


I wish you a wonderful 2016!

hijack111:
I discovered this "bug" while I am trying to use Skim as my previewer (I work on Mac), for which I have to provide `%S` to my `$pdf_update_command`. Please check the configuration I feed to vimtex (as a development on #204 ):

```
let g:vimtex_latexmk_options = "$pdflatex =~ s/ / -synctex=1 /;"
let g:vimtex_latexmk_options .= "$pdf_update_method = 4;"
let g:vimtex_latexmk_options .= "$pdf_update_command = \"/Applications/Skim.app/Contents/SharedSupport/displayline -g -r 999999999 %S\";"
let g:vimtex_latexmk_options = "-e '" . g:vimtex_latexmk_options . "'" 

let g:vimtex_view_general_viewer = '/Applications/Skim.app/Contents/SharedSupport/displayline'
let g:vimtex_view_general_options_latexmk = '-g -r 999999999'
let g:vimtex_view_general_options = '-g -r @line @pdf @tex'
```

lervag:
I see. Could I suggest that you instead populate a `.latexmkrc` file, i.e.:

```perl
$pdflatex = 'pdflatex -synctex=1';
$pdf_update_method = 4;
$pdf_update_command = '/Applications/Skim.app/Contents/SharedSupport/displayline -g -r 999999999 %S';
```

Notice that you can have project specific `latexmkrc` files as well as global and per user.

I think this will solve your issue with no need to update vimtex. I actually did not intend for such complicated options to get passed through `g:vimtex_latexmk_options`, and I don't really see a case where it will be necessary. I think the common latexmk configuration files is the better way to go, since this will also make things work if you use latexmk through your shell.

PR #326, opened 2016-01-14
Updated SumatraPDF setting example

blyoa:
Currently, `g:vimtex_view_general_options` example for SumatraPDF doesn't contain `-reuse-instance`,
so even if a SumatraPDF window is already opened by `:VimtexCompile`, another window is opened by `:VimtexView`.
(By contrast, the example for Okular and qpdfview contains `--unique`, so opened window is reused.)

How about adding `-reuse-instance` to `g:vimtex_view_general_options` in the SumatraPDF setting example to avoid opening multiple windows?

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -815,7 +815,7 @@ Options~
           Backward search must be set up from the viewer. >
 
   let g:vimtex_view_general_viewer = 'SumatraPDF'
-  let g:vimtex_view_general_options = '-forward-search @tex @line @pdf'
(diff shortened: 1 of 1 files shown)


lervag:
I have to take your word for this, since I don't use windows myself. I want to keep the docs within the 80 columns limit, so I took the liberty to push the update with a continued line update.

blyoa:
I'm sorry for breaking the style of the 80 columns limit and thank you for your adjustment.

lervag:
No problem, thanks for noticing the mistake :-)

fre. 15. jan. 2016, 12.26 skrev blyoa <notifications@github.com>:

Issue #262, opened 2015-10-28
Add support for neovim


sschaef:
Right now vimtex doesn't fully work with neovim. Running latexmk works without problems, but notifying neovim about any errors/warnings is not possible since it doesn't include the `clientserver` module. neovim instead uses the [`msgpack-rpc`](https://neovim.io/doc/user/msgpack_rpc.html) protocol to communicate with clients.

It would be a welcome feature get support for neovim.

lervag:
Thanks for opening the issue! I'm considering to use neovim myself these days, but I haven't gotten there yet.

Do you know if there are more issues than the callback feature? If so, I think we should create a task list in your original post.

I've understood that the `msgpack-rpc` protocol is supposed to be more powerful, but do you know if there is a similarly easy way to communicate with a server as the `clientserver` functionalit in Vim? That is, in Vim I can do this:

    vim --servername NAME --remote-expr 'vimtex#latexmk#callback()`

How would I do the same thing in neovim using `msgpack-rpc`? I don't really like the idea of creating a dedicated script (ruby/python) just to do this simple thing.

Another thing: Since neovim allows detached subprocesses, I could perhaps utilize this. Alternatively, I could use the terminal in neovim to run the latexmk process. Ideas and input is very much welcome here!

sschaef:
Indeed, for such a simple thing, neovim job control API would be sufficient (but I have never used it by myself therefore can't help you here): https://neovim.io/doc/user/job_control.html

Here is an example that shows how to use it: https://gist.github.com/tarruda/1b5010531bb6ae7382a4

Beside from that I don't think any further changes should be necessary. neovims full help page: https://neovim.io/doc/user/nvim.html

phcerdan:
I just want to pop out to say that I am using vimtex right now with neovim. I am pretty sure that I am missing features, but for me is already good enough. I have sync it with Zathura and seeing the changes in the pdf after saving the file in neovim is all I require.

Just in case somebody comes to this issue thinking that the plugin cannot run in neovim. Hope it helps, cheers.

lervag:
@sschaef I've added a comment at the [neovim issue #1750](https://github.com/neovim/neovim/issues/1750). I think there might be hope that neovim gets a similar functionality to vim with `--remote-expr`, which would make it easy to add support for neovim in vimtex. I'll leave this issue unresolved while waiting for an update at the neovim side.

lervag:
[...]

PR #334, opened 2016-01-27
Update remote call for latest version of neovim-remote

nacnudus:
See #262.


lervag:
Thanks! I'll check this tomorrow, there might be more places that needs updating, and I want to check that everything works as it should with normal Vim. :)

lervag:
I'm sorry, but this breaks compatibility with regular Vim, and thus is not an option.

PR #346, opened 2016-02-06
New general callback and expand @line for general_options_latexmk

nhooyr:
Expanding the `@line` escape allows us to open the file in the previewer wherever the cursor is, and the
callback lets people implement scripts as they like it to open/update their favourite pdf viewer to wherever the cursor is after compilation. latexmk's pdf_update_command doesn't let us do that, it can only update the file, but this allows anyone to easily make their pdf viewer go to wherever the cursor currently is in vim.

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -131,9 +131,18 @@ endfunction
 
 " }}}2
 function! s:general.latexmk_append_argument() dict " {{{2
+  let opts = g:vimtex_view_general_options_latexmk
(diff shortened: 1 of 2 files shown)


lervag:
The callback is a good idea. I have a slightly more neat way of implementing this, though, so I will first accept your pull request, then modify it slightly.

lervag:
Please see the updated version, let me know what you think.

nhooyr:
LGTM

PR #356, opened 2016-02-12
Completion patterns for \cite, \ref and \includeonly etc.

djt308:
Modified the patterns as they did not cover some forms of biblatex and cleveref commands such as, `\parencite[pre][post]{key}`, `\crefrange{label1}{label2}`, `\cpageref{label1}` and `\cpagerefrange{label1}{label2}`.

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -68,7 +68,7 @@ endfunction
 " {{{1 Bibtex
 
 let s:bib = {
-      \ 'pattern' : '\v\\\a*cite\a*%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*',
+      \ 'pattern' : '\v\\\a*cite\a*%(\[\s*[^\]\{\}]*\]){0,2}\s*\{[^\}]*',
       \ 'enabled' : 1,
       \ 'bibs' : '''\v%(%(\\@<!%(\\\\)*)@<=\%.*)@<!'
       \          . '\\(bibliography|add(bibresource|globalbib|sectionbib))'
@@ -223,7 +223,7 @@ endfunction
 " {{{1 Labels
 
 let s:ref = {
-      \ 'pattern' : '\v\\v?%(auto|eq|[cC]?%(page)?|labelc)?ref%(range)?\s*\{%([^\}]*%(\}\{)?)',
+      \ 'pattern' : '\v\\v?%(auto|eq|[cC]?%(page)?|labelc)?%(ref\{\s*[^\}]*|refrange\{\s*%([^,\}]*%(\}\{)?))',
       \ 'enabled' : 1,
       \}
 
@@ -321,7 +321,7 @@ endfunction
 " {{{1 Filenames (\includegraphics)
 
 let s:img = {
-      \ 'pattern' : '\v\\includegraphics\*?%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*',
+      \ 'pattern' : '\v\\includegraphics\*?%(\[\s*[^\]\{\}]*\]){0,2}\s*\{[^\}]*',
       \ 'enabled' : 1,
       \}
 
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -223,7 +223,7 @@ endfunction
 " {{{1 Labels
 
 let s:ref = {
-      \ 'pattern' : '\v\\v?%(auto|eq|page|[cC]|labelc)?ref%(range|name\{\w*\})?\s*\{%([^\}]*%(\}\{)?)',
+      \ 'pattern' : '\v\\v?%(auto|eq|[cC]?%(page)?|labelc)?ref%(range)?\s*\{%([^\}]*%(\}\{)?)',
       \ 'enabled' : 1,
       \}
 
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -68,7 +68,7 @@ endfunction
 " {{{1 Bibtex
 
 let s:bib = {
-      \ 'pattern' : '\v\\\a*cite\a*%(\s*\[[^]]*\])?\s*\{[^{}]*',
+      \ 'pattern' : '\v\\\a*cite\a*%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*',
       \ 'enabled' : 1,
       \ 'bibs' : '''\v%(%(\\@<!%(\\\\)*)@<=\%.*)@<!'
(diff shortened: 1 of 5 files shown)


lervag:
I think this looks like a valuable contribution. I'll add some specific notes, just to make sure nothing is broken.

autoload/vimtex/complete.vim
```diff
@@ -68,7 +68,7 @@ endfunction
 " {{{1 Bibtex
 
 let s:bib = {
-      \ 'pattern' : '\v\\\a*cite\a*%(\s*\[[^]]*\])?\s*\{[^{}]*',
+      \ 'pattern' : '\v\\\a*cite\a*%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*',
```
lervag: I could parse the regexp, but it seems easier to ask: Why do you change this? Could you give an example of what the old version did not match (or matched erroneously)?

autoload/vimtex/complete.vim
```diff
@@ -223,7 +223,7 @@ endfunction
 " {{{1 Labels
 
 let s:ref = {
-      \ 'pattern' : '\v\\v?%(auto|eq|page|[cC]|labelc)?ref\s*\{[^{}]*',
+      \ 'pattern' : '\v\\v?%(auto|eq|[cC]?%(page)?|labelc)?ref%(range)?\s*\{%([^\}]*%(\}\{)?)',
```
lervag: I understand that you add support for e.g. `cpageref` and `crefrange` and similar. What does the last part fix?

doc/vimtex.txt
```diff
@@ -1377,9 +1377,10 @@ documents with |neocomplete| and |vimtex|s omni completion function: >
   endif
   let g:neocomplete#sources#omni#input_patterns.tex =
         \ '\v\\%('
-        \ . '\a*%(ref|cite)\a*%(\s*\[[^]]*\])?\s*\{[^{}]*'
-        \ . '|includegraphics%(\s*\[[^]]*\])?\s*\{[^{}]*'
-        \ . '|%(include|input)\s*\{[^{}]*'
+        \ . '\a*%(cite)\a*%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*'
+        \ . '\a*%(ref)%(range)?\s*\{%([^\}]*%(\}\{)?)'
+        \ . '|includegraphics\*?%(\s*[^\]]*\]?){0,2}\s*\{[^\}]*'
+        \ . '|%(include%(only)?|input)\s*\{[^\}]*'
```
lervag: Here and for the `YCM`: Does this fix more than simply making things consistent with the above changes?

autoload/vimtex/complete.vim
djt308: It would not match the case for two optional arguments, as in `\parencite[prenote][postnote]{key}`.
It would only match one or zero optional arguments, hence the `{0,2}`. The same applies to the changes to `\includegraphics{}`, although I have also included the case of the starred version, i.e. `includegraphics*{}`.

autoload/vimtex/complete.vim
djt308: The end bit is for the range references; it detects the closing of the first set of braces and the immediate opening of the second set of braces, in order to prompt the menu to open a second time, e.g. `\crefrange{<popupmenu> --> \crefrange{label1}{<popupmenu>`.

I wasn't happy with this solution as it also allowed the same behaviour for non-range references. I have fixed this in commit 393e9bb.

doc/vimtex.txt
djt308: No, it's just consistent with the other changes.

autoload/vimtex/complete.vim
```diff
@@ -68,7 +68,7 @@ endfunction
 " {{{1 Bibtex
 
 let s:bib = {
-      \ 'pattern' : '\v\\\a*cite\a*%(\s*\[[^]]*\])?\s*\{[^{}]*',
+      \ 'pattern' : '\v\\\a*cite\a*%(\[\s*[^\]\{\}]*\]){0,2}\s*\{[^\}]*',
       \ 'enabled' : 1,
       \ 'bibs' : '''\v%(%(\\@<!%(\\\\)*)@<=\%.*)@<!'
```
djt308: There was a problem with being able to add additional sets of braces to the end and having the popup menu opening when each new set was started. This was because the opening square bracket was not explicitly included. Braces have also been excluded from being match between square brackets as an opening brace triggers the popup menu.

autoload/vimtex/complete.vim
```diff
@@ -223,7 +223,7 @@ endfunction
 " {{{1 Labels
 
 let s:ref = {
-      \ 'pattern' : '\v\\v?%(auto|eq|page|[cC]|labelc)?ref\s*\{[^{}]*',
+      \ 'pattern' : '\v\\v?%(auto|eq|[cC]?%(page)?|labelc)?%(ref\{\s*[^\}]*|refrange\{\s*%([^,\}]*%(\}\{)?))',
       \ 'enabled' : 1,
       \}
```
djt308: I didn't like the previous implementation as it allowed a second set of braces to trigger the popup menu to open for non-range references, e.g. `\ref{<popupmenu> --> \ref{label}{<popupmenu>`. The new pattern also prevents the menu from staying open if a user attempts to enter a list of labels in a range reference, instead of just a single label for each of the two arguments e.g. `\crefrange{label1,...,labelN}{label1,...,labelM}` instead of `\crefrange{label1}{label2}`. For non-range references, the menu will stay open to allow a list of labels.

lervag:
Ok. I really appreciate the effort! However, I am wondering: Would you be willing to submit this as separate pull requests, one for each of:

1. Support for `\includeonly`
2. Improve regexps for label completion (i.e. the `s:ref` regexps)
3. Improve regexps for bibliography completion (i.e. the `s:bib` regexps)

I would also be very happy if you could add to the completion test file under [`test/feature/completions/main.tex`](https://github.com/lervag/vimtex/blob/master/test/feature/completions/main.tex) the commands that you mean should now be supported by the improved regexps for label and bibliography completion. This would make it easy for me in the future to check if further updates will break old behaviour.

I understand that this request may seem somewhat high on the workload side, so please forgive me for that. The point is that I think this will make it much easier to pinpoint problems in the future (for instance with bisection), e.g. if problems are related to this update. Regexps are tedious, and even though I consider myself well aquinted with advanced regexps and I think this does look very good, I also know that I often spend time debugging old regexps...

I very much hope that you are willing to do this. If not, then I think I want to do it myself when I get the time (I will make sure to give you credit in my commits).

djt308:
Okay, I will submitt separate pull requests for each component.

lervag:
Great, thank you! I really appreciate it!

PR #358, opened 2016-02-14
Alter regex for \cite to allow [0,1,2] optional arguments, eg \parencite[arg1][arg2]{key}

djt308:


lervag:
I think there is an error here. For some reason, you haved added `{` and `}` to the list of atoms to not match while matching the optional arguments. The error is that you write `\{` and `\}` --- this is NOT necessary, and will instead mean that `\` is added to the group. The correct syntax should be `[^{}\]]`, but my question is still valid: Why do you want to add `{` and `}`? Please give me any example that this is necessary, else I think you should remove it.

You also make the same error in the last part, where you add `\` to the `{` and `}`, which is again wrong.

lervag:
To be clear, I think the following should suffice:

```vim
       \ 'pattern' : '\v\\\a*cite\a*%(\s*\[[^]]*\]){0,2}\s*\{[^{}]*',
```

lervag:
To be very clear: The above mentioned error/confusion seems to be present a lot of places.

djt308:
Strange, I don't get any errors. I'm using youcompleteme; I don't have neocomplete.

I'm new to vim regexes and I find them a little confusing, but I thought the `\v` implied all literal non-alphabetic ASCII characters had to be escaped.

I added the `{}` and `\[` were added at the same time when I was fixing the problem I commented on in #356 see commit 393e9bb. I mistakingly thought that typing a `{` while entering an optional argument would trigger the completetion menu to appear, but the literal `[]` prevent this. Therefore, removing the `{}` does not affect any functionality.

lervag:
`\v` does imply exactly what you mean, except within the `[...]` groups, where any character will be matched explicitly.

I notice that the ycm regexps are not vim regexps, and so perhaps things work differently there.

I will be happy to accept the pull requests (this, #359 and #360) when you update according to the discussed changes.

djt308:
Are there any further problems with this, #359 and #360?

lervag:
Sorry, I was not notified about the updates! I'll look into this tonight!

PR #359, opened 2016-02-14
Alter regex for \ref to provide completion for most cleveref commands.

djt308:


lervag:
See #358: You should remove the added `\`s from the match groups `[^...]`, and if you have added `{}` to a match group for an optional argument, then I want to know why first.

djt308:
See comment in #358

PR #360, opened 2016-02-14
Alter regex for \includegraphics

djt308:
to provide completion for up to two optional arguments and the starred version.


lervag:
Again see #358.

PR #367, opened 2016-02-16
Add examples of various commands to test completion functionality.

djt308:


test/feature/completions/main.tex
```diff
@@ -9,6 +11,8 @@
 \usepackage[nonumberlist,acronymlists={gloss,symbolslist}]{glossaries}
 \input{glossary.tex}
 
+\includeonly{sub1,sub2 with spaces}
```
lervag: I just realized that this might not need any treatment from vimtex! `\includeonly` does not really include anything, instead, it controls which of the other `\include`d files to compile. Thus, I actually propose to revert the previous pull request, unless you have a good argument for why I'm wrong.

lervag:
Except for my one comment, this is very good! It will be very helpful in future debugging and testing for the label and citation completion. Thanks!

lervag:
Btw: I'll accept after we settle the minor thing about the `\includeonly`.

djt308:
Okay, so after having a closer look at the functions in `autoload/vimtex.vim` and, `autoload/vimtex/parser.vim` I believe they have nothing to do with completion, therefore the `%(only)?` bit can be removed from the patterns, correct me if I'm wrong; I'm relying on your intimate knowledge of your code to pickup on my mistakes or incorrect assumptions.

However, I believe that having completion for `\includeonly` is worthwhile and that the changes to `completion.vim` and `vimtex.txt` should remain. I'm happy to handle the clean up of `autoload/vimtex.vim` and, `autoload/vimtex/parser.vim` in whatever manner you see fit.

lervag:
I agree that filenames should be completed for `\includeonly`. I've cleaned up now. I think I'm ready to accept this, then.

PR #370, opened 2016-02-17
Document omni completion setup for VimCompletesMe

Konfekt:


doc/vimtex.txt
```diff
@@ -1420,7 +1421,25 @@ To enable automatic completion with |youcompleteme|, use the following options:
         \ 're!\\includegraphics([^]]*])?{[^}]*',
         \ 're!\\(include(only)?|input)\{[^{}]*'
         \ ]
+<
```
lervag: Just mentioning: The end `<` is not needed when the following line is not indented. Though, it doesn't really matter.

doc/vimtex.txt
```diff
@@ -1420,7 +1421,25 @@ To enable automatic completion with |youcompleteme|, use the following options:
...
+VimCompletesMe~
+                                                  *vimtex-complete-vcm*
+A plugin that overrides <tab> to trigger various built-in completions.                
+The plugin is available here: https://github.com/ajh17/VimCompletesMe.
 
+The following options may be used to enable automatic completion for LaTeX
+documents with |neocomplete| and |vimtex|s omni completion function: >
```
lervag: How is neocomplete relevant here? If both plugins are needed, what settings are needed for neocomplete?

lervag:
`VimComlpetesMe` is not an autocomplete plugin. It seems somewhat out of place, and I don't understand why you mention neocomplete in the description for autocomplete settings.

In addition, I think it makes sense to add some minor introductory text to the beginning of the section, i.e. [here](https://github.com/lervag/vimtex/pull/370/files#diff-4bb7bb82dfb788844c6c9e003beccb2dR1379).

Konfekt:
Ok, fixed. VimCompletesMe surely does not do much (but what defines an autocomplete plugin?!), but what it does, like neocomplete, is decide whether to use omni-completion depending on the text preceding the cursor (and fall back to the built-in `<c-p>` completion otherwise, whereas neocomplete to user-completion). Vim's built-in completion has, at least in TeX, one advantage in its dictionary completion.

lervag:
Actually, I tend to think that one should differentiate between automatic completion and simply completion. Vim does not offer any autocompletion, and nor do VimCompletesMe as far as I can see. It does not mean it is not useful, it is simply a matter of definitions. I think autocompletion means that the completion starts without pressing any keys.

In any case, I think this looks good. Thanks!

PR #373, opened 2016-02-18
Fix completion patterns for YCM.

djt308:


doc/vimtex.txt
```diff
@@ -1394,7 +1394,7 @@ documents with |neocomplete| and |vimtex|s omni completion function: >
   let g:neocomplete#sources#omni#input_patterns.tex =
         \ '\v\\%('
         \ . '\a*cite\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
-        \ . '|\a*ref%(\s*\{[^}]*|range\s*\{[^,}]*%(\}\{)?)'
+        \ . '|\a*ref%(\s*\{[^}]*|range\s*\{[^,}]*%(}\{)?)'
         \ . '|includegraphics\*?%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
         \ . '|%(include%(only)?|input)\s*\{[^}]*'
```
djt308: As far as I can tell from vim's docs, I don't think closing braces need to be escaped.

djt308:
I think the patterns are about as minimalistic as they can be.

doc/vimtex.txt
lervag: You are right, but I still find the syntax more understandable. It seems more consistent to escape both opening and closing braces.

lervag:
See my comment on the commit. I think the current syntax is more consistent. Unless you have a good argument for why it is not?

Are you using youcompleteme? If you are, then I will accept this (possibly with the above mentioned change). If not, then I want to mention that I am worried that the use of braces for the `re!` patterns might be bad, i.e. we now use `{0,2}` for 0 to 2 matches, but we also use `(}{)?` to match 0 or 1 occurences of `}{`, thus the braces both have special meaning and no special meaning. It seems strange to me.

djt308:
Yes, I'm using youcompleteme and the above pattern is working.

Braces only have special meaning in the context of repetition with the following syntax: `{m,n}` ,  `{m}`, `{m,}`  and  `{,n}` , where,  `m`  and  `n`  are integers. For all other cases, braces are considered literal. As far as I know, this is the case for Python regular expressions, which I assume is the flavour used in youcompleteme.

Let me know what you want with respect to escaping braces?

lervag:
Great, thanks for the clarification. I'll accept the pull request as is, it works, and if people get confused by the unescaped closing brace, then let them :)

djt308:
Actually, for the youcompleteme pattern, neither of the braces are escaped for the above reasons. Only the vim flavoured patterns require a literal opening brace to be escaped because of the `\v`. As far as I can tell, if `\v` wasn't used, all braces would be considered literal and the syntax for repetition would be `\{m,n}`. However, not using `\v` would mean that other characters, such as `()|` would have to be escaped to provide their special meaning. Please correct me if I am wrong.

lervag:
No, you are quite right.

Issue #403, opened 2016-03-30
Glossary completion


mbrunnen:
Hi
Thanks for this nice new feature #263 :)
I finally found my problem and can reproduce it:
Change in [`test/features/completions/main.tex`](https://github.com/lervag/vimtex/blob/master/test/features/completions/main.tex) on line 12:
    `\input{glossary.tex}`
to
    `\loadglsentries{glossary}`
and restart vim.
Then the completion breaks for me.
Best regards,
Manoel

lervag:
Fixed through #404.

PR #404, opened 2016-03-30
Fix for issue #403

mbrunnen:
Fixed issue #403. 
Additionally I changed the test case for glossaries in the way, they propose it in the [official manual](http://ftp.fernuni-hagen.de/ftp-dir/pub/mirrors/www.ctan.org/macros/latex/contrib/glossaries/glossaries-user.pdf)

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -396,7 +396,7 @@ endfunction
 " {{{1 Glossary
 
 let s:gls = {
-      \ 'patterns' : ['\v\\gls\s*\{[^}]*$'],
+      \ 'patterns' : ['\v\\(gls|Gls|GLS)(pl)?\s*\{[^}]*$'],
       \ 'enabled' : 1,
       \}
 
--- a/test/features/completions/glossary.tex
+++ b/test/features/completions/glossary.tex
@@ -1,5 +1,5 @@
-\newglossary[gla]{gloss}{glb}{glc}{Glossar}
-\newglossary[sya]{symbols}{syb}{syc}{Symbolverzeichnis}
+\newglossary*{gloss}{Glossary}
+\newglossary*{symbols}{List of symbols}
 
 \makeglossaries
 
--- a/autoload/vimtex/parser.vim
+++ b/autoload/vimtex/parser.vim
@@ -9,7 +9,7 @@ endfunction
 
 " }}}1
 function! vimtex#parser#init_script() " {{{1
-  let s:input_line_tex = '\v^\s*\\%(input|include|subimport|subfile)\s*\{'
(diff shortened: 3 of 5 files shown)


autoload/vimtex/complete.vim
```diff
@@ -396,7 +396,7 @@ endfunction
 " {{{1 Glossary
 
 let s:gls = {
-      \ 'patterns' : ['\v\\gls\s*\{[^}]*$'],
+      \ 'patterns' : ['\v\\(gls|Gls|GLS)(pl)?\s*\{[^}]*$'],
```
lervag: Are there any differences between `\gls`, `\Gls`, and `\GLS`? Can we simply use case insensitive matching (`[gG][lL][sS]`)?

test/features/completions/glossary.tex
```diff
@@ -1,5 +1,5 @@
-\newglossary[gla]{gloss}{glb}{glc}{Glossar}
-\newglossary[sya]{symbols}{syb}{syc}{Symbolverzeichnis}
+\newglossary*{gloss}{Glossary}
```
lervag: Could you explain this update; is there a reason why the old version should not also be tested?

lervag:
Hi! I just want to say that I've noticed your pull request, and I think it might be acceptable. However, I want to consider the change to the parser; I am not quite sure if it is suitable to update the include regexp. I need to think about this, but I don't have the time right now.

I have also begun writing more rigorous tests through the Vader plugin, and I might want to add tests for this as well. I could do that afterwords, of course, but I want to have a clear head before merging.

autoload/vimtex/complete.vim
mbrunnen: It termines the case of the reference. gls = abc letters, Gls = Abc, GLS = ABC. Others are not defined. So we would match much more: gLS, GLs,...

test/features/completions/glossary.tex
mbrunnen: Because gla, glb, glc are the file-extensions, that I chose arbitrary. With the starred version of the command the file-extensions are chose automatically: gloss-gls, gloss-glo, gloss-glg. The standard extensions for a single glossary (by omitting \newglossary at all) are gls, glo, glg. Anyway, the generated files are not used in your approach.

mbrunnen:
I am aware that changing the parses is probably not the most elegant way, no hard feelings ;)

lervag:
Ok, I'll merge this, then possibly make some additional updates.

lervag:
I just pushed an update that moves the added `|loadglsentries` from the predefined `s:input_line_tex` regexp to a manually added regexp specifically for the gls completion. It seems to work as expected.

mbrunnen:
affirmative.

PR #418, opened 2016-04-05
Added a new filename completer for pdfpages

mbrunnen:
If it's wanted, you can merge the file completion for including whole pdf pages. It's done by the package pdfpages. I use it for data sheets, certificates, etc.


lervag:
Thanks, this seems well implemented and useful!

PR #419, opened 2016-04-05
Added filename completer for standalone files

mbrunnen:
File completion for standalones. The package is called standalone. I use them for Tikz, so i don't have to render the picture every time and also it is a good way to test the Tikz-picture.


lervag:
Thanks!

PR #422, opened 2016-04-06
Fix file-line-error by letting latexmk deal with it

frankier:
The current method can cause problems with a latexmkrc containing something like the following:

`$pdflatex = "env openout_any=a lualatex %O %S";`

In this case, a command like `env openout_any=a -file-line-error lualatex -file-line-error ...` will be run.

Instead we can just pass -file-line-error to latexmk, which will pass it one for us.

### DIFF
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -414,8 +414,7 @@ function! s:latexmk_build_cmd() " {{{1
   let cmd .= ' -verbose -pdf ' . g:vimtex_latexmk_options
 
   if g:vimtex_latexmk_file_line_error
-    let cmd .= ' -e ' . vimtex#util#shellescape(
-          \ '$pdflatex =~ s/ / -file-line-error /')
+    let cmd .= ' -file-line-error '
   endif
 
   if g:vimtex_latexmk_build_dir !=# ''


lervag:
Wow. Does this really work? It seems almost too simple...

lervag:
I never noticed the `-showextraoptions` option of latexmk. I'm curious: Is there any reason NOT to have `-file-line-error`? In the old version, I have used an option simply because of problems like you describe, so that people can turn it off. With the option passed directly to latexmk, I think perhaps there is no reason to have this be optional..?

frankier:
Yes, I've tested it and no I can't think of a reason for it to be optional now actually.

lervag:
Do you know if this is a recent update to latexmk, or if it may be expected to be commonly available?

frankier:
From this page: http://users.phys.psu.edu/~collins/software/latexmk-jcc/versions.html it looks like perhaps V. 4.31 dated 30 March 2012 might've added it if it wasn't in a previous version. In terms of Ubuntu this is in Trusty but not precise (see http://packages.ubuntu.com/search?suite=precise&searchon=names&keywords=latexmk and http://packages.ubuntu.com/search?suite=trusty&searchon=names&keywords=latexmk ). Assuming this is when it was introduced, it's been around for 4 years now, which is a reasonable amount of time.

lervag:
Agreed. I like this very much, so I think we'll begin with simplifying. I'll consider workarounds if anyone reports issues due to old versions.

lervag:
Again, thanks. I've now also removed the old option.

PR #428, opened 2016-04-08
Fixes #421

### LINKED ISSUE #421: QuickFix entries are incorrect
After compiling a `tex` file that contains errors, the quickfix window pops up. However, `:cn` and `:cp` does not correctly navigate through the entries in quickfix. In fact, it opens a new buffer.

Here are the steps to reproduce this problem on Windows.

Step 1: open gvim by double click gvim icon and open a `tex` file that contains two errors.
Step 2: compile using `\ll`, then quickfix window opens.
Step 3: use `:cn` to navigate to the next error. Instead of pointing to the error, gvim opens a new buffer.

Remarks:
1. The way one opens the `tex` file matters. If one uses something like `cmd`, `gvim c:\test.tex`, the above problem disappears.
2. Adding `let g:vimtex_quickfix_fix_paths=1` does not solve the problem.
3. Adding `set autochdir` solves the problem.
4. The above problem can also be reproduced on Ubuntu, again as long as the file is not opened from the command line `gvim ~/test.tex` for example.

fuzzybear3965:
I have tested compilation with `:set noautochdir`. The one line change in the commit seems to fix this issue. However, paths are a little gross in Windows (e.g. `C://Users//Some User//Desktop//test.tex`). This can be ameliorated later.

### DIFF
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -410,8 +410,13 @@ function! s:latexmk_build_cmd() " {{{1
     endif
   endif
 
-  let cmd .= ' -verbose -pdf -file-line-error '
-  let cmd .= g:vimtex_latexmk_options
+  " Add general options for latexmk
+  if !empty(g:vimtex_latexmk_options)
+    let cmd .= ' ' . g:vimtex_latexmk_options
+  else
+    let cmd .= ' -verbose -pdf -file-line-error'
+    let cmd .= ' -synctex=1 -interaction=nonstopmode'
+  endif
 
   if g:vimtex_latexmk_build_dir !=# ''
     let cmd .= ' -outdir=' . g:vimtex_latexmk_build_dir
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -676,12 +676,25 @@ Options~
   Default value: 0
 
 *g:vimtex_latexmk_options*
-  A string of options that will be passed on to `latexmk`.
+  By default, |vimtex| passes the following options to `latexmk`: >
+
+     -pdf                     -- Compile to pdf documents
+     -verbose                 -- More verbose output
+     -file-line-error         -- Enable file:line:error style messages
+     -synctex=1               -- Generate synctex data
+     -interaction=nonstopmode -- Disable interactive mode
+
+<  However, if `g:vimtex_latexmk_options` is defined and is nonzero, it will
+  override the above with the user supplied options.
 
   Note: Options may also be specified indirectly to `latexmk` through
-        both a global and a project specific `.latexmkrc` file.
+        both a global and a project specific `.latexmkrc` file. One should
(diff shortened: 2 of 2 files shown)


lervag:
Note: As you can see I've recently begun continuous integration testing with Vader (by junegunn). The testing framework is very nice, and I think it will become extremely useful. However, there are some tests I've not yet managed to get working, which makes the tests fail.

lervag:
I don't quite understand why my commits are part of this pull request, but I think it can be resolved through some git rebasing. You will probably need to do `git push -f` when finished, but that is safe since you only push to your own fork.

autoload/vimtex/latexmk.vim
```diff
@@ -439,7 +439,7 @@ function! s:latexmk_build_cmd() " {{{1
     let cmd .= b:vimtex.viewer.latexmk_append_argument()
   endif
 
-  let cmd .= ' ' . vimtex#util#shellescape(b:vimtex.base)
+  let cmd .= ' ' . vimtex#util#shellescape(b:vimtex.root . '/' . b:vimtex.base )
```
lervag: I think `b:vimtex.tex` should be the same as the sum of root and base here..?

lervag:
I now see that the only actual commit is 9d885c9. Thus the pull request should only involve a single commit as well.

fuzzybear3965:
Yes, `b:vimtex.tex` is preferable to my solution. I would recommend using that, instead. I actually did rebase, recently. I something happened to make your remote and my work diverge. I've cleaned up my repository and force pushed. But, you may not want to integrate the pull request. You can reject the pull request and change `b:vimtex.base` for `b:vimtex.tex`.

lervag:
I'm curious: Why would this change fix anything? Perhaps I'm abit stupid right now, because I think I proposed this once myself. Still, I can't remember why it should solve anything...

fuzzybear3965:
Yeah, I feel stupid for it having had taken me so long to identify as the solution. Okay. the key is this: LaTeX generates the file in the log based on how the path is specified on the command line. It just copies the path from the command line the LaTeX file to the log file. Thus, if you supply it a relative path, the quickfix window will use this relative path, which is bad if your working directory in vim is not the same as the main file. If you supply it an absolute path, then you avoid this problem. Your vim working directory can be whatever you'd like and you'll still uniquely access the file that was compiled. This also has the advantage that if the user changes working directory after the compilation has completed they are still able to access the file using the quickfix window. If you try to generate a path relative to the working directory at the start of compilation you are not guaranteed that this will be the same after compilation when the user is using the quickfix window.

lervag:
Actually, I don't think it is necessary to do this. Also: For multi-file projects, we have the same problem for included files; the suggested fix only works on the main file.

Please see #421 for information about my latest commit where I think I've fixed the issue. I guess we can continue the discussion there.

fuzzybear3965:
Your fix is good. Don't accept this pull request.

PR #442, opened 2016-04-23
Add '@tex' to Skim 'g:vimtex_view_general_options'

aldur:
I couldn't get forward search to work without the `@tex` flag.
Then, I found the solution in issue #204, I added it to the rc file and all went well.
So it may be worth adding it to the docs.


lervag:
Thanks!

PR #452, opened 2016-05-03
Check g:tex_fast before enabling additional highlighting.

manu-mannattil:
The default syntax highlighting script shipped with Vim defines the global variable `g:tex_fast` which allows one to selectively enable/disable highlighting for certain regions.  If it's set and does not contain the character `b`, bold/italic syntax highlighting should be disabled.  Similarly, if it doesn't contain the character `r`, texRefZone syntax highlighting should be disabled.

### DIFF
--- a/after/syntax/tex.vim
+++ b/after/syntax/tex.vim
@@ -22,21 +22,23 @@ syntax match texInputFile /\\includepdf\%(\[.\{-}\]\)\=\s*{.\{-}}/
 
 " {{{1 Italic font, bold font and conceals
 
-let conceal = (has('conceal') && get(g:, 'tex_conceal', 'b') =~# 'b')
-      \ ? 'concealends' : ''
+if get(g:, 'tex_fast', 'b') =~# 'b'
+  let conceal = (has('conceal') && get(g:, 'tex_conceal', 'b') =~# 'b')
+        \ ? 'concealends' : ''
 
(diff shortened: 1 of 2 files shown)


lervag:
Sorry for not responding faster. I will not accept this immediately. I think it requires an update to the docs as well, in particular, the `tex_fast` variable should be mentioned and explained (somewhat) under the section `SYNTAX HIGHLIGHTING`.

lervag:
By the way, do these optimizations really give _any_ performance benefit?

manu-mannattil:
Thanks for reviewing the pull request!  I have updated the documentation to mention the changes.  I do see some improvements in performance (especially with long wrapped lines), when syntax highlighting is disabled for certain groups.

lervag:
Thanks!

Do you see this improvement also for the stuff added by vimtex? I guess this is a nice addition anyways, to make things behave consistent with the internal plugin. However, I don't really expect my added rules to to give any performance reductions (that is, not these, at least - the nested syntax additions may affect performance).

manu-mannattil:
I haven't noticed any performance degradation due to the additional highlighting added by vimtex.  Like you said, I thought it'd be nice if things remained consistent with the default ft-tex-plugin.

lervag:
Ok. I agree that this is a good addition. Thanks again!

PR #460, opened 2016-05-29
Improved syntax highlighting for bib files

danielbmarques:
This adds syntax highlighting for all the entry types and field types
used by biblatex.

### DIFF
--- a/after/syntax/bib.vim
+++ b/after/syntax/bib.vim
@@ -1,0 +1,36 @@
+syn keyword bibType contained   article book mvbook inbook bookinbook suppbook booklet collection
+syn keyword bibType contained   mvcollection incollection suppcollection manual misc online patent
+syn keyword bibType contained   periodical suppperiodical proceedings mvproceedings inproceedings
+syn keyword bibType contained   reference mvreference inreference report set thesis unpublished
+syn keyword bibType contained   xdata customa customb customc customd custome customf conference
+syn keyword bibType contained   electronic masterthesis phdthesis techreport www artwork audio
+syn keyword bibType contained   bibnote commentary image jurisdiction legislation legal letter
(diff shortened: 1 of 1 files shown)


lervag:
Interesting. However, I think this would be better to include in the internal `bib.vim` syntax file, see `:e $VIMRUNTIME/syntax/bib.vim`. At least, I propose that you email Bernd Feige first. If he does not want to include your updates, I will consider to add it to vimtex.

lervag:
By the way, I would be happy if you could squash the commits. You can do this with

```sh
git rebase -i
git push -f
```

In the rebase, use the `fixup` option on the last commit (or the `squash` if you want to change the commit message).

danielbmarques:
Ok, I wrote to him, I'll let you know what he says.

Also I tried the squashing thing but couldn't work it out, I'm still getting the hang of git, sorry.
Is this what I was supposed to do?

captura de tela 2016-05-29 as 19 40 42
captura de tela 2016-05-29 as 19 41 20

I tried some other variations, but it seems it didn't work.

lervag:
Sorry, I should have written that you should do:

```sh
git rebase -i 257d2f9
```

Since you've already pushed the branch, it will not automatically define which commits to rebase, thus you have to specify it. 257d2f9 is the hash for the corresponding master branch, and so this will start an interactive rebase session for all commits in your branch. Now you can chosse to "fixup" the typo commit.

danielbmarques:
Ah, ok, now I got it, thanks. :)
But Bernd Feige just answered me and he is going to add the new keywords to the internal file, so I guess we can close this anyway.

lervag:
No problem! And I'm happy to hear Bernd Feige accepted the changes. I think this is a more appropriate solution that will benefit _all_ LaTeX users, regardless of whether or not they use vimtex.

Thanks!

PR #462, opened 2016-05-31
Use full list of arguments (line, pdf, and tex file (current buffer)) in the latexmk callback function for Skim.

rionda:
This allows to correctly place the reading mark at the current line when
dealing with multifile projects. Beforehand, the initial placement was
wrong (for multifile projects).

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -840,16 +840,17 @@ Options~
               if !a:status | return | endif
 
               let l:out = b:vimtex.out()
+              let l:tex = expand('%:p')
               let l:cmd = [g:vimtex_view_general_viewer, '-r']
               if !empty(system('pgrep Skim'))
-                call extend(l:cmd, ['-g'])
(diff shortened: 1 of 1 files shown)


doc/vimtex.txt
```diff
@@ -840,16 +840,17 @@ Options~
               if !a:status | return | endif
 
               let l:out = b:vimtex.out()
+              let l:tex = expand('%:p')
```
lervag: This should rather be:

```vim
let l:tex = b:vimtex.tex
```

lervag:
See my comment. Except for that, this seems like a good change.

rionda:
@lervag I thought so too, about the use of b:vimtex.tex, but it seems that b:vimtex.tex contains the "main tex file", not the current tex file, which may be different in a multifile project. I actually tried with "b:vimtex.tex" but it didn't work for me.

lervag:
Ah, sorry, I'm wrong. One should of course pass the local tex file, not the main tex file.

Thanks!

rionda:
Thank you!

PR #466, opened 2016-06-03
Don't add line number, pdf, and (current) tex file twice to the list of options. 

rionda:
We already do it when calling the command.


lervag:
Thanks!

PR #471, opened 2016-06-09
Order candicates in graphix autocompletion for better experience

chengs:
This commit is to enhance the experience in the omni completion, for including graphics.
The candidates are ranked by a:regex, so that a file that matches the input has a priority in the completion list.

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -362,12 +362,20 @@ function! s:img.complete(regex) dict " {{{2
   call filter(self.candidates, 'v:val !=# l:output')
 
   call map(self.candidates, 'strpart(v:val, len(b:vimtex.root)+1)')
-  call map(self.candidates, '{
-        \ ''abbr'' : v:val,
-        \ ''word'' : v:val,
-        \ ''menu'' : '' [graphics]'',
-        \ }')
+  let res = []
+  for fp in self.candidates
+    let cand = {
+        \ 'abbr' : fp,
+        \ 'word' : fp,
+        \ 'menu' : '[graphics]',
+        \}
+    if fp =~ a:regex
+      let cand.rank = 99
+    endif
+    call add(res, cand)
+  endfor
 
+  let self.candidates = res
   if g:vimtex_complete_img_use_tail
     for l:cand in self.candidates
       let l:cand.word = fnamemodify(l:cand.word, ':t')


lervag:
I have not heard of the `rank` attribute before. It does not seem to be documented (e.g. in `:h complete-func`). In any case, I think the changes may be simplified greatly to simply adding one line:

```vim
  call map(self.candidates, '{
        \ ''abbr'' : v:val,
        \ ''word'' : v:val,
        \ ''menu'' : '' [graphics]'',
        \ ''rank'' : v:val =~# a:regex ? 99 : 0,
        \ }')
```

That is, the `rank` may be added directly during the map call. No need to use the for loop. However, I am not quite sure I understand what this does. I would be happy to see some relevant documentation.

chengs:
You're right.
This rank attribute is only designed in NeoComplete.
I check the vim's omni focus itself and do not find the rank attribute.


```
------------------------------------------------------------------------------
CANDIDATE ATTRIBUTES			*neocomplete-candidate-attributes*

				*neocomplete-candidate-attribute-name*
word		String		(Required)
		The completion word of a candidate. It is used for matching
		inputs.

				*neocomplete-candidate-attribute-abbr*
abbr		String		(Optional)
		The abbreviation of a candidate. It is displayed in popup
		window. It is omitted by |g:neocomplete#max_keyword_width|.

				*neocomplete-candidate-attribute-kind*
kind		String		(Optional)
		The kind of a candidate. It is displayed in popup window.

				*neocomplete-candidate-attribute-menu*
menu		String		(Optional)
		The menu information of a candidate. It is displayed in popup
		window.

				*neocomplete-candidate-attribute-info*
info		String		(Optional)
		The preview information of a candidate. If 'completeopt'
		contains "preview", it will be displayed in |preview-window|.

				*neocomplete-candidate-attribute-rank*
rank		Number		(Optional)
		The completion priority.
```

lervag:
Thanks. It remains to understand why we should do this change. You want to give priority to the candidates that match the currently written text (or regexp).

If I understand neocomplete correctly, it will call the omni completion function automatically when we write e.g. `\includegraphics{`. This will generate a full list of candidates. when you continue to type, the omni completion plugin will not be called anymore, instead, neocomplete will sort and update the completion candidates. Thus, I would say that instead of adding a priority, one should instead remove all candidates that does not match the regexp. I just pushed this now, because I realized that this was missing from several of the completion types.

Please test the latest version and see if this improves things for you.

chengs:
@lervag 

The lastest version works perfectly.
Your implementation is better than mine.
I think it it time to close this pull request.

lervag:
Great, happy to hear it. Thanks for noticing the bad behaviour and thus fixing the completion bugs!

PR #472, opened 2016-06-11
Add option vimtex_fold_section_character for custom fold characters

wikimatze:
I've discovered the `g:vimtex_fold_enabled` while preparing my talk about your plugin. 

I liked the folding symbols from <https://github.com/matze/vim-tex-fold>, so I ported it here.

The result of setting `let g:vimtex_fold_section_character = '➜'` in my `vimrc`:

vimtex_custom_folding_character

I've used vader for testing but I'm not quite sure how to run the tests, the README has no contribution guide.

Cheers,
Matthias

### DIFF
--- a/autoload/vimtex/fold.vim
+++ b/autoload/vimtex/fold.vim
@@ -8,6 +8,7 @@ function! vimtex#fold#init_options() " {{{1
   call vimtex#util#set_default('g:vimtex_fold_enabled', 0)
   call vimtex#util#set_default('g:vimtex_fold_manual', 0)
   call vimtex#util#set_default('g:vimtex_fold_comments', 0)
+  call vimtex#util#set_default('g:vimtex_fold_section_character', 0)
   call vimtex#util#set_default('g:vimtex_fold_preamble', 1)
   call vimtex#util#set_default('g:vimtex_fold_envs', 1)
   call vimtex#util#set_default('g:vimtex_fold_parts',
@@ -234,7 +235,12 @@ endfunction
 function! vimtex#fold#text() " {{{1
   " Initialize
   let line = getline(v:foldstart)
-  let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . '*' : ''
+
+  if type(g:vimtex_fold_section_character)
+    let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . g:vimtex_fold_section_character : g:vimtex_fold_section_character
+  else
+    let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . '*' : ''
+  endif
   let title = 'Not defined'
   let nt = 73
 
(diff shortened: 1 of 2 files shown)


lervag:
Thanks, I guess we can add this. However, I think it needs some more work. I'll address with some comments to the code.

Note: I have just "recently" started to use Vader for testing. When you submit a pull request, travis will automatically test your code with the existing set of tests. I have not really considered explaining how this is done to contributors, because I think it makes it too difficult to contribute. I can worry about testing, and other people can submit ideas and suggestions for improvements. At least, that's my current idea.

autoload/vimtex/fold.vim
```diff
@@ -234,7 +235,12 @@ endfunction
 function! vimtex#fold#text() " {{{1
   " Initialize
   let line = getline(v:foldstart)
-  let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . '*' : ''
+
+  if type(g:vimtex_fold_section_character)
+    let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . g:vimtex_fold_section_character : g:vimtex_fold_section_character
+  else
+    let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . '*' : ''
```
lervag: This line is really exactly the same as 240, except that in 240 you _always_ use the fold section character. If we could agree on either of the styles, then we could simplify these five lines into a single line. I think that would be good.

So, personally, I find it is nice to only indicate only foldlevels larger than one, but it seems you prefer that also the first foldlevel has a fold character. Do you have a good reason why this would be better?

autoload/vimtex/fold.vim
```diff
@@ -234,7 +235,12 @@ endfunction
 function! vimtex#fold#text() " {{{1
   " Initialize
   let line = getline(v:foldstart)
-  let level = v:foldlevel > 1 ? repeat('-', v:foldlevel-2) . '*' : ''
+
+  if type(g:vimtex_fold_section_character)
```
lervag: I would instead use `empty(g:vimtex_...)`, and then consistently use a string for the variable. The default value should then either be `''` or `'*'`, depending on the result of the other discussion.

wikimatze: Agree!

autoload/vimtex/fold.vim
wikimatze: I'm used to it because of [vim-tex-fold](https://github.com/matze/vim-tex-fold) - so it's only a personal matter. I will remove the duplication from the code!

wikimatze:
Yeah, I have seen travis in action  - a great way to be more open for PRs on vim-plugins. Agree with your way of adding tests on your own for new functionality.

lervag:
Thanks. I made a minor change.

lervag:
See ec882ca80254c91ee005b204b80fb6029e1891a8 (seems I used the wrong issue reference (#474 instead of #472).

lervag:
I wanted to ask if we could rename the option, for instance to `g:vimtex_fold_levelmarker`. The marker character is used for every fold, not just the sections. Thus `section_character` seems a little bit off.

wikimatze:
Done in #476 (yeah, another PR :smile:)

lervag:
Great, thanks!

Issue #480, opened 2016-06-20
Continuous compiling can be started multiple times if relative paths are used in the root specifier


mbrunnen:
When I start compiling in continous mode I am able to specify the root file, but it doesn't recognise that there's already a compilation process running. 

When I start `:VimtexCompile` in `main.tex` and then switch to `subfolder/section.tex` and run the command `:VimtexCompile` again, it finds the correct root file, but it starts a new compiling process, instead of stopping the old one. I can only start and stop the compilation in the file, where I started the corresponding process.

Unfortunately I am forced to use Windows 7 with gvim.

I have two files:
`main.tex`:
```
\documentclass{article}

\title{Test file.}
\author{Tester Testington}

\begin{document}
\maketitle
\input{subfolder/section.tex}
\end{document}
```

and `subfolder/section.tex`:
```
%!TEX root = ../main.tex
\section{Test with another file in a subfolder}
Lorem ipsum...
```

my `vimrc` looks like this:
```
set nocompatible
set shellslash

" Load Vimtex
let &rtp  = $HOME . '\vimfiles\bundle\vimtex,' . &rtp
let &rtp .= ',' . $HOME . '\vimfiles\bundle\vimtex\after'

" Load other plugins, if necessary
" let &rtp = '~/path/to/other/plugin,' . &rtp

filetype plugin indent on
syntax enable

" Vimtex options go here
let g:tex_flavor='latex'
let g:vimtex_latexmk_continuous=1
let g:vimtex_latexmk_background=0
let g:vimtex_mappings_enabled=1
let g:vimtex_latexmk_build_dir='build'
let g:vimtex_view_general_viewer = 'SumatraPDF'
```

lervag:
I think the latest patch should fix this, can you try? I am not quite sure, though. What I did was to add a `fnamemodify` to the `getcwd`, as I am thinking that `getcwd` might have returned a relative path. I have tried to consistently use absolute paths in vimtex, which should not really require any resolves as you have suggested in #481.

mbrunnen:
I'm at 304b760624f856f95d3b85edc1a18dbe92443574, which is now the newest push, but the problem still exists. Actually I don't see any changes in the corresponding function `get_main_from_specifier` in file `vimtex.vim`. Latest change here was 7.6.2016. Maybe you didn't push?

lervag:
I was sure I pushed this, but obviously I was wrong. Sorry. Try now.

mbrunnen:
[...]

PR #481, opened 2016-06-20
fix for relative paths in root specifier for issue #480

mbrunnen:
This is a dirty fix for issue #480, which worked for me.


lervag:
See the comment in #480.

(Btw: I really appreciate that you've taken the time to both fix and open a pull request, even though I might have solved it in another way.)

PR #482, opened 2016-06-22
fix #480: Use absolute paths, pt. 2

### LINKED ISSUE #480: Continuous compiling can be started multiple times if relative paths are used in the root specifier

mbrunnen:
The string in the specifier was not part of `fnamemodify`, so no effect. Also in my case the first string of the array was used, so my `b:vimtex.tex` was still "C:/path/to/main/subfolder/../main.tex".

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -541,8 +541,8 @@ function! s:get_main_from_specifier(spec) " {{{1
         if filereadable(l:filename) | return l:filename | endif
       else
         for l:candidate in [
-              \ expand('%:p:h')             . '/' . l:filename,
-              \ fnamemodify(getcwd(), ':p') . '/' . l:filename
+              \ fnamemodify(expand('%:p:h') . '/' . l:filename, ':p'),
+              \ fnamemodify(getcwd() . '/' . l:filename, ':p')
               \]
           if filereadable(l:candidate) | return l:candidate | endif
         endfor


lervag:
No, I don't really think this is the right solution. First, there is no need to include the `'/' . l:filename` in the `fnamemodify`, since the entire point of that is to ensure that the path is absolute and not relative. You say the path becomes `C:/path/to/main/subfolder/../main.tex`, which seems to be "correct" but too complex. It seems more appropriate to use `simplify`, but I am not entirely convinced that should be necessary either.

I'll continue this in the issue #480.

lervag:
I just pushed a new update that should fix the issue.

PR #487, opened 2016-06-28
Drop -pdf as the default option to pass to latexmk.

midchildan:
Hi. First, thanks for making this awesome plugin.

This commit drops the `-pdf` option from the defaults for `g:vimtex_latexmk_options` . I made this change because it would be better to specify the kind of output you want in your latexmkrc file instead of your vim configuration file . This is important for some people in non-English speaking countries where people might use two different variants of TeX, one for their native language and the other for English.

In my case, I use a TeX engine called p-TeX, which is widely used in Japan. Unlike most TeX engines widely used for creating English documents, p-TeX doesn't directly output PDF. I have a problem with the current default because it overrides the option I specified in the latexmkrc file, and tries to output PDF directly. So if I want to have different kinds of output for each document depending on the language, I would have to change my vim configuration each time, which is a little tedious. 

Additionally, dropping `-pdf` from the defaults would make the output of `VimtexCompile` more consistent with what you would get when you type `$ latexmk` on the command line, leading to fewer surprises.

### DIFF
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -424,7 +424,7 @@ function! s:latexmk_build_cmd() " {{{1
   if !empty(g:vimtex_latexmk_options)
     let cmd .= ' ' . g:vimtex_latexmk_options
   else
-    let cmd .= ' -verbose -pdf -file-line-error'
+    let cmd .= ' -verbose -file-line-error'
     let cmd .= ' -synctex=1 -interaction=nonstopmode'
   endif
 
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -665,7 +665,6 @@ Options~
 *g:vimtex_latexmk_options*
   By default, |vimtex| passes the following options to `latexmk`: >
 
-     -pdf                     -- Compile to pdf documents
      -verbose                 -- More verbose output
      -file-line-error         -- Enable file:line:error style messages
      -synctex=1               -- Generate synctex data


lervag:
Ok, sorry this is taking so long. I am considering this, and I am not quite sure what is the best approach here. I agree that it seems sensible to be able to remove the `-pdf`. But at the moment, the options will "just work" for most people. If I remove `-pdf`, then it will not.

Thus, this is a change this might break things for people who don't have custom `.latexmkrc` files. And so, it is not something I will accept without discussion and before I understand how to deal with it correctly.

lervag:
I'm curious, is it not enough that you simply set the following option:

```vim
    let g:vimtex_latexmk_options = ' -verbose -file-line-error'
      \ ' -synctex=1 -interaction=nonstopmode'
```

This will remove `-pdf` from latexmk options for you, allowing the kind of behaviour that you want.

midchildan:
Maybe another option would be to make a new option like `g:vimtex_latexmk_output` that defaults to `-pdf` . That way, you can both preserve the default setting while allowing more flexibility by setting the new option to an empty string.

lervag:
No, you don't need to use different options for vimtex every time. Instead, use what I suggested in combination with a `~/.latexmkrc` file with `$pdf_mode = 1` to turn on pdf mode by default. Then use project specific `.latexmkrc` files that override the one in `$HOME` whenever necessary.

midchildan:
Oh, I guess that would work. I somehow haven't thought about that.

lervag:
Do you agree that it solves the problem, and that it is not necessary to remove the `-pdf` part from the default options?

midchildan:
Ok. Thanks!

lervag:
Great! And no problem. I'm all for improvement, but I won't change things unless convinced it is for the better. Don't hesitate to open new issues or pull requests if you have other problems/thoughts.

PR #499, opened 2016-07-22
Match \cite* and \Cite

danielbmarques:
Those two commands still weren't highlighted properly, now they are.


lervag:
You're right, thanks.

Issue #454, opened 2016-05-15
zathura synctex feature broken on version 0.3.6 for Ubuntu 14.04.3 LTS


wikimatze:
Based on #338 I installed version 0.3.3 and forward and backward searching is working fine. I'm running 14.04.3 LTS, Trusty Tahr and as soon as I switched to version 0.3.6 zathura will not open when I press `\ll`. I don't 

Does anybody know what I have to do to get it running?

lervag:
I'm using Zathura version 0.3.6, and it works mostly as expected. Both backward and forward search works.

Could you please follow the guidelines and add a complete description of how to reproduce your problem?

wikimatze:
My `vimrc`:


```
set nocompatible " Be iMproved

let g:neosnippet#expand_word_boundary = 1
let g:plug_threads = 40

call plug#begin('~/.vim/plugged/') 

Plug 'Shougo/vimproc.vim'
Plug 'lervag/vimtex'

call plug#end()

let g:latex_view_general_viewer = 'zathura'
let g:vimtex_view_method = 'zathura'
```

Tex-File `test.tex`:

```text
\documentclass{article}
\title{Cartesian closed categories and the price of eggs}
\author{Jane Doe}
\date{September 1994}
\begin{document}
   \maketitle
   Hello world!
   \pagebreak
   b
   \pagebreak
   c
\end{document}
```

My `~/.latexmkrc`:

```sh
# always create a pdf
$pdf_mode = 1;

# how pdflatex will be executed
$pdflatex = 'pdflatex --shell-escape -synctex=1 %O %S';
```

And the version:

```sh
wm@wm» latexmk --version

Latexmk, John Collins, 22 April 2016. Version 4.45
```

Running the tex-file with: `vim --servername vim test.tex`. Pressing `\lo` gives me `vimtex: No output exists`. Pressing `\li` give me:


```sh
b:vimtex : test                                                                                                                                                                               
  pid : 0                                                                                                                                                                                     
  root : '/home/wm/bla'                                                                                                                                                                       
  aux : '/home/wm/bla/test.aux'                                                                                                                                                               
  log : '/home/wm/bla/test.log'                                                                                                                                                               
  out : '/home/wm/bla/test.pdf'                                                                                                                                                               
  tex : '/home/wm/bla/test.tex'                                                                                                                                                               
  base : 'test.tex'                                                                                                                                                                           
  viewer                                                                                                                                                                                      
[...]

PR #504, opened 2016-07-28
Change order of arguments in s:zathura.start

chmue:
Hi Karl

### Explain the issue

I am trying to set up reverse synctex support in the zathura PDF viewer. Specifically, I would like to use `g:vimtex_view_zathura_options` to supply a custom `-x` option to zathura.

Currently, vimtex essentially hard-codes this option because the vimtex default value is added to the command line that is executed to start zathura _after_ `g:vimtex_view_zathura_options`. This overrides anything specified for `-x` via the vimtex option. You hinted at the problem in #454 [(specific comment)](https://github.com/lervag/vimtex/issues/454#issuecomment-219425818).

#### Steps to reproduce

1. Supply the `-x`/`--synctex-editor-command` command line option via `g:vimtex_view_zathura_options`, e.g. the following test setup:

    ```vim
    let g:vimtex_view_method = 'zathura'
    let g:vimtex_view_zathura_options = '-x firefox'
    ```

2. Open the viewer from vimtex by hitting `lv`.
NB: close all open instances of zathura such that a new instance is launched by vimtex.

3. Hit `Shit-left mouse click` in the opened PDF to have zathura do the reverse lookup.

#### Expected behaviour

Firefox opens.

#### Observed behaviour

Depending on whether you use vim or neovim, the file opens or nothing happens.

### Proposed fix

The fix I propose changes the order in which the command line for starting zathura is built. This allows definitions of `-x` supplied via `g:vimtex_view_zathura_options` to override the vimtex default. At the same time, the behaviour of vimtex should not change for all users who do not specifically include a `-x` in their definition of `g:vimtex_view_zathura_options`.

Just FYI, this is the setup I actually use and that works well after the small change to `s:zathura.start`:

```vim
let g:vimtex_view_method = 'zathura'
let g:vimtex_view_zathura_options = '-x '''
    \ . 'nvr --servername "' . v:servername . '"'
    \ . ' --remote-send "<C-\><C-n>:buffer \"\%{input}\"<CR>:\%{line}<CR>:normal\! ztzv<CR>"'''
```

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -346,10 +346,11 @@ endfunction
 " }}}2
 function! s:zathura.start(outfile) dict " {{{2
   let exe = {}
-  let exe.cmd  = 'zathura ' .  g:vimtex_view_zathura_options
+  let exe.cmd  = 'zathura '
   let exe.cmd .= ' -x "' . g:vimtex_latexmk_progname
         \ . ' --servername ' . v:servername
         \ . ' --remote +\%{line} \%{input}"'
+  let exe.cmd .= ' ' .  g:vimtex_view_zathura_options
   let exe.cmd .= ' ' . vimtex#util#shellescape(a:outfile)
   call vimtex#util#execute(exe)
   let self.cmd_start = exe.cmd


lervag:
Thanks, it seems like a simple and efficient change.

Issue #514, opened 2016-08-07
Nested syntax highlighting for Gnuplot (gnuplottex package)


matthewjohnrussell:
Would it be difficult to add syntax highlighting for nested Gnuplot commands that one puts inside a `gnuplottex` environment?

    \begin{gnuplottex}[terminal=..., terminaloptions=...]
        ...
    \end{gnuplottex}

It seems that in general, it is possible to enable such nested highlighting (many other vim plugins, e.g. for web frameworks with HTML+other language templates). I was able to modify [this](http://vim.wikia.com/wiki/Different_syntax_highlighting_within_regions_of_a_file) to enable Gnuplot highlighting in a plain text file, but didn't have time to make it work with LaTeX.

lervag:
Done. It seems to work pretty well, but please test and let me know if there is something missing.

PR #515, opened 2016-08-07
Use correct gnuplot environment for nested syntax (#514)

matthewjohnrussell:
Thanks for adding the nested Gnuplot syntax. I've only changed the environment from `gnuplottex` to `gnuplot`, which is what the `gnuplottex` Latex package uses.


lervag:
Thanks!

Issue #433, opened 2016-04-11
[Feature] Recognize `%! TEX root = main.tex` in bibfile to enable vimtex-toc


kiryph:
I like the feature `vimtex-toc` to jump to the bibfile (`,lt`). However, I miss the possibility to open the `vimtex-toc` from the bibtex file. Is it possible to recognize the corresponding tex file via `%! TEX root = main.tex` line or alternatively by the `latexmain` extension in same or above folder to show the table of contents? (BTW: comments in bibtex files do not have to have any leading symbol, cp e.g. http://tex.stackexchange.com/questions/93972/comment-out-sections-of-text-in-bib-file)

lervag:
If you open a bib file through the table of contents, then the table of contents will be available from the bib file as well. I think this makes sense, because a bib file can be used for many projects, and as such, it should not belong to any one project. Thus, if you open the bib file through other means, then one can not really know where the bib file belongs. But if you open it through the table of contents, then the current update "remembers" the TOC and allows to reopen it.

lervag:
Oh, and btw: the only support for bibfiles provided by vimtex is currently a custom indent function. Syntax highlighting for bib files come from Bernd Feige's syntax plugin that comes shipped with Vim, see `:e $VIMRUNTIME/syntax/bib.vim`. I propose that you contact Bernd Feige if you have comments about syntax highlighting issues.

PR #520, opened 2016-08-10
Add warning if .bib file hasn't been opened from .tex file

benutzer193:
I am not completely satisfied with the PR, because the keybindings stay if you open a .tex document after opening the .bib file and open the bib file from that .tex document without closing the buffer.

As an alternative I tried adding

```
if exists('g:vimtex_data[get(b:, "vimtex_id")].root')
  finish
endif
```

before

```
if exists('b:did_ftplugin')
  finish
endif
let b:did_ftplugin = 1
```
(I could add the keybinding warnings this way as well of course...)

but it seems bib.vim is only parsed once, when the buffer opens.
Do you have a idea to fix this issue or  do you think it doesn't matter?

### DIFF
--- a/autoload/vimtex/toc.vim
+++ b/autoload/vimtex/toc.vim
@@ -117,6 +117,11 @@ endfunction
 " }}}1
 
 function! vimtex#toc#open() " {{{1
+  let l:bib_warning = 'In order to use the ToC the .bib file has to be opened from within a .tex document!'
+  if !exists('g:vimtex_data[get(b:, "vimtex_id")].root')
+    call vimtex#echo#warning(bib_warning)
+    return
+  endif
   if vimtex#index#open(s:name) | return | endif
 
   if !exists('b:vimtex')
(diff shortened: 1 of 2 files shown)


benutzer193:
eb201bdf61fbd1e673e56eb9f427052e1ceea1a1 fixes the problem mentioned in the first comment.

lervag:
Let's take a step back here. What is the expected behaviour?

I have previously decided to not do very much for bib files. People use such files in very different ways, and so it is not as easy to generalize, e.g. to detect any main tex files.

However, I did once add mappings and commands to open the TOC, because this will be available indirectly if it has been opened once before. As you have noticed, this is the current state of things: If you open a bib file through a TOC, or after the TOC has been opened at least once, then the TOC will open. If the TOC has not been opened, then it will not be available from any bib files. I think you are right that in this latter case, there should be a warning.

benutzer193:
yes, that's all this commit does, it checks if there is already a main tex file if not warns the user that the commands not work. Without this commit you will encounter an error and can't open the toc.

I already saw #433 and your opinion on .bib files. I think that's completely fine and that's why I just added the warning. 

Perhaps I'll change the behavior locally that it parses for a %! TEX root line if there is no  main file yet, because I only add this line to bib files intended for a specific .tex document.

lervag:
The error is more involved. I'm working on a better fix and adding a warning. Pusing soon.

lervag:
I think my latest updated should solve your issue.

benutzer193:
yes, works for me!

lervag:
Great :)

PR #573, opened 2016-09-27
Avoid doautocmd warning

blyoa:
If users don't define autocmd for VimtexEventCompileStarted or  VimtexEventCompileStopped,
Vim gives error message (and maybe also hit-enter) as follows.

ss


lervag:
Thanks, and sorry for forgetting this! :)

PR #579, opened 2016-09-29
Added a quick start, closes #574, thanks to @nasenatmer

### LINKED ISSUE #574: Document a getting started guide
This is an enhancement request.

 I think that a getting started paragraph, or a typical use case, in the README.md would be in order. Also, subparagraphs documenting some basic but crucial use cases, animated gifs? 

I think that at least there should be an explanation how to build the document. Now, in order to build I just execute `:VimtexCompile` upon vim startup and latexmk continuous build start. I do not feel though it is an intended way to use the plugin.

tdi:


lervag:
Thanks! I decided to do some minor modifications, and so I just added this manually.

Issue #571, opened 2016-09-26
Fold multiline [re]new command/environment[*]


kiryph:
I'd like to propose following extensions for folding of command definitions:

- \newcommand{\cmd}[nargs][optargdefault]{defn}
- \newcommand*{\cmd}[nargs][optargdefault]{defn}
- \renewcommand{\cmd}[nargs][optargdefault]{defn}
- \renewcommand*{\cmd}[nargs][optargdefault]{defn}

Fold here the second mandatory curly argument `{defn}` to `{...}` and `[optargdefault]` to `[...]` if present.

and definition of environments
- \newenvironment{env}[nargs][optargdefault]{begdefn}{enddefn}
- \newenvironment*{env}[nargs][optargdefault]{begdefn}{enddefn}
- \renewenvironment{env}[nargs][optargdefault]{begdefn}{enddefn}
- \renewenvironment*{env}[nargs][optargdefault]{begdefn}{enddefn}

Fold here the second and third mandatory curly argument `{begdefn}{enddefn}` to `{...}{...}` and `[optargdefault]` to `[...]` if present.

A few typical examples:
i)
```latex
\renewcommand{\marginpar}{%
  \marginnote%
}
```
fold to
```latex
\renewcommand{\marginpar}{...}
```
ii)
```latex
\newcommand*{\StoreCiteField}[3]{%
  \begingroup
    \global\let\StoreCiteField@Result\relax
    \citefield{#2}[StoreCiteField]{#3}%
  \endgroup
  \let#1\StoreCiteField@Result
}
```
fold to
```latex
\newcommand*{\StoreCiteField}[3]{...}
```
iii)
```latex
\newenvironment{theo}[1][]{%
  \stepcounter{theo}%
  \ifstrempty{#1}{%
    \mdfsetup{
      frametitle={%
      \tikz[baseline=(current bounding box.east),outer sep=0pt]%
        \node[anchor=east,rectangle,fill=blue!20] {\strut Theorem~\thetheo};%
      }
    }%
  }%
  {% else ifstrempty:
  \mdfsetup{
    frametitle={
    \tikz[baseline=(current bounding box.east),outer sep=0pt]%
      \node[anchor=east,rectangle,fill=blue!20]{\strut Theorem~\thetheo:~#1};}%
  }
  }%
  \mdfsetup{
    innertopmargin=10pt,
    linecolor=blue!20,
    linewidth=2pt,topline=true,
[...]

PR #588, opened 2016-10-06
Extend #571: Fold cmds following 3 patterns

kiryph:
with options to assign commands to these patterns


lervag:
For future reference: see #589.

PR #596, opened 2016-10-23
Make 'insert new command' work also after non-^ and non-\s regex

Aster89:
The insert new command functionality, mapped on <F7> by default, doesn't work when the word is preceded by non-space and non-start-of-line character. This commit allows insertion after any non-letter, since LaTeX commands cannot contain other charaters than a-zA-Z

[There's something wrong in the comparison of the two files. I changed only the line 108 from
`let l:re = '\v%(^|\s)\zs\w+\ze%(\s|$)'`
to
`let l:re = '\v%(^|\A)\zs\w+\ze%(\s|$)'`,
but in the comparison it seems that the line number is wrong (in fact line 60 is split in two in the comparison tool of GitHub. Don't know, but I think the commit is clear.]

### DIFF
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -105,7 +105,7 @@ function! vimtex#cmd#delete() " {{{1
 endfunction
 
 function! vimtex#cmd#create() " {{{1
-  let l:re = '\v%(^|\s)\zs\w+\ze%(\s|$)'
+  let l:re = '\v%(^|\A)\zs\w+\ze%(\s|$)'
   let l:c0 = col('.') - (mode() ==# 'i')
 
   let [l:l1, l:c1] = searchpos(l:re, 'bcn', line('.'))


lervag:
Thanks!

Aster89:
@lervag, I've thinking to another edit since the present commit.

It is about <F7>-mapped function while in insert mode. I use it very very frequently, and I find it far more useful than when in normal mode. Nevertheless, I'm thinking about modifiying its behavior when the cursor is in the middle of a word, for the following reason.
It happens, not rarely at all, that I want to highlight a word with `\emph`, `\textit`, or whatever.
Consider this scenario, `put THIS in emphasis|`, where I want to enclose `THIS` in `\emph{}`; I'm tempted to do `<Esc>3biemph<F7>`, but this clearly produces `put \emphTHIS{| in emphasis` instead of `put \emph{|THIS in emphasis`.

I hope I'm not bothering you with this thoughts.

lervag:
First, of course, no bother! But I would have preferred if you opened a new issue (or PR) for this.

However, I would think that this particular case should have a different command. I would think the following idea, starting from the beginning of your scenario: `<esc>3b<f5>emph<enter>`. Here `<f5>` would be a new normal mode mapping that would surround the word under the cursor with the given command.

Although, I also think that you have a point that the current `<f7>` insert mode mapping should work as you point out. I would be happy if you are able to create a pull request for this. If you do, I think it would be good to branch of the latest master branch.

Aster89:
Yes, I think that the f5 command you have devised would be very useful. It would be the fastest way to enclose a _single_ word in a command.

For what concerns enclosing more than a word, then I think it's not worth the effort of creating a command which prompts the insertion of the number of words after which the closing bracket has to be put. In this case I think the best choice is to retain the concept of your f7 command (i.e. open the command with the word under the cursor and let it opened, leaving to the user the responsibility to put the closing bracket where they want it to be) plus the modification that I've suggested and you approved.

Yes, I can do modifications on my own fork and then do a pull request. I'll do it as soon as I can. I hope/think I should find time enough to do it before the weekend.

Rmano:
@Aster89 , notice that with UltiSnips and this snippet: 

```
snippet _ "Latex emph" i
\emph{${1:${VISUAL}}} 
endsnippet
```

in your case I will do `viw` (visual select inner word), then `TAB`, `_`, and TAB again. It is really handy both correcting and writing on the first time (`_`,` TAB`, write text, `TAB` again and continue...)  


Notice that there is a space after the `\emph` line...

lervag:
I think @Rmano has a good idea here. Note also that you can get a pretty powerful feature from [sandwich.vim](https://github.com/machakann/vim-sandwich) that can give interactive surround commands, e.g. `saiwi...`, where `sa` is an operator to add surroundings, `iw` is the inner word text object, and `i` says to use an interactive surrounding where you type the former and latter surroundings.

Still, I think it is nice to have a normal and visual mode command to add a surrounding command.

lervag:
Let me know what you think. I added an `<f8>` mapping in normal and visual mode that asks for a user input command that the current word or visual selection will be surrounded by as a command.

PR #597, opened 2016-10-23
Feature/cmd folding

lervag:
This PR is based on #589 and includes some of the commits by @kiryph. It has been updated and implemented with new option names and a more functional approach.

It also includes some refactoring. It should probably have been committed in more logical commits, but what the heck...

### DIFF
--- a/autoload/vimtex/fold.vim
+++ b/autoload/vimtex/fold.vim
@@ -27,9 +27,36 @@ function! vimtex#fold#init_options() " {{{1
         \   'subsubsection',
         \ ])
   call vimtex#util#set_default('g:vimtex_fold_documentclass', 0)
-  call vimtex#util#set_default('g:vimtex_fold_usepackage', 1)
-  call vimtex#util#set_default('g:vimtex_fold_newcommands', 1)
   call vimtex#util#set_default('g:vimtex_fold_markers', 1)
+
+  " Fold command pattern 1: \<command>{<multi-line mandatory arg>}
+  "  folded to '\<command>{...}'
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern1', 1)
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern1_list',
+        \ [
+        \   'hypersetup',
+        \   'tikzset',
+        \ ])
+  " Fold command pattern 2: \<command>[<multi-line optional arg>]{<short mandatory arg>}
+  "  folded to '\<command>[...]{<short mandatory arg>}'
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern2', 1)
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern2_list',
+        \ [
+        \   'usepackage',
+        \   'includepdf',
+        \ ])
+  " Fold command pattern 3: \<command>{short mandatory arg}[]{}{\n}
+  "  folded to '\<command>{<short mandatory arg>} ...'
+  " NOTE: final '}' has to be on its own line!
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern3', 1)
+  call vimtex#util#set_default('g:vimtex_fold_cmd_pattern3_list',
+        \ [
+        \   '%(re)?new%(command|environment)',
+        \   'providecommand',
+        \   'presetkeys',
+        \   'Declare%(Multi|Auto)?CiteCommand',
+        \   'Declare%(Index)?%(Field|List|Name)%(Format|Alias)',
+        \ ])
 
   " Disable manual mode in vimdiff
   let g:vimtex_fold_manual = &diff ? 0 : g:vimtex_fold_manual
@@ -59,9 +86,10 @@ function! vimtex#fold#init_script() " {{{1
         \   '%(front|main|back)matter',
         \   'appendix',
         \   'part',
-        \   'usepackage',
-        \   '%(re)?new%(command|environment)',
-        \ ], '|') . ')'
+        \ ]
+        \ + g:vimtex_fold_cmd_pattern1_list
+        \ + g:vimtex_fold_cmd_pattern2_list
+        \ + g:vimtex_fold_cmd_pattern3_list, '|') . ')'
 endfunction
 
 " }}}1
@@ -130,7 +158,6 @@ function! s:foldmethod_in_modeline()
 endfunction
 
 " }}}1
-
 function! vimtex#fold#refresh(map) " {{{1
   setlocal foldmethod=expr
   execute 'normal! ' . a:map
@@ -167,27 +194,38 @@ function! vimtex#fold#level(lnum) " {{{1
     return '0'
   endif
 
-  " Fold usepackages
-  if g:vimtex_fold_usepackage
-    if line =~# '^\s*\\usepackage\s*\[\s*\%($\|%\)'
-      let s:usepackage = 1
+  " Fold command pattern 1 (e.g. hypersetup)
+  if g:vimtex_fold_cmd_pattern1
+    if line =~# '^\s*\\\%('.join(g:vimtex_fold_cmd_pattern1_list, '\|').'\)\s*{\s*\%($\|%\)'
+      let s:cmd_pattern1 = 1
       return 'a1'
-    elseif get(s:, 'usepackage', 0) && line =~# '^\s*\]{'
-      let s:usepackage = 0
+    elseif get(s:, 'cmd_pattern1', 0) && line =~# '^\s*}'
+      let s:cmd_pattern1 = 0
       return 's1'
     endif
   endif
 
-  " Fold newcommands (and similar)
-  if g:vimtex_fold_newcommands
-    if line =~# '\v^\s*\\%(re)?new%(command|environment)\*?'
-          \ && indent(a:lnum+1) > indent(a:lnum)
-      let s:newcommand_indent = indent(a:lnum)
+  " Fold command pattern 2 (e.g. usepackages)
+  if g:vimtex_fold_cmd_pattern2
+    if line =~# '^\s*\\\%('.join(g:vimtex_fold_cmd_pattern2_list, '\|').'\)\s*\[\s*\%($\|%\)'
+      let s:cmd_pattern2 = 1
       return 'a1'
-    elseif exists('s:newcommand_indent')
-          \ && indent(a:lnum) == s:newcommand_indent
+    elseif get(s:, 'cmd_pattern2', 0) && line =~# '^\s*\]{'
+      let s:cmd_pattern2 = 0
+      return 's1'
+    endif
+  endif
+
+  " Fold command pattern 3 (e.g. newcommands)
+  if g:vimtex_fold_cmd_pattern3
+    if line =~# '\v^\s*\\%('.join(g:vimtex_fold_cmd_pattern3_list, '|').')\*?'
+          \ && indent(a:lnum+1) > indent(a:lnum)
+      let s:cmd_pattern3 = indent(a:lnum)
+      return 'a1'
+    elseif exists('s:cmd_pattern3')
+          \ && indent(a:lnum) == s:cmd_pattern3
           \ && line =~# '^\s*}\s*$'
-      unlet s:newcommand_indent
+      unlet s:cmd_pattern3
       return 's1'
     endif
   endif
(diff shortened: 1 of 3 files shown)


kiryph:
Thanks for creating an updated pull request.

However, the current HEAD (47334e0) of the branch `Feature/cmd-folding` is broken:

- always folds preamble; the following settings do not have an affect:
   ```viml
let g:vimtex_fold_preamble = 0
let g:vimtex_fold_documentclass = 1
```
This works until c0093e3

- No usepackage folding despite`:echo g:vimtex_fold_cmd_single_opt` returning `['usepackage', 'includepdf']`  
This works until 2219ebe (my last commit)

**Comment by lervag** [comment]
I have probably pushed a little bit too quickly. I'll look into this as soon as possible.

Note, though: I've removed `g:vimtex_fold_documentclass`, because it was essentially exactly the same as the `usepackage` command. Thus, to get that behaviour, you can

```vim
let g:vimtex_fold_cmd_single_opt = ['usepackage', 'includepdf', 'documentclass']
```

At least: it should work. But as you say it might not work correctly now. I'll let you know when I've fixed it.

kiryph:
Minor doc update is also missing: You have added a list of associated settings under `:h vimtex-folding` which do not reflect the latest changes.

lervag:
Yes, you have to wait for the fix.

If you add 'documentclass' to the `single_opt` option, then the `preamble` option will be disabled and a warning will be displayed (unless you have also disabled the preamble folding explicitly).

Note, though, you can't do `+=`, because the `vimrc` file is read before default option is set. Thus you need to include the entire list.

kiryph:
I always struggle with the load order of vimrc w.r.t. plugins. What about moving the appending of `documentclass` to the list `g:vimtex_fold_cmd_single_opt` to `~/.vim/after/ftplugin/tex.vim`?

lervag:
Yes, that should work. I think. You could check: If it works, then the option should have the expected value...

kiryph:
That is true 😃 .

kiryph:
For clarification one could add to the documentation that a command can only be added to one of the lists:

- `g:vimtex_fold_cmd_single_opt`
- `g:vimtex_fold_cmd_single`
- `g:vimtex_fold_cmd_multi`

A possible further extension would be:

Commands can be added to `g:vimtex_fold_cmd_single_opt` and one of  
- `g:vimtex_fold_cmd_single`
- `g:vimtex_fold_cmd_multi`

This extends folding in following circumstances:

```
\cmd_single[
    option1,
    option2,
]{singl

}
```

folded to

`\hypersetup[...]{...}`

or

```
\cmd_multi[
    option1,
    option2,
]{specifier}{
      ...,
}
```

folded to

`\cmd_multi[...]{specfier} ...`

Right now, I do not know any use-case and therefore I do not see a reason to actually do this.

kiryph:
I had checked the plugin latex-suite for naming the variables for folding of latex commands.

In general latex-suite recognizes following four variables describing what to fold
   * `Tex_FoldedSections`  
       g:Tex_FoldedSections='part|addpart,chapter|addchap,section|addsec,subsection, subsubsection, paragraph, subparagraph'
   * `Tex_FoldedEnvironments`  
       g:Tex_FoldedEnvironments = 'verbatim,comment,eq,gather,align,figure,table,thebibliography,keywords,abstract,titlepage'
   * `Tex_FoldedCommands`  
       empty on default
   * `Tex_FoldedMisc`
     g:Tex_FoldedMisc='comments,preamble,item,<<<'

The documentation of version 1.9.0 (released September 2016) can be found here:
https://github.com/vim-latex/vim-latex/blob/v1.9.0/doc/latex-suite.txt#L1947
for command folding see in particular:
https://github.com/vim-latex/vim-latex/blob/v1.9.0/doc/latex-suite.txt#L2001

The only option `Tex_FoldedCommands` folds `\command{multi-line}`, i.e., it corresponds to `g:vimtex_fold_cmd_single`.

lervag:
I've located and squashed a couple of bugs. It seems to work as expected now. Could you test?

Thanks for the references to LaTeX-Suite. I'll look into it and see if I can find some improved names and/or ideas for improving the docs.

lervag:
I have an idea for how to improve this. How about instead of three options with lists, we use a single option dictionary, e.g.:

```vim
let g:vimtex_fold_commands = {
    \ 'hypersetup' : 'single',
    \ 'tikzset' : 'single',
    \ 'usepackage' : 'single_opt'
    \ 'includepdf' : 'single_opt'
    \ '%(re)?new%(command|environment)' : 'multi',
    \ 'providecommand' : 'multi',
    \ 'presetkeys' : 'multi',
    \ 'Declare%(Multi|Auto)?CiteCommand' : 'multi',
    \ 'Declare%(Index)?%(Field|List|Name)%(Format|Alias)' : 'multi',
    \}
```

Here the option key defines a command (or a regexp for commands), and the value defines the type of command folding.

kiryph:
I definitely like this.

I am still thinking about a general approach for all of them.
The current patterns are very similar:
- the line at the beginning consists of three parts  
   1. `backslash <command possibly with additional stuff for the foldtext>`,
   2. `opening bracket:either [ or { ` and
   3. `%\n` or `\n`
- the ending line consists of a single part `^\s*\(]|}\)\({specifier}\)?`

They are either folded to
 - `\<command possibly with additional stuff for the foldtext>[...]{specifier}` (single_opt),
 - `\<command possibly with additional stuff for the foldtext>{...}` (single), and
 - `\<command possibly with additional stuff for the foldtext> ...` (multi)

I try to keep in mind commands with an arbitrary number of optional and mandatory arguments, something like `\command[]{}[]{}...` and single argument to fold.  Two arguments folding which I had indicated in a previous comment is nothing I would pursue right now.

kiryph:
And yes, command folding works now.

kiryph:
I have checked also [folding in the emacs plugin auctex](http://git.savannah.gnu.org/cgit/auctex.git/tree/tex-fold.el?id=5232279373cf69147b04660026cde2d3b11b9f0c#n75):

```
(defcustom TeX-fold-macro-spec-list
  `(("[f]" ("footnote" "marginpar"))
    ("[c]" ("cite"))
    ("[l]" ("label"))
    ("[r]" ("ref" "pageref" "eqref"))
    ("[i]" ("index" "glossary"))
    ("[1]:||*" ("item"))
    ("..." ("dots"))
    ("(C)" ("copyright"))
    ("(R)" ("textregistered"))
    ("TM"  ("texttrademark"))
    (1 ("part" "chapter" "section" "subsection" "subsubsection"
	"paragraph" "subparagraph"
	"part*" "chapter*" "section*" "subsection*" "subsubsection*"
	"paragraph*" "subparagraph*"
	"emph" "textit" "textsl" "textmd" "textrm" "textsf" "texttt"
	"textbf" "textsc" "textup")))
  "List of replacement specifiers and macros to fold.

The first element of each item can be a string, an integer or a
function symbol.  The second element is a list of macros two fold
without the leading backslash.

If the first element is a string, it will be used as a display
replacement for the whole macro.  Numbers in braces, brackets,
parens or angle brackets will be replaced by the respective macro
argument.  For example \"{1}\" will be replaced by the first
mandatory argument of the macro.  One can also define
alternatives within the specifier which are used if an argument
is not found.  Alternatives are separated by \"||\".  They are
most useful with optional arguments.  As an example, the default
specifier for \\item is \"[1]:||*\" which means that if there is
an optional argument, its value is shown followed by a colon.  If
there is no optional argument, only an asterisk is used as the
display string.

If the first element is an integer, the macro will be replaced by
the respective macro argument.

If the first element is a function symbol, the function will be
called with all mandatory arguments of the macro and the result
of the function call will be used as a replacement for the macro.
```

There are some features not yet present in vimtex which could also be interesting:

For example, vim's conceal feature would allow inline folding, such as `\footnote{...}` to `[f]`.

The flexible approach with `integer`, `{integer}`, `[integer]` and even alternatives with `||` is something I would also evaluate for your dictionary suggestion.

lervag:
I agree that the patterns are very similar. However, I think it is very difficult to find a general solution. For instance, in which cases do we actually display the content of the first argument `{...}` (i.e. as for `single_opt`) and when should it be shortened to `{...}` or similar? Although it would be very nice with a good general solution, I think the current solution is close enough and works pretty well.

Of course, if you are able to find a solution that solves this in a more general solution, I would be very happy to consider a pull request!

In the meantime, I'll update the option style and begin to clean up the PR to make it ready for merge.

Note: I'm going to force push this branch while cleaning up. This means you might need some `git reset --hard` commands to load the updated branch for testing.

lervag:
Ok, now I've rebased the branch on top of master and cleaned up my commits. It should make it easier to see what is changed. I mostly preserved your commits (but squashed the one that fixed markers).

Let me know what you think. If you think the current version is OK, then we can merge.

kiryph:
I have tested the rebased branch (last commit is e419085). As you said I had to do a `$ git reset origin/feature/cmd-folding --hard`. I have one issue and minor notes about the documentation. 

- First of all, unfortunately documentclass folding does not work. My current configuration in my vimrc is:

   ```
   " Vimtex: {{{2
   let g:vimtex_fold_enabled = 1
   let g:vimtex_fold_preamble = 0
   autocmd FileType tex let g:vimtex_fold_commands.documentclass = 'single_opt'
   ```  
  After opening a tex file, `:echo g:vimtex_fold_commands` contains documentclass but does not work; it still folds the preamble despite having disabled it explicitly.
- two missing commas in the docs for the default value of `g:vimtex_folding_commands` (lines 506 and 507 containing `usepackage` and `includepdf`)
- I would add to the table of examples in the docs a column denoting the variant illustrated in the respective example

   ```
    Variant       Not folded                     Folded
    ======================================================================
    'single'      \hypersetup{                   \hypersetup{...}
                    option 1,
                    ...,
                    option n
                    }

    'single_opt'  \usepackage[                   \usepackage[...]{name}
                    option 1,
                    ...,
                    option n
                    ]{name}

    'multi'       \newcommand{\command}[3]{      \newcommand{\command} ...
                    Hello #1, #2, and #3.
                  }
   ```
- and finally I would add to the docs explicitly how to fold the options of the documentclass under `|vimtex-folding-commands|`; the comment in g:vimtex_fold_preamble is only a remark, but IMHO this is a feature which deserves a more prominent place with an explicit setup for the vimrc, such as
```
autocmd FileType tex let g:vimtex_fold_commands.documentclass = 'single_opt'
```
Another reason for documenting this is that the variable `g:vimtex_fold_preamble` is more complicated to grasp than other folding options; so examples can be helpful to clarify things.

Since this is working and a unifying regex might introduce new problems, even though it might simplify the definition of the variable `g:vimtex_fold_commands`, but on the other hand do not necessarily make it easier to understand how the folding works, I would go with this feature branch for command folding.

Thanks for your effort regarding this issue. I hope others find this useful as well.

lervag:
Thanks! I've updated accordingly and fixed the problems.

I also implemented a slightly different behaviour from other options in that by default, setting this option will append to the default values instead of ignoring them. Thus, you can do

```vim
let g:vimtex_fold_preamble = 0
let g:vimtex_fold_commands = {
    \ 'documentclass' : 'single_opt',
    \}
```

to add folding for the `\documentclass`.

Please test again. Note that I forced pushed again to keep the branch on top of master.

kiryph:
Mmh..., I do not like the special-casing to override the defaults with a special entry in the dictionary. I am quite sure that programming style guides would not recommend this. Sorry, that I seem to be a little bit nit-picky.

Are you sure that it is not possible to use features of vim itself to append to a list defined in an ft-plugin by specifying either an autocommand or using the after/ftplugin/tex.vim. If this is not possible, I actually would re-add all defaults to my vimrc or after/ftplugin/tex.vim file, even though your current proposal allows to append to the defaults or override the defaults. 

I guess there is a timing issue when the entry is appended to the dictionary in the event of opening a tex file and the foldexpr is evaluated. However, even pressing `zx` does not update the folds accordingly. I think I do not understand something here. Possibly your init_? function is only run once and is never called again.

lervag:
There are two reasons that it is currently not possible to append to the option after initialiation:

1. I use a combined regex `s:folded` to improve fold speed, which is defined under initialization.

2. The `fold_commands` option is used to create static "objects" that are used to fold.

Both of these fall under the `init_script` function that is only called once. You _can_ force reinitialization, but I think that should be avoided. Perosnally, I don't really mind that people have to copy the entire default dictionary in order to add to it. And I don't see a reason to allow flexibility in the sense of changing the fold command objects or the `s:folded` regex after initialization. Thus I propose we do one of the following:

1. Accept the current behaviour.

2. Use the more standard approach for settings, where one must copy the entire default value in order to add to it.

I have a feeling that we both agree that the standard approach is better, even though it will lead to more code for customizations. Still, explicit is better than implicit, right?

kiryph:
Thanks for the detailed explanation. I agree that there is no reason to allow flexibility of changing fold command objects after initialization.

However, I would not use the current behaviour. The special value 'option' would also be in clash with a latex command named `\option{}`. A quick check on ctan revealed that there is actually a package `options` which has a command `\option{}`. This should support the argument not to 'misuse' the dictionary for something else.

I have another proposal:

**Two dictionaries:**

  ```viml
let g:vimtex_fold_enabled = 1
let g:vimtex_fold_preamble = 0
let g:vimtex_fold_commands_user = { 'documentclass' : 'single_opt' }
" let g:vimtex_fold_commands_defaults = {} " uncommenting would remove the defaults
" or alternatively a boolean variable to disable appending the defaults
```

For comparison the configuration for the "standard approach" by copying entire defaults:

  ```viml
let g:vimtex_fold_enabled = 1
let g:vimtex_fold_preamble = 0
    let g:vimtex_fold_commands = {
        \ 'hypersetup' : 'single',
        \ 'tikzset' : 'single',
        \ 'usepackage' : 'single_opt',
        \ 'includepdf' : 'single_opt',
        \ 'documentclass' : 'single_opt',
        \ '%(re)?new%(command|environment)' : 'multi',
        \ 'providecommand' : 'multi',
        \ 'presetkeys' : 'multi',
        \ 'Declare%(Multi|Auto)?CiteCommand' : 'multi',
        \ 'Declare%(Index)?%(Field|List|Name)%(Format|Alias)' : 'multi',
        \}
```

I am happy with both of them but not with the current behaviour relying on the special entry `option` with a preference for my suggestion using the additional variable `g:vimtex_fold_commands_user`. The reason is that this would make it possible to use the defaults, also after updating vimtex without copying the defaults again. This would also make the vimrc shorter and remove the maintenance burden. Enabling documentclass folding does not mean the user wants to override the defaults.

I certainly agree with explicit is better than implicit.

lervag:
I like your proposal. I've updated accordingly. This brings us one step closer to merging...

kiryph:
I hope this issue is now almost finished. It has gotten a little bit too lengthy.

A minor fix has to be applied to the most recent changes: the docs are speaking about the variables 

   |g:vimtex_fold_commands|
   |g:vimtex_fold_commands_default|

but the code expects

   |g:vimtex_fold_commands_user|
   |g:vimtex_fold_commands_default|

I do not know which you prefer. The shortest fix is using the first ones:

```diff
❯ git diff
diff --git a/autoload/vimtex/fold.vim b/autoload/vimtex/fold.vim
index 3893fbc..ec7c729 100644
--- a/autoload/vimtex/fold.vim
+++ b/autoload/vimtex/fold.vim
@@ -54,7 +54,7 @@ function! vimtex#fold#init_script() " {{{1
   " Set up command fold structure
   "
   let s:cmds = extend(g:vimtex_fold_commands_default,
-        \ get(g:, 'vimtex_fold_commands_user', {}))
+        \ get(g:, 'vimtex_fold_commands', {}))
   let s:cmd_types = []
   let l:cmds_all = []
   for l:type in ['single', 'single_opt', 'multi']

```

lervag:
Thanks! I'm merging now. If there should be more issues with this, please don't hesitate to report them.

And thanks a lot for all your input! I appreciate it; the feature would not be as good if you had not contributed your ideas and bug fixes!

PR #600, opened 2016-10-25
mupdf: missing output_not_readable function

jbohren:
This should probably be put in a common place, but mupdf is missing error-free reverse-search support without it.


lervag:
Thanks! I agree it should be common, so I instead moved it to `view/common.vim`.

PR #607, opened 2016-10-27
Fixed typo in VimtexCountWords' help

mabako:
The command `VimtexCountWords`, `VimtexCountLetters` (and the ! variants) run `texcount -nosub -sum`, not `texcount -nosub -sub`. 
The latter would, incidentally, negate the usage of `-nosub`.


lervag:
Thanks!

PR #625, opened 2016-11-24
Better handling of relative subimports

DelusionalLogic:
The old way of handling sub imports didn't actually allow them to be
relative, which is half the power of the import package.

The new code correctly identifies all my subimported texfiles in
a pretty complicated latex project.

### DIFF
--- a/autoload/vimtex/parser.vim
+++ b/autoload/vimtex/parser.vim
@@ -120,19 +120,20 @@ endfunction
 function! s:input_line_parser_tex(line, file, re) " {{{1
   " Handle \space commands
   let l:file = substitute(a:line, '\\space\s*', ' ', 'g')
-  let l:subimport = 0
 
   " Handle import package commands
+  let l:subimport = 0
   if l:file =~# '\v\\%(sub)?%(import|%(input|include)from)'
     let l:candidate = s:input_line_parser_tex(
           \ substitute(l:file, '\\\w*\s*{[^{}]*}\s*', '', ''),
           \ a:file,
           \ '\v^\s*\{')
     if !empty(l:candidate) | return l:candidate | endif
-    let l:file = substitute(l:file, '}\s*{', '', 'g')
 
     " Handle relative paths
     let l:subimport = l:file =~# '\v\\sub%(import|%(input|include)from)'
+
+    let l:file = substitute(l:file, '}\s*{', '', 'g')
   endif
 
   " Parse file name
--- a/autoload/vimtex/parser.vim
+++ b/autoload/vimtex/parser.vim
@@ -120,6 +120,7 @@ endfunction
 function! s:input_line_parser_tex(line, file, re) " {{{1
   " Handle \space commands
   let l:file = substitute(a:line, '\\space\s*', ' ', 'g')
+  let l:subimport = 0
 
   " Handle import package commands
   if l:file =~# '\v\\%(sub)?%(import|%(input|include)from)'
@@ -128,8 +129,10 @@ function! s:input_line_parser_tex(line, file, re) " {{{1
           \ a:file,
           \ '\v^\s*\{')
     if !empty(l:candidate) | return l:candidate | endif
-
     let l:file = substitute(l:file, '}\s*{', '', 'g')
+
+    " Handle relative paths
+    let l:subimport = l:file =~# '\v\\sub%(import|%(input|include)from)'
   endif
 
   " Parse file name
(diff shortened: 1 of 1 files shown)


autoload/vimtex/parser.vim
```diff
@@ -128,8 +129,14 @@ function! s:input_line_parser_tex(line, file, re) " {{{1
...
     if !empty(l:candidate) | return l:candidate | endif
-
     let l:file = substitute(l:file, '}\s*{', '', 'g')
+
+    " Handle relative paths
+    if l:file =~# '\v\\sub%(import|%(input|include)from)'
+      let l:sub = 1
+    endif
+
+
```
lervag: These are two unnecessary empty lines. Please remove then.

autoload/vimtex/parser.vim
```diff
@@ -146,7 +153,11 @@ function! s:input_line_parser_tex(line, file, re) " {{{1
 
   " Use absolute paths
   if l:file !~# '\v^(\/|[A-Z]:)'
-    let l:file = b:vimtex.root . '/' . l:file
+    if l:sub
+      let l:file = fnamemodify(a:file, ":h") . "/" . l:file
+    else
+      let l:file = b:vimtex.root . '/' . l:file
+    endif
```
lervag: Use `':h'` instead of `":h"`. And I prefer the more concise ternary syntax, e.g.:

```vim
let l:file = (l:sub ? fnamemodify(a:file, ':h') : b:vimtex.root) . '/' . l:file
￼```

File: autoload/vimtex/parser.vim [hunk 3]
```diff
@@ -128,8 +129,14 @@ function! s:input_line_parser_tex(line, file, re) " {{{1
           \ a:file,
           \ '\v^\s*\{')
     if !empty(l:candidate) | return l:candidate | endif
-
     let l:file = substitute(l:file, '}\s*{', '', 'g')
+
+    " Handle relative paths
+    if l:file =~# '\v\\sub%(import|%(input|include)from)'
+      let l:sub = 1
+    endif
```
lervag: Instead of this, you can do:

```vim
let l:subimports = l:file =~# '\v\\sub%(import|%(input|include)from)'
```

I also propose to use `l:subimports` (or similar) instead of just `l:sub`.

lervag:
And: Thanks! I appreciate the effort of identifying the problem and suggesting a solution! :)

DelusionalLogic:
This is my first real endeavor in vimscript. I appreciate you taking the time to look at it, and to give me some pointers.

I made the changes you suggested.

I should probably note that the import package contains something about it searching recursive search going back a directory if it's not found in the sub directory. To implement something like that we'd have to maintain a stack, which seems like a lot of effort for a very niche usecase in my opinion.

autoload/vimtex/parser.vim
```diff
@@ -120,6 +120,7 @@ endfunction
 function! s:input_line_parser_tex(line, file, re) " {{{1
   " Handle \space commands
   let l:file = substitute(a:line, '\\space\s*', ' ', 'g')
+  let l:subimport = 0
```
lervag: Now it seems this line has something to do with "handling \space commands". We could either add it as the top line with a space below, or we could add it infront of the `if l:file` block, or in a corresponding `else` block.

autoload/vimtex/parser.vim
```diff
@@ -128,8 +129,10 @@ function! s:input_line_parser_tex(line, file, re) " {{{1
           \ a:file,
           \ '\v^\s*\{')
     if !empty(l:candidate) | return l:candidate | endif
-
```
lervag: I actually prefer not to remove this line, because it separates two "ideas" or code blocks, sort of.

lervag:
Perhaps you could add an issue where you explain the last part of your comment. I could consider to add this if it makes sense to me, and if it seems not too complex.

autoload/vimtex/parser.vim
DelusionalLogic: In a way I almost think the substitute line should be below the let subimport line, since the candidate and subimport lines relate more to each other than the substitute line.

autoload/vimtex/parser.vim
DelusionalLogic: I agree. I'll move it under the comment.

DelusionalLogic:
I added issue #626 

I'm not entirely sure how the documentation is meant to be read, further suggesting that the functionality probably sees very little use.

lervag:
Very good work, thank you!

DelusionalLogic:
Thanks for your patience :)

PR #628, opened 2016-11-28
Infinite recursive main discovery

DelusionalLogic:
Main discovery always ended up starting on the base file when recursing.
To fix this we use the file stored in l:file instead of % (current file).

Trying to echo out l:file and seeing which directories it traverses makes me believe this is what was actually intended.


lervag:
Yes, very good. I never caught this because I always have my source files in the same directory (at least almost). Thank you for finding this and resolving it!

PR #639, opened 2016-12-14
Added notice for terminal users in readme's installation section

Szero:
Hello, I know this is mentioned in the documentation of this program, but it gave me quite a hustle to figure out for a first time since I wasn't using any plugins that required `+clientserver` functionality of vim until now, so I thought that vimtex could be at cause here rather than some vim ambiguity. So to clarify things I think this deserves spot in readme since it's actually an installation step for some use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,14 @@ Or use some other plugin manager:
 - [neobundle](https://github.com/Shougo/neobundle.vim)
 - [pathogen](https://github.com/tpope/vim-pathogen)
 
+### Mention for terminal users
+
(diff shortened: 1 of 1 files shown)


lervag:
Hi. First: I appreciate the effort of actually performing the change you want and creating a pull request!

However, I am not going to merge this. The reason is that the README file is not a documentation, and the info that you supply is already very clearly written in the documentation. See `:h vimtex-requirements`. It's the second subsection in the introduction. I understand that most people won't read the entire documentation, but as a minimum, I should be allowed to expect that people read the introduction.

PR #646, opened 2016-12-22
Option to enable %-movement within comments

tbrk:
Rationale:
  Folds are often delimited using patterns in comments, like:

    %{-{1
    ...
    %}-}1

  and it is useful to be able to skip between the delimiters.


lervag:
I'm curious if there is a good reason in the first place to ignore the comments. Perhaps it is better to just remove that restriction altogether?

tbrk:
I can't think why they should be ignored. I don't think they are ignored in other filetypes. I agree that it would be even simpler to just remove the restriction and to avoid adding more configuration options.

lervag:
Agreed then, and done. I don't think this will break anything, but let me know if I'm wrong!

tbrk:
Great. Thank you.

PR #658, opened 2017-01-10
Corrected Okular options in documentation

Libbum:
When I was setting up the synctex forward search to Okular I noticed an issue with the documentation which led to a bit of confusion. 

You must use `file:@pdf` rather than just `@pdf` in the `g:vimtex_view_general_options` string, otherwise Okular will not be able to open the file.

I've just amended the documentation to reflect this, no other changes are required.


lervag:
I'll take your word for it. Thanks!

PR #674, opened 2017-01-26
Improved parsing of equation numbers and omni-completion for subfile command

slaiba:
1. The parsing of equation numbers did not work in documents where the numbers for example include the chapter number, e.g. things like 1.3 or, for appendices A.3.
Further, with some classes a "\relax" ends up in front of the label, which also causes the parsing to fail.
These problems are fixed by checking the entry from the aux-file against a regular expression.

2. in light of the support for the subfiles package, the \subfile{} command should also be supported by the omnicompletion. I therefore added it as an alternative to \include{} and \input{}


lervag:
Thanks for this!

PR #678, opened 2017-02-03
Modified latexmk.vim to allow empty option

tokoro10g:
This PR changes the default value of `g:vimtex_latexmk_options`.

The new code explicitly sets the options by default instead of implicitly modifying the value.
The empty value passed to the option now simply means that the user does not want to pass any options to latexmk.

Note: Those who have `let g:vimtex_latexmk_options = ''` in their `.vimrc` are affected by this change.


lervag:
Thanks! I think this looks good. Don't remember why I did it like it was, but I can't see why if there was a good reason. I'll therefore merge, and if it should lead to unexpected problems we'll solve them as they arise.

PR #685, opened 2017-02-16
Change the behavior of create-cmd function

Aster89:
Changed the behavior of cmd#create, according to [this discussion](https://github.com/lervag/vimtex/pull/596).


lervag:
I think you missed the normal mode behaviour, which should not change. Thus I chose to apply your fix with a minor change. Let me know if this fixes things for you.

Aster89:
Well, actually I did intentionally, since I never use this function in normal mode. More specifically my thought is as follow. Why should I ever write a command without backslash (è.g. `textsl'), then press <ESC> instead of <F7>, then move in the middle of the word (somehow), and then press <F7> again? I could do this only by pressing <ESC> erroneously, in which case the cursor would still be on the last letter, thus allowing <F7> to do its job. The same holds if I erroneously press <ESC> while writing the command on the left of a word. What do you think about this?

lervag:
Hmm. Ok, what I find here is that you are, sort of, saying that the normal mode version of `<f7>` is not really useful. And that when you want to use it, it is mostly because you erroneously pressed ESC. How about the following: We remove the normal mode `<f7>` and instead map the new `<f8>` to `<f7>` in normal and visual mode?

lervag:
To follow up: This would sort of make sense, since the commands are all about creating commands. In insert mode, it makes sense to quickly turn the current word into a command, but in normal mode and visual mode it makes more sense to wrap the word or selection inside a command specified by the user.

Aster89:
@lervag, I've just updated the plugin to see the new commit's effect. I summarize the changes (more to myself, ahahah)

1. `<F7>` in insert mode: transofrms `emph|This` in `\emph{This|` (like we said it should do, [here](https://github.com/lervag/vimtex/pull/596)).
2. `<F7>` in normal mode: transforms `emph|This` in `\ehmpThis{|`.
3. `<F7>` in visual mode: does nothing.
4. `<F8>` in insert mode: does nothing (actually it inserts the text `<F8>`, unlike 3., I don't know if it's normal or not).
5. `<F8>` in normal mode: prompts to insert a command in which the word under the cursor is to be enclosed.
6. `<F8>` in visual mode: if the first three letters of `enrico` are selected (`r` blinking), it propts to insert a command, but when this is entered it has a strange behavior: if `hello` is inserted the result is `|\helloen{r}ico` in normal mode (`|` = cursor position).

I have the following remarks

1. This is the upgraded version of the previous insert mode `<F7>` mapping. Perfect!
2. I think it's useless, but --- if I correctly understood your last comment --- you said you want to move the `<F8>` normal mode mapping to this mapping (`<F7>` normal mode mapping). If so, skip to remark 5.
3. In your last comment --- if I correctly understood it --- you said you want to move the `<F8>` visual mode mapping to this mapping (`<F7>` normal mode mapping). If so, skip to remark 6.
4. Ok, I agree to let `<F8>` mappings free, since I use it with UltiSnips :smile: 
5. This is a new very useful command (which well goes together with commands like `csc` in normal mode, ...). Perfect!
6. It should be a mistake, or have I misunderstood the trailing part of your last comment?

lervag:
Ok, try now. I merged the commands onto the `<f7>` key. I think the visual mode version should work now; there was a bug in the previous implementation.

Aster89:
As long as I know (and my knowing is limited by my scarse knowledge of vimscripting) I think that the behavior is perfect now! You now much better than me wheter this solution (for what concerns `n` and `x` modes only) of prompting the insertion of a word is more, less, or as fast as leaving Vim in insert mode in between `\` and `{`.

[I make this observation just because of a memory, about another feature, which appears now (I've just checked) to be wrong: if somebody asked me, I would have sweard that typing `csc` in command mode while the cursor is on the content of a LaTeX command would have left Vim in insert mode with the cursor in between `\` and `{`.]

Anyway, I'm glad to have helped you to improve vimtex. Furthermore, I take the occasion to thank you so much, for this plugin is unpredictably fast, customizable, and... don't know, I think "self-explaining" would be  a good translation from italian. And, most of all, there's a lot of Vim philosophy in it! Much more than in plenty of other plugins.

lervag:
Thank you for those kind words, @Aster89, I am very happy to hear that you like the plugin!

The reason I think that the prompt is better than entering insert mode is that I in general want to keep myself in normal mode as much as possible. Thus, by letting `<f7>` in visual mode and normal mode prompt for a command name, I also keep myself in normal mode. I believe this fits pretty well with what one might call "Vim best practises".

Regarding `csc`: I think this worked as you remember once, quite a while ago. But as you must have observed, today it instead prompts for a replacement command and keeps the cursor where it is.

Btw: When you say command mode, you really mean normal mode.

Aster89:
Yes, normal :smile:

PR #690, opened 2017-02-20
Add tsc to toggle starred commands

kiryph:


kiryph:
DO NOT MERGE: there is an issue regarding already starred commands.

kiryph:
As far as I haved checked, this should work after merging the PR #693 .

lervag:
Thanks! I like how you have written this almost exactly the way I would have done it. Whether or not my style of code is good, I still very much appreciate the effort!

kiryph:
Well, I copied and adapted most of the code lines from `vimtex#cmd#change()` and `vimtex#env#toggle_star()`. That's the reason that the code looks very similar to yours.

I guess one could have also refactored the existing code to have a generic change command function with arguments similar as it is done in `env.vim`with `vimtex#env#change(open, close, new)`. I tried to stick to make minimal changes to your code so that the diff looks as small as possible.

There is one thing I was wondering about whether one should use the character based versions of the string functions, such as `strcharpart()` instead of `strpart()` and `strchars()` instead of `strlen()`. Could one run into a problem when somebody uses, e.g. an Asian language?

lervag:
I think you might be right about using the character based versions, but OTOH, these are all very new functions. Thus they won't work for old versions. I would have to write wrappers that work on both new and old vim versions. I could do it, but so far I have not seen any issues due to this. So, my conclusions is that it is not really an issue. If it becomes an issue, then I could solve it as you suggest on new vim versions.

I agree it would make sense to refactor the functions in a similar fashion as for `env.vim`. I might do it if/when I get the time.

PR #696, opened 2017-02-22
Add missing delimiter \lVert and \rVert

kiryph:
\lVert and \rVert represent double bars
http://ctan.org/tex-archive/obsolete/info/math/voss/mathmode/Mathmode.pdf#page=24


lervag:
Thanks!

PR #699, opened 2017-02-25
Interference between vimtex#cmd#create_ask and indent module (clean)

Aster89:
The insertion of `}` before `{` caused an unwanted indent reduction, which resulted in the first char of last visual selection to move rightward and, consequently, in the insertion of the `\command{` in the wrong place.

### DIFF
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -152,9 +152,10 @@ function! vimtex#cmd#create_ask(visualmode) " {{{1
     let l:pos_start = getpos("'<")
     let l:pos_end = getpos("'>")
 
-    normal! `>a}
     normal! `<
     execute 'normal! i\' . l:cmd . '{'
+    normal! `>
+    execute 'normal! ' . (strlen(l:cmd) + 2) . 'la}'
 
     let l:pos_end[2] += 1
     if l:pos_end[1] == l:pos_start[1]


Aster89:
@lervag , I noticed that the indentation rules still interfers with this command in unusual situations.
I triggered it by chance when I decided to split a too big cell content in a tabularx env, e.g.
```
& tooooo looong coooontent of a ceeeell &
```
changed to
```
& tooooo looong %
coooontent of a ceeeell &
```

I think that this is gonna happen whenever one decides to not respect, for some reason and in some way, indentation. So... don't you think it'd be better to somehow dectivate auto-indentation during the execution of the command?

lervag:
Perhaps. But I don't quite understand how to reproduce this. That is, I don't quite see what you mean. Could you explain how to reproduce this issue?

Aster89:
```
\begin{tabular}{ccc}
    a few words & a few words & a few words\\
    a few words & a few words & a few words\\
    a few words & very long text in % new line to avoid breaking the alignment
                  this cell   & a few words\\ % <-----HERE
    a few words & a few words & a few words\\ % <-----HERE
    a few words & a few words & a few words\\
\end{tabular}
```
Try enclosing the content of the cells of the two highlighted rows in some command using the visual mode mapping. Trying to put `this cell` in `emph`asis produces the following effect
```
\begin{tabular}{ccc}
    a few words & a few words & a few words\\
    a few words & a few words & a few words\\
    a few words & very long text in % new line to avoid breaking the alignment
    \emph{this cell   & a few wor}ds\\ % <-----HERE
    a few words & a few words & a few words\\ % <-----HERE
    a few words & a few words & a few words\\
\end{tabular}
```

lervag:
Ah, I see. I think my latest patch fixes this.

Issue #700, opened 2017-02-26
Zathura view method cannot find PDF if `$out_dir` is given in .latexmkrc


j-jith:
When I don't want LaTeX polluting my source directory with its build files, I use a simple .latexmkrc which looks like this:
```
$out_dir = 'build';
```
If I use `g:vimtex_view_method='zathura'` now and try the command `:VimtexView`, I get this error - `vimtex warning: viewer can not read PDF file!`. It turns out that the `b:vimtex.out(1)` and `b:vimtex.viewer.synctex` do not match with the pdf and synctex files generated by latexmk due to vimtex being unaware  of `$out_dir`.

I have verified that if I specify `let g:vimtex_latexmk_build_dir=<build dir>` in my .vimrc, everything works as expected.

### Minimal working example

```tex
\documentclass{minimal}
\usepackage{lipsum}
\begin{document}
\lipsum
\end{document}
```

### vimrc

```vim
set nocompatible

" Load Vimtex
let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'

filetype plugin indent on
syntax enable

" Vimtex options go here
let g:vimtex_view_method = 'zathura'
```

### latexmkrc

```
$out_dir = 'build';
```

j-jith:
I have just tried this with qpdfview. It shows the same error as zathura. The vimrc used was

```vim
set nocompatible

" Load Vimtex
let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'

filetype plugin indent on
syntax enable

" Vimtex options go here
let g:vimtex_view_general_viewer = 'qpdfview'
let g:vimtex_view_general_options = '--unique @pdf\#src:@tex:@line:@col'
let g:vimtex_view_general_options_latexmk = '--unique'
```

I have created a new pull request regarding this.

lervag:
Interesting. First, I've never thought about parsing the `.latexmkrc` file for options. It seems like a good idea, at least for this particular case.

I'll continue the discussion in #701.

PR #701, opened 2017-02-26
Reads $out_dir from local latexmkrc. $out_dir given precedence over g…

j-jith:
…:vimtex_latexmk_build_dir

Regarding issue #700 

- Added function `s:get_latexmkrc_out_dir` to get `$out_dir` from `.latexmkrc` in the buffer's local/root directory.

- `$out_dir` is given precedence over `g:vimtex_latexmk_build_dir`.

- Modified function `s:get_ext` so that `b:vimtex.out(1)` and `b:vimtex.viewer.synctex` have the correct paths.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -739,11 +739,35 @@ function! s:get_out(...) dict " {{{1
 endfunction
 
 " }}}1
+
+function! s:get_latexmkrc_out_dir(dir) " {{{1
+  " check if local latexmkrc exists
+  if filereadable(a:dir.'/.latexmkrc')
+    let l:pattern = '\%(^\s*\$out_dir\s*=\s*[''"]\(.\+\)[''"]\s*;\)'
+    let l:out_dir = matchlist(readfile(a:dir.'/.latexmkrc'), pattern)
+    if len(l:out_dir) > 1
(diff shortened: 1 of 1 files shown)


andymass:
Support for automatically determining the output directory is a much needed feature- but would it be more reliable to get the output with something like `latexmk -deps | grep target` instead of parsing `.latexmkrc`?

lervag:
Thanks, I really appreciate that you both see something you find missing and actually submit a suggestion for how to solve it!

I hope you don't mind that I implement this myself. I want to use the `...build_dir` option in general, and so instead of changing the `get_ext` function I only change how the `build_dir` option is set.

lervag:
Let me know if this fixes the issue for you. The main differences from your implementation is:

1. I check both `latexmkrc`, `.latexmkrc` and `~/.latexmkrc`
2. I allow the parsed line to not end in semicolon (this is OK if there is only one line)
3. My version only modifies the `g:vimtex_latexmk_build_dir` setting itself and nothing else

j-jith:
Thanks a lot @lervag. I think this fixes all the issues I had.

lervag:
No problem! And thank you, I think this was an improvement to vimtex. :)

PR #702, opened 2017-03-03
Add 'g:vimtex_quickfix_height'

aldur:
I added an option to customize the height of the quickfix window.
This is my first VimL contribution, so feel free to point out anything not "standard".

### DIFF
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -32,6 +32,7 @@ function! vimtex#latexmk#init_options() " {{{1
   call vimtex#util#set_default('g:vimtex_quickfix_autojump', '0')
   call vimtex#util#set_default('g:vimtex_quickfix_mode', '2')
   call vimtex#util#set_default('g:vimtex_quickfix_open_on_warning', '1')
+  call vimtex#util#set_default('g:vimtex_quickfix_height', '')
 
   if exists('g:vimtex_quickfix_ignore_all_warnings')
     echoerr 'Deprecated option: g:vimtex_quickfix_ignore_all_warnings'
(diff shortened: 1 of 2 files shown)


lervag:
I'm not sure about this. I find this to be a more general issue that is more about how the quickfix window works than about LaTeX editing.

E.g., I personally use the following autocommand:

```vim
augroup quickfix_autocmds
  autocmd!
  autocmd BufReadPost quickfix set modifiable
  autocmd BufReadPost quickfix call AdjustWindowHeight(2, 30)
augroup END

function! AdjustWindowHeight(minheight, maxheight)
  execute max([a:minheight, min([line('$') + 1, a:maxheight])])
        \ . 'wincmd _'
endfunction
```

This sets the height of the quickfix window to the number of entries, but not larger than the maxheight or smaller than the minheight. You could use a similar approach to set a specific height. The most simple version would be:

```vim
autocmd BufReadPost quickfix 20wincmd _
```

This sets the quickfix height to 30 lines.

The bottom line is: I don't think this should be added to vimtex. But please feel free to try to convince me I'm wrong!

aldur:
I actually like your approach, because it solves a more general issue.

The main reason I came up with a `vimtex` specific solution is that [Neomake](https://github.com/neomake/neomake/blob/eb1f7d495208b22680d1ef769c02929e29a94c30/doc/neomake.txt#L309), [Syntastic](https://github.com/vim-syntastic/syntastic/commit/9a6895d28de62a1f6592883fc88471cc25cf555d) and possibly others seem to prefer a "plugin" specific solution.

Are there any downsides with your approach?

lervag:
No, there are no downsides as far as I can tell. My approach is just a fairly general approach to control how the quickfix list will appear. However, I can't guarantee that there is not a better approach or that mine is free from any bugs. I have used it for some years now, though, and it seems to work fine.

PR #703, opened 2017-03-04
Delete environment along with begin/end lines

Aster89:
A change is made to `vimtex#env#delete` in a way that it deletes `\begin{...}` and `\end{...}` lines. Before this change the command `dse` used to leave empty lines in place of `begin`/`end` lines or, had options been present, the `[...]` part in place on the whole `begin` line.
(It does not make use of `vimtex#evc#change anymore`.)

### DIFF
--- a/autoload/vimtex/env.vim
+++ b/autoload/vimtex/env.vim
@@ -103,6 +103,12 @@ function! vimtex#env#delete(type) " {{{1
   if empty(l:open) | return | endif
 
   call vimtex#env#change(l:open, l:close, '')
+  if (getline(l:close.lnum) =~ "^ *$")
+    execute l:close.lnum . 'd'
+  endif
+  if (getline(l:open.lnum) =~ "^ *$")
+    execute l:open.lnum . 'd'
+  endif
 endfunction
 
 function! vimtex#env#toggle_star() " {{{1
--- a/test/vader/env.vader
+++ b/test/vader/env.vader
@@ -41,3 +41,22 @@ Expect tex (Verify):
 
 Before:
 After:
+
+Given tex (Environments: foo and bar):
+  \begin{equation}
+    f(x) = \begin{bmatrix}
+        1 & 2 \\
+        3 & 4 \\
+      \end{bmatrix}
+  \end{equation}
+
+Do (Delete env: foo (TODO)):
+  jfmdse
+
+Expect tex (Verify):
+  \begin{equation}
+    f(x) = 
+        1 & 2 \\
+        3 & 4 \\
+  \end{equation}
+


lervag:
This change assumes that the `\begin{...}` and `\end{...}` commands are on separate lines, with nothing else there. Thus it breaks cases like this:

```tex
\begin{equation}
f(x) = \begin{bmatrix}
    1 & 2 \\
    3 & 4 \\
  \end{bmatrix}
\end{equation}
```

There many different styles of writing TeX, and I am sure it is not OK to assume that we can delete the entire lines that has `\begin` and `\end`.

Instead of the proposed change, perhaps a better solution is to check if the changed lines become empty. That is, before the end of the function `vimtex#env#delete`, we add something like this:

```vim
if empty(getline(l:close.lnum))
  execute l:close.lnum . 'd'
endif
if empty(getline(l:open.lnum))
  execute l:open.lnum . 'd'
endif
```

I'm writing this fast and have not tested it, so it might not be entirely correct. But perhaps you could try this and test if it achieves what you want without breaking my before mentioned example?

lervag:
I've just added my example to the test suite. So, if you try to update your PR, it should fire a test that should let us know if it works.

Note, though, that I should probably add many more tests... :)

lervag:
Ah, also note: I've marked the new test with TODO, which means that even if it fails, it will report as OK. Thus the test suite will still probably pass. But you can see in the results if the test has passed or not (just click the test links).

Aster89:
@lervag , to be honest I haven't thought about different styles of writing TeX, which results in my suggestion to be pointless (at least in the context of project which is shared among a lot of people).

My initial idea was to remove those lines as well as indentation of the other lines between them. But now you've made me think of the possibility that someone simply doesn't ident...

Even so, maybe your correction could be implemented. The only modification I would add is aimed at removing only-spaces lines which result when the environment to be deleted is indented. I've just tried to change if `empty(getline(__))` to `if (getline(__) == "^ *$")`, which doesn't work :disappointed:

Aster89:
I think it's not unlikely that I've just made a mess again with commits :stuck_out_tongue_closed_eyes:

lervag:
Good work! I decided to apply the patch manually, both to avoid the messy commits, and to make a couple of minor changes. See the commit (02c2b2d) to see how I did it; I slightly altered your version to:

1. Use `=~#` instead of `=~` to be specific about whether or not to ignore case, and

2. Use single quotes for strings unless double quotes are "necessary". This is because double quotes allow escape codes and similar.

I also added a couple of comments for future reference and altered and activated the test case. It seems to work pretty much as desired.

In any case, thanks for noticing this and suggesting the improvement. Also thanks for suggesting a specific implementation. This is very helpful and makes it easier to respond and follow up for me.

----

Note: In future PRs, I propose that you look into git rebasing. Working inside the confinements of a PR and a git branch makes it very safe to use git rebase to clean up the commits in a branch. This often requires that one do `git push --force`, because the push needs to change the server commits. However, I still find this to be appropriate as long as it is confined to a separate branch and if the responsible person knows what he/she is doing.

PR #707, opened 2017-03-08
Delete environment into black hole register

Aster89:
If `dse` removes lines, it deletes them into the black hole register (`_`), thus not wasting the registers " and 1, nor losing the register 9.

I've found no other situations in which such a change is needed. Sometimes, e.g. for the `vimtex#cmd#create_ask` function, VimTeX *has to* keep the deleted text in some register, since it is reused afterwards.


lervag:
Thanks!

PR #708, opened 2017-03-10
Fix to insert command

Aster89:
This commit is somewhat related to the commit merged from [this PR](https://github.com/lervag/vimtex/pull/596) and is coherent with the definition of a LaTeX command ([egreg](http://tex.stackexchange.com/a/34383/87996)): it avoids, for instance, `_text<F7>` to become `\_text{`.


lervag:
Thanks!

PR #711, opened 2017-03-10
[feature] Completion for usepackage and documentclass

### LINKED ISSUE #709: Completion source for \usepackage{ based on tlmgr list --only-installed
Would you consider adding a cached omnicompletion source based on `tlmgr list --only-installed` triggered when the user has entered `\usepackage{` or `\RequirePackage{`?

lervag:
This PR fixes #709.

Remaining tasks:
- [x] Update documentation
- [x] Handle `TEXMFHOME` with missing `ls-R` (?)

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -50,6 +50,8 @@ CONTENTS                                                      *vimtex-contents*
     Complete PDF file names                     |vimtex-complete-pdf|
     Complete standalone file names              |vimtex-complete-standalone|
     Complete glossary entries                   |vimtex-complete-glossary|
+    Complete packages and documentclasses       |vimtex-complete-packages|
+                                                |vimtex-complete-classes|
     Autocomplete                                |vimtex-complete-auto|
       Neocomplete                               |vimtex-complete-neocomplete|
       deoplete                                  |vimtex-complete-deoplete|
@@ -1677,6 +1679,48 @@ Glossary entry completion from the `glossaries` package are triggered by the
 commands '\gls{', '\glspl{' and their variations.
 
 ------------------------------------------------------------------------------
+Complete packages and documentclasses~
+                                                     *vimtex-complete-packages*
+                                                      *vimtex-complete-classes*
+
+Package and documentclass completion are available for the `\usepackage{` and
+`\documentclass{` commands, respectively.
+
+The completion relies on the contents of `ls-R` files that are found with: >
+
+  kpsewhich --all ls-R
+
+Packages and documentclasses installed at `TEXMFHOME` will also be searched.
+This variable has a default value that depends on the system. The value that
+is used can be found with: >
+
+  kpsewhich --var-value TEXMFHOME
+
+However, one should be aware that Vim invokes shells as non-interactive and
+non-login shell, which means it inherits environment variables defined in
+`.bashrc` or `.zshrc` only if it is invoked from an interactive shell. This is
+typically not the case if you use a desktop environment launcher.
+
+One can make the invoked shells interactive by setting |shellcmdflag| to "-ic".
+If you want to keep them non-interactive, you can create an additional shell
+startup file where you keep your environment variables:
+
+  1. If bash is your default shell, create e.g. the file `.bashenv` containing
+     your customized `TEXMFHOME` variable and add `$BASH_ENV=$HOME/.bashenv` to
+     `$MYVIMRC` and source `$HOME/.bashenv` to `.bashrc` [0].
+
+  2. If zsh is the default shell, use `.zshenv` for customizing `TEXMFHOME`.
+     This file is always read by zsh. Nothing has to be added to the
+     `$MYVIMRC` [1].
+
+For more information on how to correctly set environment variables, see e.g.
+the SO answer by @Rmano [2].
+
+[0]: https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html
+[1]: http://zsh.sourceforge.net/Intro/intro_3.html
+[2]: http://askubuntu.com/a/356973/16395
+
+------------------------------------------------------------------------------
 Autocomplete~
                                                          *vimtex-complete-auto*
 
@@ -1707,6 +1751,8 @@ documents with |neocomplete| and |vimtex|s omni completion function: >
         \ . '|\a*(gls|Gls|GLS)(pl)?\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
         \ . '|includepdf%(\s*\[[^]]*\])?\s*\{[^}]*'
         \ . '|includestandalone%(\s*\[[^]]*\])?\s*\{[^}]*'
+        \ . '|usepackage%(\s*\[[^]]*\])?\s*\{[^}]*'
+        \ . '|documentclass%(\s*\[[^]]*\])?\s*\{[^}]*'
         \ . ')'
 
 deoplete~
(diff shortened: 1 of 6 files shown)


kiryph:
Thanks for the update. However, there is a naming issue with `s:get_texmf` which can be fixed according to this diff:

```diff
~/.vim/plugged/vimtex feature/complete-usepackage*
❯ git diff
diff --git a/autoload/vimtex/complete.vim b/autoload/vimtex/complete.vim
index 55e44ca..9334582 100644
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -544,7 +544,7 @@ endfunction
 
 function! s:pck.gather_candidates() dict " {{{2
   if empty(self.candidates)
-    let self.candidates = map(s:get_texmf('sty'), '{
+    let self.candidates = map(s:get_texmf_candidates('sty'), '{
           \ ''word'' : v:val,
           \ ''menu'' : '' [package]'',
           \}')
@@ -568,7 +568,7 @@ endfunction
 
 function! s:doc.gather_candidates() dict " {{{2
   if empty(self.candidates)
-    let self.candidates = map(s:get_texmf('cls'), '{
+    let self.candidates = map(s:get_texmf_candidates('cls'), '{
           \ ''word'' : v:val,
           \ ''menu'' : '' [documentclass]'',
           \}')
```

kiryph:
To trigger `neocomplete`, one can use following snippet:
```
if !exists('g:neocomplete#sources#omni#input_patterns')
	let g:neocomplete#sources#omni#input_patterns = {}
endif
        \ '\v\\%('
        \ . '\a*%(ref|cite)\a*%(\s*\[[^]]*\])?\s*\{[^{}]*'
        \ . '|includegraphics%(\s*\[[^]]*\])?\s*\{[^{}]*'
        \ . '|%(include|input)\s*\{[^{}]*'
        \ . '|usepackage%(\s*\[[^]]*\])?\s*\{[^}]*'
        \ . '|documentclass%(\s*\[[^]]*\])?\s*\{[^}]*'
        \ . ')'
```

lervag:
Sorry, I was too fast while finishing up my last session. It's fixed now. I also updated the autocomplete settings for neocomplete, deoplete and YouCompleteMe.

lervag:
Ok, I've added documentation and cleaned up the commits. I'm proposing a merge, but I'd be happy to get some feedback on 1) the documentation, and 2) whether or not to implement an automatic globbing of the `TEXMFHOME` directory (instead of parsing the `ls-R` file).

At the moment, I'm adding the `TEXMFHOME/ls-R` if it exists. However, I could also do this:

```vim
  let l:texmfhome = get(vimtex#kpsewhich#run('--var-value TEXMFHOME'), 0, '')
  let l:candidates += glob(l:texmfhome . '/**/*.' . a:filetype, 0, 1)
  call map(l:candidates, 'fnamemodify(v:val, '':t:r'')')
```

This seems to work as expected. Do you see any reasons why I should _not_ do that instead of the current version that reads the `ls-R` file?

lervag:
Ok, I think my suggestion is better than parsing the `ls-R` file for `TEXMFHOME`, so I just pushed it. I'll still wait for some feedback, though.

doc/vimtex.txt
```diff
@@ -1645,6 +1647,25 @@ Glossary entry completion from the `glossaries` package are triggered by the
...
+The completion relies on the contents of `ls-R` files that are found with: >
+
+  kpsewhich --all ls-R
+
+Packages and documentclasses installed at `TEXMFHOME` will also be searched.
+
+Note: If you set `TEXMFHOME` in your `.bashrc` or `.zshrc` or similar, then it
+      might not be inherited by `gvim`, since `gvim` is often started by the
+      desktop environment (from menus, hotkeys, etc.). If you should suspect
+      this happens to you, please figure it out yourself.
```
kiryph: IMHO, `please figure it out yourself` is not very helpful. Following information might help a user:

> vim invokes shells as non-interactive and non-login shell, i.e. it does not read `.bashrc` or `.zshrc`. It only inherits environment variables defined in `.bashrc` or `.zshrc` if it is invoked from an interactive shell. This is typically not the case if you use a desktop environment launcher.
> One can make the invoked shells interactive by setting `set shellcmdflag=-ic`. If you want to keep it non-interactive, you can create an additional shell startup file where you keep your environment variables: 
> 1) if bash is your default shell,  create e.g. the file `.bashenv` containing your customized TEXMFHOME variable and add `$BASH_ENV=$HOME/.bashenv` to `$MYVIMRC` and `export $BASH_ENV=$HOME/.bashenv` to `.bashrc`. (https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html)
> 2) if zsh is the default shell, use `.zshenv` for customizing TEXMFHOME. This file is always read by zsh. Nothing has to be added to the `$MYVIMRC`. (http://zsh.sourceforge.net/Intro/intro_3.html)

See following questions and webpage for detailed information on shell startup files and vim:
- https://stackoverflow.com/q/14841208/bash-commands-from-vim
- https://superuser.com/q/585528/gvim-does-not-source-bashrc-bash-profile-or-profile
- https://stackoverflow.com/q/28017016/path-variable-not-updating-when-sourcing-bashrc-inside-vim
- https://shreevatsa.wordpress.com/2008/03/30/zshbash-startup-files-loading-order-bashrc-zshrc-etc/

Rmano: You can also point to this answer: http://askubuntu.com/a/356973/16395 (I wrote that when debugging why my TEXMFHOME was seem to work erratically)  ;-)

kiryph: One should mention that TEXMFHOME does not have to be set manually, it is by default defined depending your system:

- OSX:
 ```
❯ kpsewhich --var-value TEXMFHOME
/Users/kiryph/Library/texmf
```
- Linux: 
```
$ kpsewhich --var-value TEXMFHOME
/home/kiryph/texmf
```

**Comment by kiryph** [comment]
As far as I have tested it with a package being only in TEXMFHOME, completion works just fine.
```
❯ kpsewhich --all MyriadPro.sty
/Users/kiryph/Library/texmf/tex/latex/MyriadPro/MyriadPro.sty
❯ kpsewhich --all MinionPro.sty
/Users/kiryph/Library/texmf/tex/latex/MinionPro/MinionPro.sty
/usr/local/texlive/texmf-local/tex/latex/MinionPro/MinionPro.sty
❯ kpsewhich --all scrartcl.cls
/usr/local/texlive/2016/texmf-dist/tex/latex/koma-script/scrartcl.cls
```

I still have some problems with neocomplete that the cache is out-of-date and I have to update the neocomplete menu manually to see all candidates. Maybe I should consider switching to one of the newer completion plugins which query the completions source asyncrounously but update the menu automatically as soon as the completion source is finished. However, most of them require you to write a dedicated completion source; providing a simple `omnifunc` is not sufficient to take advantage of the async feature.

lervag:
Thanks! I've updated the docs according to your comments. Let me know what you think. I'll consider your last comment later today, @kiryph. I have to run now. ;)

lervag:
@kiryph Yes, the caching in neocomplete does not always seem to work as expected. However, it seems to work pretty well for me. I'm also considering to try a different (auto)complete plugin, but I haven't had the time to test and choose one yet. I'm also considering to go back to more regular Vim completion similar to SuperTab, since I've noticed I actually don't really use completion that much.

lervag:
I think the feature seems pretty robust now, and so I'll rebase and merge this now. Still, please feel free to either continue the discussion here or open new issues if you think there are things that should be improved or changed.

PR #714, opened 2017-03-14
Move details about changing TEXMFHOME to faq

kiryph:
Keep the documentation for the completion of
\usepackage & \documentclass a little bit lighter and
separate the discussion of changing TEXMFHOME


lervag:
Thanks! I agree this makes the documentation lighter.

PR #721, opened 2017-03-16
Missing space when adding b:vimtex.engine to build cmd

kiryph:


lervag:
Thanks! I added it manually, because I just pushed some other commits. Also, you forgot one (there were two cases of the same issue). :)

Issue #637, opened 2016-12-10
wincmd p unreliable


languitar:
### Explain the issue

As explained in https://github.com/mhinz/vim-signify/issues/193 I have encountered a jumping cursor when combining fugitive, vimtex and an autocmd to update vim-signify. I could get rid of this issue, by applying the following patch to the vimtex code:

```diff
diff --git a/autoload/vimtex/latexmk.vim b/autoload/vimtex/latexmk.vim
index 939d192..bf340b9 100644
--- a/autoload/vimtex/latexmk.vim
+++ b/autoload/vimtex/latexmk.vim
@@ -346,6 +346,8 @@ function! vimtex#latexmk#errors_open(force) " {{{1
   if !exists('b:vimtex') | return | endif
   cclose
 
+  let winid = win_getid()
+
   let log = b:vimtex.log()
   if empty(log)
     if a:force
@@ -386,7 +388,7 @@ function! vimtex#latexmk#errors_open(force) " {{{1
   if s:open_quickfix_window
     botright cwindow
     if g:vimtex_quickfix_mode == 2
-      wincmd p
+      call win_gotoid(winid)
     endif
     redraw
   endif
```

lervag:
Interesting. I wasn't aware of the `win_...` functions. They seem to be relatively new. That makes it a little bit scary to use them, since it would/could lead to errors for people who use old Vim versions.

Perhaps we could fix this with "older" syntax? Perhaps we could use the more "standard" `winnr()` function and then `execute l:winnr . 'wincmd p'`?

languitar:
@mhinz used a conditional version in the experimental commit he prepared for me to test this. Use that function if it exists, else winnr. Maybe that could be way to go?

mhinz:
`win_getid()` and `win_gotoid()` were both introduced in the same commit: [7.4.1557](https://github.com/vim/vim/commit/86edef664). Thus I just check if that function is available and fall back to using `winnr()`/`wincmd p` otherwise: https://github.com/mhinz/vim-signify/pull/194/commits/3a8b31d45ba19a96144267bad054a2a88bf2d9c9

That said, the fallback isn't 100% reliable as well, since window numbers can change when new windows open (or close). The reliable way for older versions would be setting a `w:var` and searching that window at the end by looping over `winnr('$')`.

lervag:
[...]

PR #735, opened 2017-03-22
Reliably return to previous window in modern Vim

languitar:
Use window IDs to return to the window that is currently being edited
after opening the quickfix list if these functions are available. Else,
use the previous mechanism of window numbers.

This is related to the changes previously discussed in #637.


lervag:
I'll accept this, but I'm going to change it. I'd rather hide the complexity in higher order functions.

lervag:
I also implemented the window variable trick, which should probably work better than my previous method, as explained by @mhinz. I guess I overlooked that part the last time.

languitar:
Thanks!

PR #742, opened 2017-03-27
PR to fix #630 (improved subfile support)

### LINKED ISSUE #630: `VimtexToggleMain` to preserve root folder.
According to current behavior `VimtexToggleMain` changes the root folder to the base directory of the included file.

When the subfiles package is used with included files located in subdirectories this behavior is undesired. The problem is that subfiles package includes the main file's preample. The preamble may include `usepackage/include' calls that refer to the local files in the same folder with the main file. In this case the compilation of the included file in its own directory will crash since latex will not be able to find the files to include/use.

I suggest to either change the default behavior of the `VimtexToggleMain` in order to preserve root folder, or introduce `VimtexToggleMain!` version which does this.

lervag:
This should fix #630.

### DIFF
--- a/autoload/vimtex/compiler.vim
+++ b/autoload/vimtex/compiler.vim
@@ -36,7 +36,13 @@ function! vimtex#compiler#init_state(state) abort " {{{1
   if !g:vimtex_compiler_enabled | return | endif
 
   try
-    let a:state.compiler = vimtex#compiler#{g:vimtex_compiler_method}#init()
+    let l:options = {
+          \ 'root': a:state.root,
+          \ 'target' : a:state.base,
+          \ 'target_path' : a:state.tex,
+          \}
+    let a:state.compiler
+          \ = vimtex#compiler#{g:vimtex_compiler_method}#init(l:options)
   catch /vimtex: Requirements not met/
     call vimtex#echo#echo('- vimtex#compiler was not initialized!')
     call vimtex#echo#wait()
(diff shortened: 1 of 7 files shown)


lervag:
Ok, no. This PR will only add the improved compiler layer, it will not fix #630. I'll do that in more steps.

It now remains to update the documentation.

jonfin:
Thank you. Now it is working. Even with an output dir set through latexmkrc.

lervag:
Great :)

jonfin:
Sorry, I was to early.
I found a small error with my last project. The script doesn't change correctly change if the path preamble of the sub file is like this
`\documentclass[../main.tex]{subfiles}`
but it is working when I change it to:
`\documentclass[main.tex]{subfiles}`

If I change it in the test case to the `../main.tex`, I get this info
```
vimtex project: sub                                                                                                                 
  base: sub.tex                                                                                                                     
  root: /home/jonas/630/sub                                                                                                         
  tex: /home/jonas/630/sub/sub.tex                                                                                                  
  out:                                                                                                                              
  log:                                                                                                                              
  aux:                                                                                                                              
  compiler: latexmk                                                                                                                 
    configuration:                                                                                                                  
      continuous: 1                                                                                                                 
      callback: 1                                                                                                                   
      latexmk options:                                                                                                              
        -verbose                                                                                                                    
        -pdf                                                                                                                        
        -file-line-error                                                                                                            
        -synctex=1                                                                                                                  
        -interaction=nonstopmode                                                                                                    
  viewer: General
```

lervag:
That's by intention, although it may still be wrong. It is important that the root is set appropriately, but for subfiles, it is not entirely clear what is appropriate. Is it _always_ right to use the main file root?

The current implementation assumes that the use of relative paths imply that the current file defines the root, whereas the use of "dangling" paths implies that the root is the root of the main file.

jonfin:
Thank you for the clarification. From me it's sound OK.

lervag:
Great, and thanks for your input on this matter.

Den tir. 4. apr. 2017, 22.10 skrev Jonas <notifications@github.com>:

Issue #651, opened 2017-01-04
Improve vimtex compiler and error parsing layer


lervag:
# Description

There seems to be sufficient desire for using alternative compilers and improved error parsers for me to start work on this. It will require a full refactoring/rewrite of the compiler layer (vimtex/autoload/latexmk.vim). The following should be some good starting points for a todo list:

1. [x] Refactor the latexmk code into a generalized compiler interface and a "latexmk submodule"
2. [x] Separate the compiler code from error parsing
3. [x] Add a proof-of-concept layer for latexrun as an alternative to latexmk

# Some remarks

* This issue replaces/includes #585. Point 2 above should ultimately provide an interface that would allow to use `rubber-info` as a layer between the Vim errorformat and the LaTeX log files. Thanks to @teto for the suggestions and links: [](http://tex.stackexchange.com/questions/10558/use-latexmk-to-filter-the-log-file) and [pplatex](https://github.com/stefanhepp/pplatex).
* This issue replaces #602 (points 1 and 3 above should fix #602).
* This work may/will take some time. Since it is also a major update, I will work on it in a separate branch. I will let (interested) users know about the progress when there is something to update, and I would very much appreciate feedback both on the description of my ideas here and on progress when I have started the work.

teto:
Great initiative !
Maybe ou can get some insipiration from TexStudio which allows to choose different build systems.
I like how vim-grepper https://github.com/mhinz/vim-grepper allows to switch between different search tools with 'Tab'. This might be useful for testing, to easily switch between different tex compilers.

clason:
Excellent idea! For reference, another popular compiler seems to be https://github.com/cereda/arara.

languitar:
Is there any progress on the improved error parsing originating from #585 which I could try out? Or any way to get this going with a hack in the meantime for my document? It's quite bothersome that I cannot go to the location of an error in a huge document like a thesis.

languitar:
Btw, I have just learned that the rubber error parser is also broken and sometimes swallows error messages.

lervag:
I'm sorry, but no, not yet. The reason is that this is a rather big undertaking. Considering that I've had some time lately to work on issues, I might actually start on this in not too long.

I don't think there is any easy hacks either. And I'm not even sure if a successful implementation of the intended feature here will really solve your problem. But I hope so. :)

lervag:
[...]

PR #743, opened 2017-03-27
Separate general compiler feature from specific latexmk interface

lervag:
This relates to #651.

Remains:
- [x] Documentation
- [x] Testing


lervag:
The code passes all the tests (although I should probably add more). It seems to work well on my own projects.

It remains to add documentation, then I'll propose this is merged.

lervag:
@clason says it does not work with `neovim-remote`. Error "E731: dictionary used as string" after `\le`.

lervag:
@clason I'm not quite sure what is wrong now. It works with `vim` and `gvim`. Could you give some more details? E.g., does the error say something about exactly where it fails (you can get the entire error message with `:messages`)? Or could you provide an example of how to reproduce?

clason:
Sure! (not quite) minimal `vimrc`:

```
" mappings for easier movement
let mapleader = ','
let maplocalleader = ','
" plugins via vim-plug
call plug#begin('~/.vim/plugged')
Plug 'lervag/vimtex', {'for': 'tex'}
call plug#end()
" vimtex settings
let g:tex_flavor='latex'
let g:vimtex_quickfix_warnings = {
            \ 'font' : 0,
            \ 'overfull' : 0,
            \ 'underfull' : 0,
            \ 'packages' : {
            \   'default' : 0,
            \ },
            \}
let g:vimtex_compiler_progname = 'nvr'
```

Minimal `test.tex`:
```
vimrc
\documentclass{article}

\begin{document}

sdfsdf
\sdfsdfs

\end{document}
```

Compiling opens the pdf preview, but not the quickfix list; `VimtexErrors` does open the quickfix list, but not before throwing the error

```
Error in execution of "function 25":
Line 26:
E731: Dictionary used as String
```

lervag:
Thanks! I fixed the quickfix error. It remains to fix the problem with callbacks for neovim. I'll look at it later tonight.

clason:
Nice! Now it works as expected: Quickfix list opens automatically if there's an error or warning. The only possible callback problem I can see is that it doesn't print `compile: success` or `fail`, but that's purely cosmetic.

lervag:
Hm, @clason: See [here](https://github.com/neovim/neovim/issues/5056#issuecomment-290151557): My most recent fix should not be necessary on newer neovim versions. I was curious if it would be bad of me to revert my fix and instead demand that neovim users use a more recent version of neovim?

clason:
Yeah, that issue did sound familiar... But I am using the latest released neovim version (0.1.7) which does not have that change; the new version (0.2.0) is long overdue, but expected Real Soon(tm). But as soon as it is released, I agree that the fix can be reverted.  (And I would even understand if you preferred to avoid the extra code path at the expense of accepting short-term breakage for us early-adopters until the new version is released.)

lervag:
Although I understand it is a little bit annoying, I think I want to allow a short-term breakage. I feel this is more of a bug at neovim, and it feels wrong to add a hack to vimtex when I know that a fix has been merged upstream.

I hope you can live with the minor inconveniance for a little while. This feature hasn't been merged yet either, but I plan to merge this week unless something important/unexpected bad comes up.

It should be pretty easy to workaround, either by disabling the callback, or by manually editing the code.

In any case, the other problem was actually a good catch, because it was a rather important bug. It is fixed now, so everything should work as expected on a recent neovim version now.

clason:
It is a bit (but not very) annoying since I a) completely transitioned to neovim on Linux and b) tend to check in changes before manually checking for errors... But I completely understand, and I appreciate you trying to support neovim as well as vanilla vim. 

Is this something I could fix locally on the vimrc level, short term?

lervag:
You can do this:

```vim

autocmd User VimtexEventInitPost au! vimtex_quickfix_fix_dirs
```

This will remove the autocommand that fixes paths, which might be slightly inconvenient. But most errors should show up correctly, I think. And the error will not be there.

clason:
That's perfectly fine as a workaround until the new version is finally released. Thanks!

lervag:
No problem. I'm happy we found a good compromise here.

I'll merge this PR tomorrow unless there are more comments/issues.

PR #757, opened 2017-04-03
Use system() by default

lervag:
Cf. #651. This changes the process.vim interface so that it uses `system()` by default. The only time that `:!` is used is when one wants a process to run in the foreground (i.e. if the option `background = 0`).

This also provides some minor fixes to other process related stuff and to the compiler layer.


lervag:
This does not work on Windows. I have no idea how to make it work on windows, but I won't merge it since it will break things.

I think a better idea is to keep the current state, but instead add support for the new Vim 8 jobs API.

PR #758, opened 2017-04-04
Preserve root dir when changing to local file

jonfin:
Dear lervag
I had the same problem as in #630 that i have multiple files in subdirs. I tried your patch #742. Unfortunately this wasn't working as aspect. So I wrote a small workaround. 

My problem: I never wrote something in vimscript before. So I don't know how to compare two strings to get the file path relative to the tex root dir. The workaround is only working in the case where you can switch the cwd from vim to the project folder.

I hope it will help you to solve this problem.

### DIFF
--- a/autoload/vimtex/state.vim
+++ b/autoload/vimtex/state.vim
@@ -63,6 +63,12 @@ function! vimtex#state#toggle_main() " {{{1
           \ : b:vimtex_local.main_id
     let b:vimtex = vimtex#state#get(b:vimtex_id)
 
+    " Reset the root dir to dir from main file
+    let b:vimtex.root = vimtex#state#get(b:vimtex_local.main_id).root
+    " Get the active file realtive to the root dir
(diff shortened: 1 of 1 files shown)


lervag:
Hi @jonfin. Thanks for this. However, I am curious if you could help me understand exactly what is not working. I just updated #742 (rebased the branch on top of current master). When you say it does not work, could you please be more specific?

lervag:
Btw: I appreciate the effort on the PR, but I'm not going to merge it. The reason is that you are solving the problem in the wrong end. The `root` and `base` should be set correctly when defining the local state object, i.e. in the `#init_local()` function.

lervag:
I took a new look at #742, and I found an error. After updating, rebasing and testing, I think it now works as intended. Could you please verify?

I took the liberty to merge the PR, because it should have not side effects, and I think it solves our issue.

PR #764, opened 2017-04-06
Update vimtex.txt

rafaeln:
Wrong variable name. Maybe an older name lingered on in the documentation?


lervag:
Thanks, nice catch!

rafaeln:
Not so much of a catch, actually. I've been using this plugin for a while, and so I had the old variable name in my vimrc. I noticed I stopped being able to do backward search a while ago and was about to file an issue report, but before I did so I decided to grep the plugin's files for the variable name to see what code employed it. Only the help files had that variable name, and that's how I realized what had happened.

lervag:
Ah, I see I forgot to add this to the changelog. I'll do that now.

Issue #650, opened 2017-01-02
Provide completion source based on cwl files from kile/texstudio


kiryph:
Would you consider adding a completion source for deoplete/neocomplete/youcomplete based on cwl files from kile https://cgit.kde.org/kile.git/tree/src/data/complete, latexing https://github.com/LaTeXing/LaTeX-cwl, or texstudio https://sourceforge.net/p/texstudio/hg/ci/default/tree/completion/

These are static files containing lists of latex commands. Nothing fancy.

This idea has already been applied to Sublime text plugins http://www.latexing.com/features.html, [LaTeXTools](https://github.com/SublimeText/LaTeXTools/pull/431) and for the Atom editor plugin exist a pull request (https://github.com/msiniscalchi/atom-latextools/pull/53).

I have googled a little bit more about cwl files. I think texstudio has extended its capabilities. Compare the description of cwl in kile and texstudio:

1. https://websvn.kde.org/*checkout*/tags/kile/2.0/src/README.cwl (cwl 0.1)
1. http://texstudio.sourceforge.net/manual/current/usermanual_en.html#CWLDESCRIPTION

For example, texstudio shows certain completions only in math environments. I do not know whether neocomplete can support this.

kiryph:
As Minimal Starting Point, one can add the cwl files as a dictionary for neocomplete:

1. Download kile's cwl files with  
 `$ wget -r -P ~/Documents/Mwe/vim-cwl-completion/kile/ -l1 -nH --cut-dirs=8 --no-parent -e robots=off https://cgit.kde.org/kile.git/plain/src/data/complete/tex/`
2. If you use `necomplete`, add following to your vimrc to enable autocompletion for commands of selected packages 
```
let g:neocomplete#sources#dictionary#dictionaries = {
  \ 'tex' : ''
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/amssymb.cwl,'
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/amsmath.cwl,'
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/biblatex.cwl,'
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/booktabs.cwl,'
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/graphicx.cwl,'
  \ .'~/Documents/Mwe/vim-cwl-completion/kile/latex-document.cwl,'
[...]

PR #770, opened 2017-04-09
Completion source for LaTeX commands

lervag:
Adds a completion source for LaTeX commands. Very simple implementation, but should still be pretty useful.

When finished, this should solve #650.


lervag:
Ok, so now it remains to generate good command completion sources. I think perhaps it is not desirable to include every imaginable command. Instead, I think, it would be more useful to only have the most relevant commands. Or, what do you guys think?

@kiryph @poppyschmo

Note: The current implementation supports:

- Loading class sources (prefixed with `class-`) and package sources
- `#include:` to include nested packages/classes
- Adding a mode indicator (this probably needs adjustment/improvement)
- Adds indicator to the completion menu about the package provider

The complete file syntax is this:

```
# include: package
command[ mode]
...
```

Here the mode string is not necessary, and if it is not provided, the command is completed in both normal and math mode.

The implementation lacks unicode symbols. I don't see how to add that in a simple manner. But if you guys would help me to generate the source files, then we could define a syntax such as this:

```
command, mode, symbol
```

Or perhaps there is a way to "infer" the symbol based on the command name?

Note: I've decided _not_ to use the `cwl` files directly, because they include very much more than just simply the command names. Even though it is some work to convert the `cwl` files to vimtex, I think it is worth the effort. But I also question the necessity of adding every possible command, and not just the frequently used ones.

kiryph:
Is there a possibility to setup neocomplete with this?

Regarding the conversion of latex commands to unicode. Following scripts might be helpful to create these files:

- http://mirrors.ctan.org/macros/latex/contrib/unicode-math/unicode-math-table.tex
- https://code.google.com/archive/p/unicodeitlinux/
- https://github.com/vikhyat/latex-to-unicode

`\right` and pressing `ctrl-x ctrl-o` shows me also candidates which do not start with `right`. I am not sure whether I like this. I think I appreciate this behaviour for citations.

lervag:
I use this to setup neocomplete (note the last line that will match commands):

```vim
      \ '\v\\%('
      \ . '\a*cite\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
      \ . '|\a*ref%(\s*\{[^}]*|range\s*\{[^,}]*%(}\{)?)'
      \ . '|hyperref\s*\[[^]]*'
      \ . '|%(include%(only)?|input)\s*\{[^}]*'
      \ . '|usepackage%(\s*\[[^]]*\])?\s*\{[^}]*'
      \ . '|documentclass%(\s*\[[^]]*\])?\s*\{[^}]*'
      \ . '|\a*'
      \ . ')'
```

I'll look into your suggested links for unicode conversion. Also, I agree that for this completion type, we should anchor the regex. I'm updating and pushing now.

kiryph:
I miss the completion of `\includegraphics` from https://cgit.kde.org/kile.git/plain/src/data/complete/tex/graphicx.cwl

demo-complete

poppyschmo:
The symbol stuff is pretty trivial. [This `.json` file](https://github.com/poppyschmo/deoplete-latex/blob/master/rplugin/python3/deoplete/resources/latest.json) includes most of the common unicode glyphs. Also, many of [these scripts](https://github.com/poppyschmo/deoplete-latex/tree/master/rplugin/python3/deoplete/resources) were made expressly to scrape various sources for these same symbols, though with more of an eye toward the "web"/js side of things.

lervag:
Thanks, @poppyschmo! I already figured out a lot of this stuff myself, e.g. from [this SO post](http://tex.stackexchange.com/questions/162385/an-extensive-list-of-latex-symbols-and-unicode-equivalents). Do you have any idea if your sources are better than the method from that post?

lervag:
Ok, I ended up combining symbols from several sources. I also made a script to add symbols to complete files. In any case, the current version should also show symbols when available.

lervag:
The question now is: Is the feature appropriate in itself? In that case, I propose to clean up and merge, independently from working on improved sources for command completion.

lervag:
Ok, I've ended up writing a parser for the `cwl` files and generated completion files based on `texstudio`s cwl files. I think we now have a large set of completion candidates, as well as symbols for most (relevant) commands.

What is lacking now is mode specific completion, but I think this should be fine for now.

If someone would be willing to learn the format of the cwl files from texstudio and explain how to parse the mode specifiers, then I could try to parse that as well and add mode specific command completion.

lervag:
I've merged this now. Feel free to open new issues, I think there are several things that can be improved still.

PR #775, opened 2017-04-11
Add Skim to the list of supported viewers

kiryph:


lervag:
And you even alphabetically sorted the list. :)

kiryph:
`:sort` is always there for me 😄 .

PR #799, opened 2017-04-22
Set did_indent flag only if enabled

patrick96:
Otherwise there is no way to have vimtex installed and use the built-in
indentexpr because did_indent would be set to 1 eventhough no indentexpr
was actually set


lervag:
Thanks!

PR #816, opened 2017-04-29
pplatex-based logfile parser

languitar:
This PR adds support for using `pplatex` to parse log files, which is more stable in detecting the real source file in multi-file documents in my cases.

`pplatex` requires that `-file-line-error` is NOT passed to the LaTeX compiler. Otherwise, errors are missed. I have documented this, but maybe there is a more automatic way to set this?

I have tested this with a recent git version of pplatex. One thing that remains is that I am unable to ignore the last few empty lines in the generated output using `errorformat`. Maybe someone who has more knowledge of this can have a look at what I have done wrong with the patterns?

### DIFF
--- a/autoload/vimtex/qf/pplatex.vim
+++ b/autoload/vimtex/qf/pplatex.vim
@@ -1,0 +1,76 @@
+" vimtex - LaTeX plugin for Vim
+"
+" Maintainer:   Karl Yngve Lervåg
+" Email:        karl.yngve@gmail.com
+" Initial work: Johannes Wienke
+" Email:        languitar@semipol.de
+"
+
+function! vimtex#qf#pplatex#new() " {{{1
+  return deepcopy(s:qf)
+endfunction
+
+" }}}1
+
+
+let s:qf = {
+      \ 'name' : 'LaTeX logfile using pplatex',
+      \}
+
+function! s:qf.init() abort dict "{{{1
+  " Each new item starts with two asterics followed by the file,
+  " potentially a line number and sometimes even the message itself is on the
+  " same line. Please note that the trailing whitspaces in the error formats
+  " are intentional as pplatex produces these.
+
+  " Start of new items with file and line number, message on next line(s).
+  setlocal errorformat=%E**\ Error\ \ \ in\ %f\\,\ Line\ %l:\ 
+  setlocal errorformat+=%W**\ Warning\ in\ %f\\,\ Line\ %l:\ 
+  setlocal errorformat+=%I**\ BadBox\ \ in\ %f\\,\ Line\ %l:\ 
+
+  " Start of new items with only a file.
+  setlocal errorformat+=%E**\ Error\ \ \ in\ %f:\ 
+  setlocal errorformat+=%W**\ Warning\ in\ %f:\ 
+  setlocal errorformat+=%I**\ BadBox\ \ in\ %f:\ 
+
+  " Start of items with with file and message on the same line. There are
+  " no BadBoxes reported this way.
+  setlocal errorformat+=**\ Error\ in\ %f:\ %m
+  setlocal errorformat+=**\ Warning\ in\ %f:\ %m
+
+  " Anything that starts with three spaces is part of the message from a
+  " previously started multiline error item.
+  setlocal errorformat+=%C\ \ \ %m
+
+  " Items are terminated with two newlines.
+  setlocal errorformat+=%-Z
+
+  " Skip statistical results at the bottom of the output.
+  setlocal errorformat+=%-GResult%.%#
+  setlocal errorformat+=%-G
+endfunction
+
+function! s:qf.setqflist(base, jump) abort dict "{{{1
+  if empty(a:base)
+    let l:log = b:vimtex.log()
+  else
+    let l:log = fnamemodify(a:base, ':r') . '.log'
+  endif
+
+  if empty(l:log)
+    call setqflist([])
+    throw 'Vimtex: No log file found'
+  endif
+
+  execute "setlocal" "makeprg=pplatex\\ -i\\ " . l:log
+
+  if a:jump
+    silent make l:log %:S
+  else
+    silent make! l:log %:S
+  endif
+endfunction
+
+" }}}1
+
+" vim: fdm=marker sw=2
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -93,6 +93,9 @@ Feature overview~
                                                               *vimtex-features*
 
 - Document compilation with `latexmk` or `latexrun`
+- LaTeX log parsing for quickfix entries using
+  - internal method
+  - `pplatex`
 - Compilation of selected part of document
 - Support for several PDF viewers with forward search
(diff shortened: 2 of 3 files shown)


languitar:
I am not sure whether this is related, but I also see that vimtex now autojumps to the first entry of the quickfix list after compilation despite `g:vimtex_quickfix_autojump` being set to 0.

languitar:
I have sorted out the errorformat issue. Parsing now works as expected in my case.

languitar:
I have also fixed the autojump issue.

languitar:
I still see the raw pplatex output flashing across the whole screen sometimes for a few milliseconds. No idea which part is responsible for that.

autoload/vimtex/qf/pplatex.vim
```diff
@@ -0,0 +1,67 @@
...
+  else
+    let l:log = fnamemodify(a:base, ':r') . '.log'
+  endif
+
+  if empty(l:log)
+    call setqflist([])
+    throw 'Vimtex: No log file found'
+  endif
+
+  execute "setlocal" "makeprg=pplatex\\ -i\\ " . l:log
```
lervag: I'm curious: Could you use simply `setlocal makeprg=pplatex\ -i`, then use two arguments for `:make`, i.e. `execute 'silent make' l:log '%:S'`?

autoload/vimtex/qf/pplatex.vim
```diff
@@ -0,0 +1,67 @@
...
+  setlocal errorformat+=%-GResult%.%#
+  setlocal errorformat+=%-G
+endfunction
+
+function! s:qf.setqflist(base, jump) abort dict "{{{1
+  if empty(a:base)
+    let l:log = b:vimtex.log()
+  else
+    let l:log = fnamemodify(a:base, ':r') . '.log'
+  endif
```
lervag: It could be necessary to use `fnameescape()` here if e.g. `l:log` contains a space.

autoload/vimtex/qf/pplatex.vim
```diff
@@ -0,0 +1,67 @@
...
+  setlocal errorformat+=%I**\ BadBox\ \ in\ %f\\,\ Line\ %l:\ 
+  " badbox entries without column
+  setlocal errorformat+=%I**\ BadBox\ \ in\ %f:\ 
+  " anything that starts with three spaces is part of the error message
+  setlocal errorformat+=%C\ \ \ %m
+  " messages are terminated with two newlines
+  setlocal errorformat+=%-Z
+  " skip result lines
+  setlocal errorformat+=%-GResult%.%#
+  setlocal errorformat+=%-G
```
lervag: This part (above) is a little bit "out of (my) style". I prefer to start comments with a capitol letter. I also prefer to group related code with a slightly larger comment first as well as separated by a single empty line. I would probably not have commented each line of errorformat. However, I understand (and agree) that the `errorformat` is a pretty obscure format to read, and so I can accept that we comment each line for transperancy. But then I want some more newlines between each group of comment + code.

autoload/vimtex/qf/pplatex.vim
```diff
@@ -0,0 +1,67 @@
+" vimtex - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
```
lervag: I would not mind if you want to add your own name/email here as well. It's up to you, of course, but obviously you deserve credit for writing this!

doc/vimtex.txt
```diff
@@ -991,6 +991,11 @@ Options~
     This is the standard method which parses the normal LaTeX output. See
     |g:vimtex_quickfix_latexlog| for configuration.
 
+  pplatex~
+    Uses `pplatex` to parse the LaTeX output file. `pplatex` requires that
+    `-file-line-error` is NOT passed to the LaTeX compiler. Adapt
+    |g:vimtex_compiler_latexmk| as necessary.
+
```
lervag: I would want some more information here, e.g. a short description of `pplatex` and a link to the main website. This would probably also warrent a line in the feature list (README.md and the top of the doc file).

lervag:
Regarding the flashing: I would suspect this occurs only in terminal Vim/nvim? I have not tested this feature myself yet, I will after you've responded to the initial review. But I think the problem is that the `:make` command will run a shell command. See `:h :make`:

>... This works almost like typing
>
>     ":!{makeprg} [arguments] {shellpipe} {errorfile}".

I think a solution to this is to avoid using the `:make` command, and instead first running the `pplatex` command either with a `:silent !...` or a `:silent system()` or similar, ensuring that the output is parsed to some temporary file. Then afterwards you can use the `:cfile`/`:cgetfile`similar to how it is done in `qf/latexlog.vim`. This way it should not be too difficult to avoid the flashing, I think.

autoload/vimtex/qf/pplatex.vim
languitar: Fun thing about this is that I copied that fragment from one of the other qf files ;) So it would also be broken there?

autoload/vimtex/qf/pplatex.vim
languitar: I can't get that to work. Nothing happens if I try your proposed solution.

languitar:
Apart from the `make` issue, I have tried to address your comments. However:

* For `fnameescape`: the code is taken from your native log parsing functions. It would need to be changed there as well. Can you have a look whether this is necessary? I have not enough knowledge about such vim internals
* I am only working in neovim, so I only see flickering there.

autoload/vimtex/qf/pplatex.vim
lervag: Yes, you're probably right. Let's leave this be, then, I'll consider to fix this myself later and apply in both qf files.

lervag:
I took the liberty of applying some minor changes before merging. There is still flickering, but it seems to work well now. Let me know what you think.

Note: I still think we can remove the flickering if we do as I suggested in my previous comment.

I also think it could be nice to do a similar trick as in the latexlog qf type to change the statusline title for the quickfix window. If you want, I could add this as well.

lervag:
By the way, thanks for the good work!

languitar:
Thanks for applying this! Feel free to add you proposed changed. I just don't have enough experience with vim scripting to get this going.

languitar:
Btw. one thing that would be very nice would be to automatically (or more easily) change the latex command line arguments to get rid of the file-line-error thing.

lervag:
I've updated, please see the latest version. It should not automatically remove the file-line-error option when one uses `latexmk`. Further, the quickfix window title should be improved and the flickering should be gone.

lervag:
Btw, I'm curious: Do you have an example where `pplatex` will now produce a better quickfix list than the latexlog method?

languitar:
Flickering is gone and changing the compiler also works well. I have just found, that there are still some more errorformat entries that pplatex might produce. I will create a new PR for this.

lervag:
Great, thanks for the new PR and for reminding me about the previous issue.

PR #834, opened 2017-05-12
Support cleveref \label[⟨type ⟩]{...} for foldtext

kiryph:
see http://mirrors.ctan.org/macros/latex/contrib/cleveref/cleveref.pdf#page=9


lervag:
Thanks!

PR #841, opened 2017-05-16
Recursive exploration fix

DelusionalLogic:
Vimtex discarded the first part of the subimport command, which wont do
for recursive searching. For the project i'm working on that meant we
ran into an infinite loop when you have recursive subimporting

### DIFF
--- a/autoload/vimtex/parser.vim
+++ b/autoload/vimtex/parser.vim
@@ -192,7 +192,7 @@ function! s:parser.input_line_parser_tex(line, file, re) abort dict " {{{1
   let l:subimport = 0
   if l:file =~# '\v\\%(sub)?%(import|%(input|include)from)'
     let l:candidate = self.input_parser(
-          \ substitute(l:file, '\\\w*\s*{[^{}]*}\s*', '', ''),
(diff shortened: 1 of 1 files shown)


lervag:
Thanks, I think I understand why this could be necessary. However, I would be very happy if you could provide a minimal example where this is needed or where this would fix a problem. That way I could test this myself and see that it does what it should do.

DelusionalLogic:
It requires the file you are subimporting be named the same as the current file, but also that the subimport isn't the first import of the file, since the issue needs the currentfile variable be overwritten.

It takes 3 files to replicate, how would you like them? I'm guessing not in the test directory, since that's ignored.

lervag:
I appreciate your concern for my time/burden; I am very happy when users provide well written and clear issue reports, and even more happy when the include suggested fixes/improvenets! Thanks!

First: I tend to make tests for issues/PRs inside the `test/issues/issue-#`. There is a template under `test/issues/template`, but you should note that the `minivimrc` has some customized settings for my personal use that may not be so intuitive.

As you will notice, I now added your test to the test directory myself to the master branch before merging your PR.

Issue #850, opened 2017-05-23
Vimtex incorrectly resets chapter count on part


DelusionalLogic:
LaTex doesn't reset chapter count when it starts a new part, Vimtex does though. That's unfortunate since it creates a disconnect between the heading numbers you see in the compiled document, and the toc vimtex generates.

The document below shows a toc like
```
Preamble
1 Chapter 1
1 Part 1
1.1 Chapter 2
```

What i'd expect to see:
```
Preamble
1 Chapter 1
1.1 Part 1
1.2 Chapter 2
```

The example document:
```tex
\documentclass{report}
\begin{document}

\chapter{Chapter 1}
Hello from Part 0

\part{Part 1}

\chapter{Chapter 2}
Hello from part 1

\end{document}

```

lervag:
FYI: This is related to #811, who proposes an option `g:vimtex_toc_reset_chapter_count_per_part` which, if enabled, will reset the chapter count for each part.

Before I update anything, I would be happy if we could agree what is the most natural behaviour, since this should the default behaviour.

lervag:
@kiryph Feel free to add your 2cents.

kiryph:
I agree with @DelusionalLogic w.r.t. to chapter numbering. I would prefer that the chapter numbers are NOT reset for a new part. Since this is the latex default, I would argue that the same should be true for vimtex.

However, I would prefer roman numbering for parts and I think there is no need to add the part number before the chapter. It should look like this:

```
      Preamble                                                                                                                                    0
I     First part                                                                                                                                  0
1   Hello                                                                                                                                       1
1.1 First section in chapter 1
2   World                                                                                                                                       1
II     Second part                                                                                                                                 0
3   Hello                                                                                                                                       1
4   Mars       
```

What options are necessary to make everyone happy, is not clear for me.

Right now, I would use latex defaults. The only option I would possibly enable for one of my documents would be `g:vimtex_toc_reset_chapter_count_per_part`. However, I could live without this.

DelusionalLogic:
I agree with everything @kiryph just said.

The part number of the identifier would indeed be superfluous, and roman numerals would be a great differentiator.

lervag:
Ok, I've fixed this now. Currently, no option, but the default should now work as you've suggested.

PR #851, opened 2017-05-23
Add testcase for issue 850

DelusionalLogic:
I hope this is how you want the testcase to be laid out.

Testcase for #850

### DIFF
--- a/test/issues/850/vimrc
+++ b/test/issues/850/vimrc
@@ -1,9 +1,12 @@
 set nocompatible
-let &rtp  = '~/.vim/bundle/FastFold,' . &rtp
 let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
 let &rtp .= ',~/.vim/bundle/vimtex/after'
 filetype plugin indent on
 syntax enable
 
-" Here we need to edit a large tex file, e.g. a thesis or long paper
-edit main.tex
+nnoremap q :qa!<cr>
+
+silent edit main.tex
+
+" Open the toc to inspect the chapter numbers
+VimtexTocOpen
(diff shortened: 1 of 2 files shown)


test/issues/850/vimrc
```diff
@@ -0,0 +1,9 @@
+set nocompatible
+let &rtp  = '~/.vim/bundle/FastFold,' . &rtp
+let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
+let &rtp .= ',~/.vim/bundle/vimtex/after'
+filetype plugin indent on
+syntax enable
+
+" Here we need to edit a large tex file, e.g. a thesis or long paper
+edit main.tex
```
lervag: If you prepend `silent` we don't need to type `<cr>` when starting the test.

test/issues/850/vimrc
```diff
@@ -0,0 +1,9 @@
+set nocompatible
+let &rtp  = '~/.vim/bundle/FastFold,' . &rtp
```
lervag: Do we really need fastfold here?

test/issues/850/vimrc
```diff
@@ -0,0 +1,9 @@
...
+let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
+let &rtp .= ',~/.vim/bundle/vimtex/after'
+filetype plugin indent on
+syntax enable
+
```
lervag: I like to map `q` to `:qall` in order to make it easy to exit the test. See e.g. the template for how I prefer to do this.

DelusionalLogic:
I tried to make it as low friction as possible by opening the TOC automatically as well.

I can't figure out how to make Vimtex not hijack q on TOC open, so you'll have to press q twice to exit. I hope that's ok, Otherwise you'll have to point me in the direction of how to do it properly.

lervag:
This is good, thanks. :)

PR #863, opened 2017-06-05
Improve package detection

andymass:
Package detection is currently deficient as it cannot handle lists such as
`\usepackage{one, two}`, or even multiple `\usepackage` statements on a
single line such as `\usepackage{one}\usepackage{two}`.  A more complex
regular expression is provided to allow packages to be detected in a
greater number of instances.

`vimtex` also does not currently recognize packages required by other
packages.  In principle, `vimtex`'s recursive parsing capability could be
used to detect such packages.  However, in this PR suggests a more robust
method, to detect the presence of an `fls` file, and simply use the `.sty` 
files listed there as the correct package list.

This PR changes package detection in two ways;

 - If an `fls` file exists (e.g., created by `latexmk`), it is parsed to 
      obtain a list of packages.  In this case, the packages found by 
      parsing the preamble will be ignored.
 - Preamble parsing is improved to handle multiple packages on one line

Additionally, this PR will also

 - Add documentclass to vimtex info
 - Add fls filename to vimtex info

Further known deficiencies:

 - The preamble package parsing does not recognize `\usepackage` spread
    over multiple lines, e.g.,
```      
   \usepackage[option1,
            option2]{package}
```
   I don't know how to get around this in the current parsing framework 
 - does not recognize hidden code in `\begin{comment}` or `\if` statements.  This is probably not worth the effort.

### DIFF
--- a/autoload/vimtex/state.vim
+++ b/autoload/vimtex/state.vim
@@ -28,6 +28,8 @@ function! vimtex#state#init() " {{{1
     call vimtex#qf#init_state(b:vimtex)
     call vimtex#toc#init_state(b:vimtex)
     call vimtex#labels#init_state(b:vimtex)
+
+    call b:vimtex.parse_fls()
   endif
 endfunction
 
@@ -55,6 +57,8 @@ function! vimtex#state#init_local() " {{{1
     call vimtex#qf#init_state(l:vimtex)
     call vimtex#toc#init_state(l:vimtex)
     call vimtex#labels#init_state(l:vimtex)
+
+    call b:vimtex.parse_fls()
   endif
 
   let b:vimtex_local = {
@@ -426,12 +430,43 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
       continue
     endif
 
-    let l:package = matchstr(l:line, '^\s*\\usepackage.*{\zs\w*\ze}')
+    " Find \usepackage[options]{package} statements
+    let l:pat = g:vimtex#re#not_comment . g:vimtex#re#not_bslash
+        \ . '\v\\usepackage\s*(\[[^[\]]*\])?\s*\{([^{}]+)\}'
+
+    let l:indx = match(l:line, l:pat)
+    while l:indx >= 0
+      let l:matches = matchlist(l:line, l:pat, l:indx)
+      for l:package in split(l:matches[2], '\s*,\s*')
+        let l:package = substitute(l:package, '^\s*\|\s*$', '', 'g')
+        if !empty(l:package)
+          let self.packages[l:package] = {}
+        endif
+      endfor
+      let l:indx += len(l:matches[0])
+      let l:indx = match(l:line, l:pat, l:indx)
+    endwhile
+  endfor
+endfunction
+
+" }}}1
+function! s:vimtex.parse_fls() abort dict " {{{1
+  let l:fls = self.fls()
+  if empty(l:fls) | return | endif
+
+  let l:fls_packages = {}
+
+  for l:line in vimtex#parser#fls(l:fls)
+    let l:package = matchstr(l:line, '^INPUT \zs.\+\ze\.sty$')
+    let l:package = fnamemodify(l:package, ':t')
     if !empty(l:package)
-      let self.packages[l:package] = {}
-      continue
+      let l:fls_packages[l:package] = {}
     endif
   endfor
+
+  if !empty(l:fls_packages)
+    let self.packages = l:fls_packages
+  endif
 endfunction
 
 " }}}1
@@ -460,6 +495,7 @@ function! s:vimtex.pprint_items() abort dict " {{{1
         \ ['out', self.out()],
         \ ['log', self.log()],
         \ ['aux', self.aux()],
+        \ ['fls', self.fls()],
         \]
 
   if !empty(self.engine)
@@ -469,6 +505,8 @@ function! s:vimtex.pprint_items() abort dict " {{{1
   if len(self.sources) >= 2
     call add(l:items, ['source files', self.sources])
   endif
+
+  call add(l:items, ['document class', self.documentclass])
 
   if !empty(self.packages)
     call add(l:items, ['packages', keys(self.packages)])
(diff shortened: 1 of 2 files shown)


autoload/vimtex/state.vim
```diff
@@ -28,6 +28,8 @@ function! vimtex#state#init() " {{{1
     call vimtex#qf#init_state(b:vimtex)
     call vimtex#toc#init_state(b:vimtex)
     call vimtex#labels#init_state(b:vimtex)
+
+    call b:vimtex.parse_fls()
```
lervag: Should add a similar call after line 59.

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
       continue
     endif
 
-    let l:package = matchstr(l:line, '^\s*\\usepackage.*{\zs\w*\ze}')
+    " strip comments
```
lervag: I like to use capital letters at beginning of comment

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
...
+    let l:line = substitute(l:line, '\(\\\)\@<!%.*$', '', '')
```
lervag: Could this use one of the regexes defined in `autoload/vimtex/re.vim`? If not, I think it would be good to add the regex there (at least the `\(\\\)\@<!%` part. I also think it would be good to use very magic, i.e. `%(\\)@<!\%`. Not, it might also be good to use the `vimtex#re#not_bslash` here, since it will search for double backslashes.

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
...
+
+    " find \usepackage[options]{package} statements
+    let l:pat = '\\usepackage\s*\(\[[^\[\]]*\]\)\?\s*{\s*\([^{}]\+\)\s*}'
```
lervag: Capital letter.

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
...
+endfunction
+
+" }}}1
+function! s:vimtex.parse_fls() abort dict " {{{1
+  let l:fls = self.fls()
+  if empty(l:fls)
+    return
+  end
+
+  let fls_packages = {}
```
lervag: Use `l:fls...`

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
...
+    endwhile
+  endfor
+endfunction
+
+" }}}1
+function! s:vimtex.parse_fls() abort dict " {{{1
+  let l:fls = self.fls()
+  if empty(l:fls)
+    return
+  end
```
lervag: For simple conditionals like this I prefer to use `|` to put them on a single line.

autoload/vimtex/state.vim
```diff
@@ -426,12 +428,46 @@ function! s:vimtex.parse_preamble() abort dict " {{{1
...
+
+  let fls_packages = {}
+
+  for l:line in vimtex#parser#fls(l:fls)
+    let l:package = matchstr(l:line, '^INPUT \zs.\+\ze\.sty$')
+    let l:package = fnamemodify(l:package, ':t')
     if !empty(l:package)
-      let self.packages[l:package] = {}
+      let fls_packages[l:package] = {}
       continue
```
lervag: This line does not do anything (but it might be a left over from myself, so this might not be your fault).

lervag: To be clear: The for loop continues in any case, so there is no reason to use `continue`.

andymass:
Thank you for your detailed comments.  I've made the changes and improved the usepackage regex using those already in `re.vim`,

> Is it really latexmk that generates the .fls files, or is it a by product of pdflatex?

You are correct, it is pdflatex which generates the .fls file when passed the `-recorder` option, which is currently passed by default in latexmk (and latexrun).

> Is it safe to assume that the .fls file (or a similar type of file) will exist for most compiled LaTeX documents?

It is reasonably safe in modern versions of pdflatex, as long as the user does not specifically disable it.  In some versions/implementations, the .fls file was named incorrectly and did not respect -output-directory.  I do not know which versions and platforms are affected.  Hence, I felt that the basic preamble parsing was still useful.

> If so, could we perhaps parse this file after the first compilation (e.g. through a hook) in order to set the loaded packages?

I think this is a good idea, but I would have to study the vimtex source to determine the best way to create such a hook.  Would it be reasonable to re-parse the fls file after each compile? 

> But I'm curious if the preample parsing is necessary when the .fls parsing is there? Preferably, I think it would be better to just have a single mechanism.

I understand, and it's up to your discretion.  Ideally, it would have to be determined which versions of (pdf,xe,etc)latex create the correct fls, and whether other versions are old enough to not support.  It is still technically possible that the user disables -recorder in their `.latexmk`.  Also, there is still the issue of editing a .tex file which has not been compiled yet.

lervag:
I've merged this. It passes tests (I've just fixed some issues with the testing system, so tests should run as usual again now), and I think it looks like a good addition.

However, I think we should improve it. In particular, I think that we could add a callback after compilation that updates the source packages. If you see in `compiler.vim`, there is  function `vimtex#compiler#callback`. This function could call the `b:vimtex.parse_fls()` function in order to update the package list.

lervag:
Feel free to add a PR for this, or an issue with reference here.

lervag:
I think we should have both preamble parsing and fls parsing. It could be possible to do a recursive search through all packages to generate the list, but I think it is better to rely on already generated data.

So, thanks for this! I think it is a nice addition!

One question: Should we, or could we, also parse documentclass(es) from the fls file?

andymass:
I think it is not necessary to parse the documentclass from the fls because parsing the `\documentclass` statement is much easier than parsing the package statements (unless there is an example of documentclass not being determined correctly).  Also, the documentclass rarely changes when one is writing a document and if it is changed, the user simply has to issue `\lx` to update it.

lervag:
Yes, you're right. I agree.

PR #870, opened 2017-06-11
Bug in `s:qf_is_open()` fixed

Aster89:
The `filter()` function accepts a List or Dictionary as the first
argument, so `l:buflist` has to be first transformed into a list by
means of `split()` and then can be passed to `filter()`. (Btw, the former
is executed "in situ" in the latter.)


lervag:
Sorry about that; should have tested before I pushed. In any case, thanks for fixing! :)

PR #888, opened 2017-06-28
Keep cursor in the LaTeX window when opening quickfix with g:vimtex_quickfix_mode=2

arekfu:
Hi,

I noticed that vimtex did not honour the `g:vimtex_quickfix_mode=2` option, in the sense that the cursor would jump to the quickfix window as soon as compilation finished (in continuous mode). This was quite annoying as it broke my workflow. I had a look at the code and I discovered (I think) two minor bugs:

1. The `s:fix_paths()` function cannot access `b:vimtex.root` because, by the time it is called, vim has already switched to the quickfix buffer and `b:vimtex` is not defined;
2. For the same reason, the `s:window_save()` function is being called too late.

This pull request fixes the bugs above and restores the expected behaviour of `g:vimtex_quickfix_mode=2`. Please let me know if you think this makes sense. Perhaps I should have opened an issue first?

Davide

### DIFF
--- a/autoload/vimtex/qf.vim
+++ b/autoload/vimtex/qf.vim
@@ -38,6 +38,7 @@ function! vimtex#qf#open(force) abort " {{{1
   if !exists('b:vimtex.qf.setqflist') | return | endif
   cclose
 
+  call s:window_save()
   try
     call b:vimtex.qf.setqflist('', g:vimtex_quickfix_autojump)
   catch /Vimtex: No log file found/
@@ -66,7 +67,6 @@ function! vimtex#qf#open(force) abort " {{{1
         \     && (s:qf_has_errors() || g:vimtex_quickfix_open_on_warning))
 
   if l:do_open
-    call s:window_save()
     botright cwindow
     if g:vimtex_quickfix_mode == 2
       call s:window_restore()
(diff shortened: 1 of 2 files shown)


lervag:
I'm not able to reproduce the problem. That is, if I test on a simple example that produces an error with a minimal vimrc file with `let g:vimtex_quickfix_mode = 2`, then the cursor moves as expected. I've tested with Vim.

lervag:
Can't reproduce with neovim either.

lervag:
I think your first fix was correct, though, and so I've added it manually. Thanks for noticing and suggesting a fix!

arekfu:
Thanks for looking into this. I am confused by what you wrote above:

> That is, if I test on a simple example that produces an error with a minimal vimrc file with `let g:vimtex_quickfix_mode = 2`, then the cursor moves as expected. I've tested with Vim.

My understanding was that the cursor should *not* move when `g:vimtex_quickfix_mode = 2`. Am I misunderstanding something?

lervag:
Sorry, I was unclear. What I should have written is that when I test this on a minimal example, the cursor stays put as expected.

Please update vimtex, then go to `<VIMTEX>/test/features/error-messages` and start Vim with `vim --servername VIM -u minivimrc` or `nvim -u minivimrc` or `gvim -u minivimrc` (whichever you prefer). Then start compiling with `\ll` or `:VimtexCompile`.

The observed behaviour, on my end, is that the cursor stays put as expected. I've tested with `neovim 0.2.0` and `Vim 8.0.628`.

PR #890, opened 2017-06-30
Fix s:vimtex.parse_engine()

c4rlo:
It broke in f74c984.


lervag:
Thanks, and sorry for adding the bug!

PR #900, opened 2017-07-08
Add call in compiler callback to parse fls on compilation

andymass:
This simple change causes vimtex to attempt to parse the fls file (if it exists) after each compilation, which updates the package list.  Cf. #863.  Note, I do not check to see if compilation was successful because this matches the behavior of opening a project that has an fls file.  I think the performance impact should be small because matching on the fls file is simple (unlike preamble parsing).

### DIFF
--- a/autoload/vimtex/compiler.vim
+++ b/autoload/vimtex/compiler.vim
@@ -73,6 +73,10 @@ function! vimtex#compiler#callback(status) abort " {{{1
   call vimtex#echo#status(['compiler: ',
         \ a:status ? ['VimtexSuccess', 'success'] : ['VimtexWarning', 'fail']])
 
+  if a:status
+    call b:vimtex.parse_packages_from_fls()
+  endif
+
   for l:hook in g:vimtex_compiler_callback_hooks
     execute 'call' l:hook . '(' . a:status . ')'
   endfor
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -36,6 +36,7 @@ CONTENTS                                                      *vimtex-contents*
     Support for multi-file projects             |vimtex-multi-file|
     Comment on internal tex plugin              |vimtex-comment-internal|
     Support for TeX specifiers                  |vimtex-tex-directives|
+    Package detection                           |vimtex-package-detection|
   Usage                                         |vimtex-usage|
     Default mappings                            |vimtex-default-mappings|
     Options                                     |vimtex-options|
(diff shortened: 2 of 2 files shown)


lervag:
Thanks! I'll merge, although some thoughts first:

* Should we add something about this in the docs? Considering this does make the clientserver/callback feature even more useful, it could be relevant to mention it in the docs..?

* Would it be OK to only parse after a successful compile? This would prevent (possibly) unexpected errors if a faulty `fls` file should be compiled if there are errors in the manuscript. But perhaps this never really happens..?

* Have you tested this yourself on large documents? I also think this has negligable impact on performance, but if I'm wrong, then this might lead to some strange "lockups" during editing during the callback calls.

andymass:
All good points, thanks.

- Yes, I can read through the docs and figure out what's appropriate to add, and include it on this PR.

- It would be also fine to check for successful compilation.  The only reason not to is for consistency- when opening a tex document, vimtex cannot check whether the last compilation was successful or not.  Interestingly, the fls file is created even when there are errors.  I don't know the implementation details, but it appears pdflatex is able to figure out which files should be included even if errors occur earlier than the paths appear.

- In my test, on a (partial) thesis, parsing an 406 line fls file containing 165 .sty lines (82 unique .sty lines) takes around 12 milliseconds.  Though, my computer is fairly fast.

```
FUNCTION  vimtex#compiler#callback()
Called 1 time
Total time:   0.028121
 Self time:   0.000078

count  total (s)   self (s)
    1              0.000004   if get(b:vimtex.compiler, 'silence_next_callback')
                                let b:vimtex.compiler.silence_next_callback = 0
                                return
                              endif
                            
    1   0.016371   0.000005   call vimtex#qf#open(0)
    1              0.000003   redraw
                            
    1              0.000003   if exists('s:output')
                                call s:output.update()
                              endif
                            
    1   0.000121   0.000009   call vimtex#echo#status(['compiler: ', a:status ? ['VimtexSuccess', 'success'] : ['VimtexWarning', 'fail']])
                            
    1   0.011544   0.000010   call b:vimtex.parse_packages_from_fls()
                            
    2              0.000005   for l:hook in g:vimtex_compiler_callback_hooks
    1   0.000057   0.000026     execute 'call' l:hook . '(' . a:status . ')'
    1              0.000000   endfor
                            
    1              0.000001   return ''

FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
    1   0.028121   0.000078  vimtex#compiler#callback()

FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
    1   0.028121   0.000078  vimtex#compiler#callback()
```

lervag:
Great, thanks for the update. I'll merge when we have agreed on a good update to the docs. :)

andymass:
Sorry, I made some changes a while ago but never gave you an update.  Let me know if you have any concerns.

Btw, I think eventually the doc could be clarified as to what exactly is meant by "callback," since with job_start `compiler#callback` is called though not through clientserver.

lervag:
Yes, `job_start` also allows a callback e.g. when a process has exited, and I use this to call the `compiler#callback` function. This is necessary when one is not using continuous mode. And most people don't read the source code, so I don't think it is necessary to update the docs on this. Except, of course, that any change that makes the docs clearer are of course good.

lervag:
Thanks a bunch for your good updates to the docs! I'm merging this as we speak. I might make some minor updates myself, feel free to comment or complain when I do. :)

PR #908, opened 2017-07-19
changed command to call texcount on windows

JackeJR:
Issue:
- On issuing `VimtexCountWords`, I get "The system cannot find the path specified."
- `texcount` is in path and can be called via commandline as well as via `!texcount`
- issue persist with minimal `vimrc`.
- Windows 10 Pro x64, TeXLive 2017

Change:
- changed command separator (from ';' to '&') for windows to allow proper calling of texcount.


lervag:
Thanks! I prefer a less redundant style of coding, so I just pushed your suggested change with some minor modifications. Hope you don't mind.

JackeJR:
no problem. your code is much more elegant and I prefer it. learned something new today, thanks.

lervag:
Great! Btw: I really appreciate that you submitted a PR to fix your problem. Thanks!

PR #928, opened 2017-07-30
Add an ignorecase option for bib completion?

andymass:
What do you think about allowing a setting for ignoring case while completing bibliography entries?  Default behavior (case is matched according to '&ignorecase') is preserved.

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -158,7 +158,11 @@ function! s:completer_bib.search(regexp) dict " {{{2
     let lines = split(substitute(join(lines, "\n"),
           \ '\n\n\@!\(\s\=\)\s*\|{\|}', '\1', 'g'), "\n")
 
-    for line in filter(lines, 'v:val =~ a:regexp')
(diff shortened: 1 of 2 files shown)


lervag:
Just curious: Why do you want this? I guess I'll accept this, but I'm a little bit uncertain why the default is not OK; when would you want different behaviour from `&ignorecase`?

andymass:
To me, it makes little sense to care about case when completing bib entries; when typing `\cite{geometry<c-x><c-o>` I would expect all articles with the word geometry/Geometry in the title.  I would argue that `=~?` should _always_ be used for bib completion in vimtex.

However, others might prefer the current behavior (especially since vim defaults to `noignorecase`).  Hence, the option separation.

I do not want to `set noignorecase` globally is because case sensitive-matching is better for most other code uses.

lervag:
I think I agree that it makes sense to complete case insensitive, and thus I'd prefer not to make a new option. I'm pushing the minor change required for this, and I think that should solve your issue.

If anyone should find this breaks their workflow, they should open an issue, and in this case, I could introduce an option.

Issue #933, opened 2017-08-03
vimtex makes <plug> mappings even though I have provided my own


andymass:
### Explain the issue

Basically, I would like to remap `tsd` and friends to something else because I frequently use vim's "till" command.  So, in the vimrc below I have provided my own.  The documentation suggests that if I have provided a custom mapping, vimtex will not create it's default one.  However, in the MWE below `:map ts` clearly shows the mappings are still being created.

````
x  tsd          @<Plug>(vimtex-delim-toggle-modifier)
n  tsd          @<Plug>(vimtex-delim-toggle-modifier)
n  tsc          @<Plug>(vimtex-cmd-toggle-star)
n  tse          @<Plug>(vimtex-env-toggle-star)
````

### Minimal working example

```tex
\documentclass{minimal}
\begin{document}
Hello World!
\end{document}
```

### Minimal vimrc file

```vim
set nocompatible

" Load Vimtex
let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'

filetype plugin indent on
syntax enable

" Vimtex options go here
nmap cd <plug>(vimtex-delim-toggle-modifier)
xmap cd <plug>(vimtex-delim-toggle-modifier)
nmap cO <plug>(vimtex-env-toggle-star)
nmap co <plug>(vimtex-cmd-toggle-star)
```

lervag:
Ah, this was supposed to work, but for some reason I forgot to check `hasmapto()` on the right-hand side. Fixed now.

andymass:
Thank you!  :smile: :+1:

kiryph:
I am also not very happy with the default mappings `tse,tsc,tsd` for the same reason (they overwrite `ts` (move till next `s`)).

So first of all thanks for this update and raising this issue.

I am already thinking for a while about good alternatives which do not clash with builtin mappings or other plugins I use.

1) The suggestion by @andymass
```
nmap cd <plug>(vimtex-delim-toggle-modifier)
xmap cd <plug>(vimtex-delim-toggle-modifier)
nmap cO <plug>(vimtex-env-toggle-star)
nmap co <plug>(vimtex-cmd-toggle-star)
```
[...]

PR #934, opened 2017-08-05
Add a count to delimiter modifier toggle command

andymass:
What does a count currently do for the delimiter modifier commands, e.g. `2tsd`?  
Two goals of this PR: 
  1. Allow `vimtex-delim-toggle-modifier` (and analogous visual) to take a count, which causes the modifier to be incremented multiple times.
  2. Create a `vimtex-delim-toggle-modifier-reverse` (and analogous visual) which decrements the delimiter modifier.

These together are very useful in the following example.  Suppose you have the following set of modifiers
```vim
let g:vimtex_delim_toggle_mod_list = [
      \ ['\bigl', '\bigr'],
      \ ['\Bigl', '\Bigr'],
      \ ['\biggl', '\biggr'],
      \ ['\Biggl', '\Biggr'],
      \]
```
and the tex fragment,
```tex
\[   \bigl(  x  \bigr)   \].
```
Then, pressing `2tsd` will cause `\big` to be replaced with `\bigg` (i.e., 2 steps bigger).  Pressing `tsD` will replace `\bigg` with `\Big`.  This lets you easily increment/decrement to find the correct delimiter size.  vim-repeat `.` works as expected.

The count/decrement also work for visual mode.  Unfortunately, it is not clear how to allow repeat to work properly in the visual case.  First, vimtex reselects the region, which causes `.` to not work.  Also vimtex does not correctly preserve the SIZE of the region which traditionally in vim would follow the cursor, and this could be done in character or block mode.  A general implementation will be much more complicated and open to some debate.  In this PR, I just kill the repeat in visual mode.

Note there is some discussion of toggle delimiter behavior in #933.  Regardless of outcome of any future mapping change, this PR is a simple generalization of the current functionality, and (as far as I know) does not step on any existing use of [count] in the context of these mappings.

### DIFF
--- a/autoload/vimtex/delim.vim
+++ b/autoload/vimtex/delim.vim
@@ -6,10 +6,16 @@
 
 function! vimtex#delim#init_buffer() " {{{1
   nnoremap <silent><buffer> <plug>(vimtex-delim-toggle-modifier)
-        \ :call vimtex#delim#toggle_modifier()<cr>
+        \ :<c-u>call vimtex#delim#toggle_modifier()<cr>
+
+  nnoremap <silent><buffer> <plug>(vimtex-delim-toggle-modifier-reverse)
+        \ :<c-u>call vimtex#delim#toggle_modifier({'dir': -1})<cr>
 
   xnoremap <silent><buffer> <plug>(vimtex-delim-toggle-modifier)
         \ :<c-u>call vimtex#delim#toggle_modifier_visual()<cr>
+
+  xnoremap <silent><buffer> <plug>(vimtex-delim-toggle-modifier-reverse)
+        \ :<c-u>call vimtex#delim#toggle_modifier_visual({'dir': -1})<cr>
 
   nnoremap <silent><buffer> <plug>(vimtex-delim-change-math)
         \ :call vimtex#delim#change_prompt()<cr>
@@ -57,17 +63,27 @@ endfunction
 
 " }}}1
 function! vimtex#delim#toggle_modifier(...) " {{{1
-  let [l:open, l:close] = a:0 == 2
-        \ ? [a:1, a:2]
+  let l:args = a:0 > 0 ? a:1 : {}
+  call extend(l:args, {
+      \ 'count': v:count1,
+      \ 'dir': 1,
+      \ 'repeat': 1,
+      \ 'openclose': [],
+      \ }, 'keep')
+
+  let [l:open, l:close] = !empty(l:args.openclose)
+        \ ? l:args.openclose
         \ : vimtex#delim#get_surrounding('delim_modq_math')
   if empty(l:open) | return | endif
+
+  let l:direction = l:args.dir < 0 ? -l:args.count : l:args.count
 
   let newmods = ['', '']
   let modlist = [['', '']] + get(g:, 'vimtex_delim_toggle_mod_list',
         \ [['\left', '\right']])
   let n = len(modlist)
   for i in range(n)
-    let j = (i + 1) % n
+    let j = (i + l:direction) % n
     if l:open.mod ==# modlist[i][0]
       let newmods = modlist[j]
       break
@@ -97,13 +113,23 @@ function! vimtex#delim#toggle_modifier(...) " {{{1
         \ . strpart(line, l:cnum + len(l:close.mod) - 1)
   call setline(l:close.lnum, line)
 
-  silent! call repeat#set("\<plug>(vimtex-delim-toggle-modifier)", v:count)
+  if l:args.repeat
+    silent! call repeat#set("\<plug>(vimtex-delim-toggle-modifier"
+          \ . (l:direction < 0 ? '-reverse' : '') .')', l:args.count)
+  endif
 
   return newmods
 endfunction
 
 " }}}1
-function! vimtex#delim#toggle_modifier_visual() " {{{1
+function! vimtex#delim#toggle_modifier_visual(...) " {{{1
+  let l:args = a:0 > 0 ? a:1 : {}
+  call extend(l:args, {
+      \ 'count': v:count1,
+      \ 'dir': 1,
+      \ 'reselect': 1,
+      \ }, 'keep')
+
   let l:save_pos = vimtex#pos#get_cursor()
   let l:start_pos = getpos("'<")
   let l:end_pos = getpos("'>")
@@ -136,7 +162,12 @@ function! vimtex#delim#toggle_modifier_visual() " {{{1
     if !empty(l:close)
 
       if l:end_pos_val >= vimtex#pos#val(l:close) + strlen(l:close.match) - 1
-        let l:newmods = vimtex#delim#toggle_modifier(l:open, l:close)
+        let l:newmods = vimtex#delim#toggle_modifier({
+              \ 'repeat': 0,
+              \ 'count': l:args.count,
+              \ 'dir': l:args.dir,
+              \ 'openclose': [l:open, l:close],
+              \ })
 
         let l:col_diff  = (l:open.lnum == l:end_pos[1])
               \ ? strlen(newmods[0]) - strlen(l:open.mod) : 0
@@ -159,7 +190,9 @@ function! vimtex#delim#toggle_modifier_visual() " {{{1
   call setpos(l:swapped? "'>" : "'<", l:start_pos)
   call setpos(l:swapped? "'<" : "'>", l:end_pos)
   call vimtex#pos#set_cursor(l:save_pos)
-  normal! gv
+  if l:args.reselect
+    normal! gv
+  endif
 endfunction
 
 " }}}1
@@ -175,7 +208,6 @@ function! vimtex#delim#change(open, close, new) " {{{1
     let l:index = index(map(copy(g:vimtex#delim#lists.delim_math.name),
           \   'v:val[' . l:side . ']'),
           \ a:new)
-    echo "\n" l:index
     if l:index >= 0
       let [l:beg, l:end] = g:vimtex#delim#lists.delim_math.name[l:index]
     else
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -375,6 +375,8 @@ function! s:init_default_mappings() " {{{1
   call s:map('n', 'csd', '<plug>(vimtex-delim-change-math)')
   call s:map('n', 'tsd', '<plug>(vimtex-delim-toggle-modifier)')
   call s:map('x', 'tsd', '<plug>(vimtex-delim-toggle-modifier)')
+  call s:map('n', 'tsD', '<plug>(vimtex-delim-toggle-modifier-reverse)')
+  call s:map('x', 'tsD', '<plug>(vimtex-delim-toggle-modifier-reverse)')
   call s:map('i', ']]',  '<plug>(vimtex-delim-close)')
 
   if get(g:, 'vimtex_compiler_enabled', 0)
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -135,7 +135,7 @@ Feature overview~
   - Change the surrounding command, environment or delimiter with
     `csc`/`cse`/`cs$`/`csd`
   - Toggle starred command or environment with `tsc`/`tse`
-  - Toggle between e.g. `()` and `\left(\right)` with `tsd`
+  - Toggle between e.g. `()` and `\left(\right)` with `tsd`/`tsD`
   - Close the current environment/delimiter in insert mode with `]]`
   - Insert new command with `<F7>`
   - Convenient insert mode mappings for faster typing of e.g. maths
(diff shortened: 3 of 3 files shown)


kiryph:
@andymass Do you also know the plugin https://github.com/bootleq/vim-cycle?

Consider following mwe-vimrc:
```viml
" invoke with
" $ vim -u ~/.vim/mwe-vimrc
set nocompatible
let g:dein#types#git#clone_depth = 1
set runtimepath+=$HOME/.vim/bundle/repos/github.com/Shougo/dein.vim
call dein#begin("$HOME/.vim/bundle/")
call dein#add('Shougo/dein.vim')
call dein#add('bootleq/vim-cycle')
call dein#end()

nmap  <C-A> <Plug>CycleNext
nmap  <C-X> <Plug>CyclePrev
vmap  <C-A> <Plug>CycleNext
vmap  <C-X> <Plug>CyclePrev
nnoremap <silent> <Plug>CycleFallbackNext <C-A>
nnoremap <silent> <Plug>CycleFallbackPrev <C-X>
" if you use speeddating by tpope
" noremap <silent> <Plug>CycleFallbackNext <Plug>SpeedDatingUp
" noremap <silent> <Plug>CycleFallbackPrev <Plug>SpeedDatingDown

" requires amsmath for `Vert` and `vert`
let g:cycle_default_groups_for_tex = [
        \   [['tiny', 'scriptsize', 'footnotesize', 'small', 'normalsize', 'large', 'Large', 'LARGE', 'huge', 'Huge'], 'hard_case', 'match_case'],
        \   [['displaystyle', 'scriptstyle', 'scriptscriptstyle', 'textstyle']],
        \   [['part', 'chapter', 'section', 'subsection', 'subsubsection', 'paragraph', 'subparagraph']],
        \   [['article', 'report', 'book', 'letter', 'slides']],
        \   [['a4paper', 'a5paper', 'b5paper', 'executivepaper', 'legalpaper', 'letterpaper']],
        \   [['scrbook', 'scrreprt', 'scrartcl', 'scrlttr2']],
        \   [['oneside', 'twoside']],
        \   [['onecolumn', 'twocolumn']],
        \   [['draft', 'final']],
        \   [['(:)', '[:]', '\{:\}', '\lvert :\rvert', '\lVert :\rVert', '\lfloor :\rfloor', '\lceil :\rceil', '\langle :\rangle'], 'sub_pairs', 'hard_case', 'match_case'],
        \   [['\big(:\big)', '\Big(:\Big)', '\bigg(:\bigg)', '\Bigg(:\Bigg)'], 'sub_pairs', 'hard_case', 'match_case'],
        \   [['\big[:\big]', '\Big[:\Big]', '\bigg[:\bigg]', '\Bigg[:\Bigg]'], 'sub_pairs', 'hard_case', 'match_case'],
        \   [['\big\{:\big\}', '\Big\{:\Big\}', '\bigg\{:\bigg\}', '\Bigg\{:\Bigg\}'], 'sub_pairs', 'hard_case', 'match_case'],
        \   [['\big\l:\big\r', '\Big\l:\Big\r', '\bigg\l:\bigg\r', '\Bigg\l:\Bigg\r'], 'sub_pairs', 'hard_case', 'match_case'],
        \   [['\big', '\Big', '\bigg', '\Bigg'], 'hard_case', 'match_case'],
        \ ]
```

With this you can use ctrl-a/x to cycle through delimiter types and delimiter size specifiers, if the cursor is on either a delimiter or a size specifier.

For example starting with following text:
```tex
\Big(a+b\Big)
```
pressing ctrl-a on `(` will change it to
```tex
\Big[a+b\Big]
```
move now the cursor on a character of `\Big` and press ctrl-a, you will get
```tex
\bigg[a+b\bigg]
```
Another nice use case is increasing or decreasing font size specifiers such as `\footnotesize`. Please try for yourself, if this is something you like to use.

BTW: toggling something for me changing the state of a two-state system (true <-> false).

andymass:
@kiryph  No, I was not aware of vim-cycle.  Thank you!  It looks incredibly useful and I actually searched for a while for something like this but couldn't find it, and considered writing my own plugin.  I will be definitely be using it.

I agree with you on the inaccurately named "toggle," but incrementing the modifier is vimtex's current behavior.  I think it terminology was a hold-over from when there were only two modifiers, `\left(` or `(`, but imo the current functionality is more useful than simple on/off.  This PR merely allows you to combine multiple steps into one, instead of `tsdtsd` you can do `2tsd`.

This PR does have one nice advantage over vim-cycle, in that it lets your cursor be anywhere inside a group of matching delimiters instead of having to be on top of the word `\big`:
```
\bigl[a+ _ b\bigr]          % _ = cursor
\biggl[a+ _ b\biggr]      % after 2tsd
```

kiryph:
I wasn't aware of `g:vimtex_delim_toggle_mod_list` and therefore I have not used this feature of vimtex. Your PR is certainly a useful extension for this vimtex feature.

I just noticed your PR by your reference to the last issue we were on and I was assuming you might like to know about cycle.vim.

However, I doubt that I will use `tsd` for changing the delimiter sizes myself. `ctrl-a/x` feels more natural to me. I typically use `tsd` to add `\left/\right` when I am too lazy to type them myself. 

I haven't used the package `mleftright` yet, but now I am wondering whether I want `tsd` to cycle through `(,left(,mleft(` or add another cycle.vim group to change from `left/right` to `mleft/mright` with `ctrl-a/x`. IMO this is about orthogonality of the features otherwise one of them feels redundant.

---
BTW, I have a tex question. Why do you write `\bigl[a+b\bigr]` instead of `\big[a+b\big]`? According to [mathmode](http://mirrors.ctan.org/obsolete/info/math/voss/mathmode/Mathmode.pdf) `bigl` and `bigr` make only sense if you invert the usage of opening and closing delimiters.

andymass:
If I understand correctly, bootleq/vim-cycle cannot handle the following cases:
`( )` becomes `\bigl(  \bigr)` and
`\cmd{foo}` becomes \cmd[\big]{foo}` (uses mathtools, see below)
Is that correct?

> Why do you write \bigl[a+b\bigr] instead of \big[a+b\big]

Basically, my understanding of good tex practice differs from that advocated in `mathmode.pdf`.  See this 
stackexchange answer for a comparison: https://tex.stackexchange.com/a/144439/.  To me, `\bigl(` clearly typesets better and is semantically more informative.

FWIW, I actually frequently use mathtools [paired delimiters](https://tex.stackexchange.com/questions/94410/easily-change-behavior-of-declarepaireddelimiter), e.g, `\parens[\big]{a+b}` (would roughly make `\bigl( a+b \bigr)`).   To be sure, mathtools has its own crustiness and issues, but at least it is more DRY (don't-repeat-yourself).  With vimtex's `tsd` (well, I remap it to something else), there is hardly any difference in coding speed between the two.  In fact, you could easily bootleq/vim-cycle to `<c-a>` increment the single `\big`->`\Big` or even create a mapping that detects when you're in a `\parens[\big]{a+b}` command using `vimtex#cmd#get_current()` and performs the increment,  (which is exactly what I do).

lervag:
Thanks, @andymass, I think this is a nice addition! I'll look through the code and comment or merge.

Thanks, @kiryph. I was not aware of vim-cycle. It seems useful! I am also a fan of plugins that expand on built-in features in a natural manner.

autoload/vimtex/delim.vim
```diff
@@ -6,11 +6,17 @@
 
 function! vimtex#delim#init_buffer() " {{{1
   nnoremap <silent><buffer> <plug>(vimtex-delim-toggle-modifier)
-        \ :call vimtex#delim#toggle_modifier()<cr>
+        \ :<c-u>call vimtex#delim#toggle_modifier()<cr>
```
lervag: You don't need `<c-u>` for normal mode maps, do you?

autoload/vimtex/delim.vim
```diff
@@ -97,13 +108,19 @@ function! vimtex#delim#toggle_modifier(...) " {{{1
         \ . strpart(line, l:cnum + len(l:close.mod) - 1)
   call setline(l:close.lnum, line)
 
-  silent! call repeat#set("\<plug>(vimtex-delim-toggle-modifier)", v:count)
+  if get(l:args, 'repeat', 1)
+    silent! call repeat#set("\<plug>(vimtex-delim-toggle-modifier"
+          \ . (l:direction < 0 ? '-reverse' : '') .')', v:count)
```
lervag: You are now using both `l:count` and `v:count`. I'm curious, should one instead pass `l:count` here?

autoload/vimtex/delim.vim
```diff
@@ -175,7 +198,6 @@ function! vimtex#delim#change(open, close, new) " {{{1
     let l:index = index(map(copy(g:vimtex#delim#lists.delim_math.name),
           \   'v:val[' . l:side . ']'),
           \ a:new)
-    echo "\n" l:index
     if l:index >= 0
```
lervag: Thanks! :)

autoload/vimtex/delim.vim
```diff
@@ -159,7 +178,11 @@ function! vimtex#delim#toggle_modifier_visual() " {{{1
   call setpos(l:swapped? "'>" : "'<", l:start_pos)
   call setpos(l:swapped? "'<" : "'>", l:end_pos)
   call vimtex#pos#set_cursor(l:save_pos)
-  normal! gv
+  if get(l:args, 'reselect', 1)
```
lervag: Is this ever used? Do you use it for personal custom mappings?

autoload/vimtex/delim.vim
```diff
@@ -136,7 +153,9 @@ function! vimtex#delim#toggle_modifier_visual() " {{{1
     if !empty(l:close)
 
       if l:end_pos_val >= vimtex#pos#val(l:close) + strlen(l:close.match) - 1
-        let l:newmods = vimtex#delim#toggle_modifier(l:open, l:close)
+        let l:newmods = vimtex#delim#toggle_modifier({'count': l:count,
```
lervag: I prefer to keep each dict entry on a separate line for cases like this, i.e.:

```vim
let l:newmods = vimtex#delim#toggle_modifier({
      \ 'count': l:count,
      \ 'dir': get(l:args, 'dir', 1), 'repeat': 0,
      \ 'openclose': [l:open, l:close]
      \})
```

I know it adds some unnecessary lines, but I find it makes it easier to see the main contents at a glance.

autoload/vimtex/delim.vim
```diff
@@ -159,7 +178,11 @@ function! vimtex#delim#toggle_modifier_visual() " {{{1
   call setpos(l:swapped? "'>" : "'<", l:start_pos)
   call setpos(l:swapped? "'<" : "'>", l:end_pos)
   call vimtex#pos#set_cursor(l:save_pos)
-  normal! gv
+  if get(l:args, 'reselect', 1)
+    normal! gv
+  endif
+
+  silent! call repeat#set('')
```
lervag: This will override other repeatable commands. How about instead passing `'repeat' : 0` to the `toggle_modifier` function?

kiryph:
@andymass 

> If I understand correctly, bootleq/vim-cycle cannot handle the following cases:
> 1. `( )` becomes `\bigl( \bigr)` and
> 2. `\cmd{foo}` becomes `\cmd[\big]{foo}` (uses mathtools, see below)

First of all bootleq/vim-cycle contains no rules for tex out of the box. All latex specfic functionality I was referring to is defined in the snippet I have given above which every user can change to his liking.

1. So regarding the first item:
I have defined following rule: if the cursor is on a single `(`, I want to cycle through delimiter types. However, this can be changed by removing following line
```
        \   [['(:)', '[:]', '\{:\}', '\lvert :\rvert', '\lVert :\rVert', '\lfloor :\rfloor', '\lceil :\rceil', '\langle :\rangle'], 'sub_pairs', 'hard_case', 'match_case'],
```
and modify other lines in following style
```
        \   [['(:)', '\bigl(:\bigr)', '\Bigl(:\Bigr)', '\biggl(:\biggr)', '\Biggl(:\Biggr)'], 'sub_pairs', 'hard_case', 'match_case'],
```
Now cycle.vim does what you want. However, you lose the ability to cycle through delimiter types. For changing delimiter types you could use again sandwich.vim, e.g. `cs([`. (Maybe I change this myself).

2. I suspect that this is not possible and on first sight does not look like a feature for cycle.vim or sandwich.vim. BTW shouldn't your question asking about following change sequence `(a+b)` to `\parens{a+b}` to `\parens[\big]{a+b}`? I think you could try following `['(:)', '\parens{:}', '\parens[\big{:}']`, but I am not sure how it works since the closing delimiter is always the same. Maybe it works as long as your always on the opening delimiter.

---

Well, mathmode.pdf is regarded as obsolete. So everything what is said within this document should be taken with care. However, it is a very comprehensive compilation of math constructs in latex. I am still looking for a non-obsolete alternative. I do not want to search tex.stackexchange for every math construct I want to use.

Thanks for your link. Indeed, I wasn't aware of the problem of 'incorrect' spacing for situations with  unary sign such as `\big(-a+b\big)`.

I have not used mathtools yet. I will give it a try to see if I like the modified spacing.

However, I have enabled concealment in vim. There is large number of concealment rules defined in $VIMRUNTIME/syntax/tex.vim which I will lose and this will in my opinion reduce readability of equations in the tex source code.

autoload/vimtex/delim.vim
andymass: Without `<c-u>` the command gets run multiple times.  The alternative is to make `vimtex#delim#toggle_modifier` a `range` function, but `<c-u>` is simpler.

autoload/vimtex/delim.vim
andymass: Yes, `l:count` is more logical.

autoload/vimtex/delim.vim
andymass: Yes, I prefer visual mappings to exit visual mode, like the standard vim mappings do.  If I get around to supporting visual repeat, this would be necessary because `.` in visual mode does nothing.

autoload/vimtex/delim.vim
lervag: Ah, yes, you are right. Good catch.

autoload/vimtex/delim.vim
lervag: Ah, I see. Then I agree this is useful.

andymass:
@lervag  I have started to document this feature, but ran into the problem that 
`tsD              |<plug>(vimtex-delim-toggle-modifier-reverse)|  `nx``
is too long for the columns in the list of mappings in `*vimtex-default-mappings*`.  I can't decide whether it would be better to a) shorten the map name b) shift all of the columns right or c) allow one line to go past the existing columns.

lervag:
I would shift the third column so that the minimum (visual) space between columns 2 and 3 are 2. That is, I would do:

```
   tsD              |<plug>(vimtex-delim-toggle-modifier-reverse)|  `nx`
```

Obv., this would mean changing the entire table, but that's no problem to me. Note, you should lenghten the lines above and below so that they are at least 1 longer (visually).

andymass:
Okay I think I have it now.  Let me know if you have any other concerns.

lervag:
Thanks! I've rebased on top of the master branch and then merged. Great work!

PR #936, opened 2017-08-07
Add note about vim-sandwich by machakann

kiryph:
to the documentation under the section
'Features provided by other plugins'

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -180,27 +180,58 @@ provided by vimtex, but that are instead provided by other plugins.
     `UltiSnips`. See |vimtex-neosnippet| and |vimtex-UltiSnips| for more info.
 
   Surround text with command, delimiter or environment~
-    vimtex itself provides only mappings to change, delete and toggle commands
-    delimiters and environments. If one wish to add these conveniently, there
-    is the plugin `sandwich` by Machakann [0] which provides this functionality.
-    Default mappings are `sa{motion/textobject}` or `{visual mode}sa` followed by
-       `c` to surround with a command
-       `e` to surround with an environment
-       left delimiter, e.g. `m(` to surround with `\left(` and `\right)`
-    For a list of delimiters see |sandwich-filetype-recipes|.
-    For consistency one may use as well the mappings for replacing `sr` and
-    deleting `sd` provided by sandwich instead of those from vimtex.
-    Note, you can repeat them with the dot command |.|.
-    One may dislike that the builtin mapping |s| is overwritten. However,
-    alternative mappings are available using as prefix `ys` instead of `sa`,
-    `cs` instead of `sr` and `ds` instead of `sd`. See |sandwich-miscellaneous|
-    how to enable them. Note, these are those of the plugin `surround` by Tim Pope
-    which, however, has only limited support for LaTeX.
+    |vimtex| itself provides only mappings to change, delete and toggle commands
+    delimiters and environments. If one wish to add these conveniently,
+    one can install one of the plugins
+        1. `sandwich` by Machakann [0] or
+        2. `surround` by Tim Pope [1]
+    which can assist by this.
     
-    `sandwich` offers many more features and consists of three vim help files
-    |sandwich.txt|, |operator-sandwich.txt| and |textobj-sandwich.txt|.
-    
+    By default both plugins provide support for single character surroundings
+    such as `'', "", (), [], {}, <>`.  However, `sandwich` supports already
+    additionally many LaTeX-specific surroundings such as quotations ``text''
+    and math delimiter `$\left(a+b\right)$` which can be further customized.
+    The default mappings of `sandwich` are `sa{motion/textobject}` or `{visual
+    mode}sa` followed by
+       `c` to surround with a command,
+       `e` to surround with an environment,
+       left delimiter, e.g. `m(` to surround with `\left(` and `\right)`.
+    For a list of delimiters see |sandwich-filetype-recipes|. For consistency
+    one may use as well the mappings for replacing `sr` and deleting `sd`
+    provided by sandwich instead of those from |vimtex|. All the mappings can
+    be repeated with the dot command |.|. There is even support for
+    `visualrepeat.vim` [2]. One may dislike that the builtin mapping |s| is
+    overwritten. However, alternative mappings are available using as prefix
+    `ys` instead of `sa`, `cs` instead of `sr` and `ds` instead of `sd`. See
+    |sandwich-miscellaneous| how to enable them. Note, these are those of the
+    plugin `surround` which can ease the transition from the plugin `surround`
+    to `sandwich`. `sandwich` offers many more features which are described in
+    three vim help files |sandwich.txt|, |operator-sandwich.txt| and
+    |textobj-sandwich.txt|.
+
+    `surround` by Tim Pope can also be customized to assist for LaTeX
+    surroundings. Support for LaTeX environments is already available via `l`
+    or `\`. To add support for LaTeX commands and quotations add these
+    lines to `~/.vim/after/ftplugin/tex.vim`
+    >
+        let b:surround_{char2nr('c')} = "\\\1command\1{\r}"
+        let b:surround_{char2nr('q')} = "`\r'"
+        let b:surround_{char2nr('Q')} = "``\r''"
+    <
+    However, note following about `surround`
+    - custom surroundings cannot be changed with `cs` or deleted with `ds`;
+      for instance, changing ``text'' to `text' with `csQq` or deleting with
+      `dsQ` does not work;
+    - the specifier for custom surroundings has to be a single letter; `m(` is
+      not allowed;
+    - current builtin support for LaTeX environments is an undocumented and
+      deprecated feature [3];
+    - it has no support for `visualrepeat.vim` [2].
+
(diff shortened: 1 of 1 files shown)


clason:
I'm a bit puzzled by the comment that `vim-surround` has only limited support for LaTeX: I've been using it for years now and never missed anything although I briefly tried `vim-sandwich`; I went back to `vim-surround` because I preferred the out-of-the-box mappings (and was too lazy to fully configure `vim-sandwich`), in particular the `ciB` et al. for changing inside `{}` pairs (which are inconvenient to enter with my keyboard layout).

If `vim-sandwich` is really much better for LaTeX specifically, I'd appreciate a few words (in the doc) on the concrete advantages; otherwise maybe give both equal billing?

kiryph:
Thanks for your response. I am more than happy to improve my suggestion for a description of sandwich.vim. Actually, it wasn't my intentation to bash vim-surround. It has the larger user base and is more tested than sandwich.vim. If a vim newbie asks me about essential vim plugins, I usually say three plugins: commentary.vim, surround.vim and unimpaired.vim.

The note in the docs is about surrounding with latex commands, environments and delimiters.
Surrounding simply with '', "", (), [], {} is not under discussion; these are useful in any programming, scripting, markup language and if I would only need those, I would have sticked to surround.vim

BTW, `ciB` works in vanilla vim, no surround.vim or sandwich.vim necessary. And it would work even when you have sandwich.vim installed. You give surround.vim credit for something which it actually does not do and it is not overwritten by sandwich.vim

See following list of missing features of surround.vim

1. **No support for latex commands out of the box**
I know you can define something with `let g:surround_{char2nr('c')} = "\\\1command\1{\r}"`. However, this has limitations see next item.

2. **let g:surround_{char2nr('c')} = "\\\1command\1{\r}" are only one-way**
Read following issue https://github.com/tpope/vim-surround/issues/141
Quote tpope (3years ago):
    > It's a hard problem to solve generally, and the chances of me doing it in the near future are essentially zero.

3. **Support for latex environments is left as an undocumented feature**
Read following commit message by Tim Pope: https://github.com/tpope/vim-surround/commit/ac67c1a7387a9e1c082fa98e7526b7b633f4beec
In short: he removed the documentation and left the functionality as undocumented feature within surround.vim. After 4 1/2 years I doubt that there will be ever a 'fully functional customization system'. We can be happy that he didn't remove it. You can always maintain your own version, but the doc should point people to a plugin with the feature, why you recommend it, documented.

4. **Limited customization, e.g. no support for custom surroundings denoted by two letters**
LaTeX has often more complex surroundings such as ` ``text''`, `\left(a+b\right)`, `\langle a+b \rangle` which require customization of surround.vim. This would be fine, but the customizability of surround.vim is limited.
Read also this issue https://github.com/tpope/vim-surround/issues/201 where someone suggests to use `q` and `Q`. I guess you can easily imagine that you will shortly run out of letters and symbols to define all the things you like to have. And as already mentioned you can only add them deletion or changing is not possible.
For latex specific customization of sandwich.vim see https://github.com/machakann/vim-sandwich/blob/master/after/ftplugin/initex/sandwich.vim. I know this can be little bit overwhelmig/intimidating/user-unfriendly, but in the long run you simply can do more with sandwich.vim.

5. **In sandwich.vim many latex specific delimiters, quotations are already supported out of the box**
Sandwich.vim offers e.g. `m(` for `\left(... \right)` `l"` for ` ``...''`. 
This is useful for newbies who do not want to configure a plugin before they can use it.

6. **No support for visualrepeat.vim in contrast to sandwich.vim**
I admit this is not latex specfic but certainly useful.

Let me see if I find a few words why it is in particular for latex user better to use sandwich.vim than surround.vim. I want to avoid blowing the paragraph up. But I see your point that it should be clear why one would use sandwich.vim instead of surround.vim. This is currently not obvious.

clason:
Thanks, I was (and am) honestly curious. (I also was not aware that `ciB` is vanilla vim.) 

Maybe three or four concrete(!) examples of useful commands that vim-sandwich provides that do not work under vanilla vim(-tex) (or surround) would suffice?

@lervag Maybe a short section "Plugins that work well with/extend vimtex" would also be a useful addition to the README.md? (IMO, the "we're not a kitchen sink" attitude is one of the selling points of vimtex.)

kiryph:
@clason What do you think about following:

```diff
+    One may dislike that the builtin mapping |s| is overwritten. However,
+    alternative mappings are available using as prefix `ys` instead of `sa`,
+    `cs` instead of `sr` and `ds` instead of `sd`. See |sandwich-miscellaneous|
+    how to enable them. Note, these are those of the plugin `surround` by Tim Pope
+    which is a possible alternative to sandwich.vim.
+    Note, sandwich.vim provides following advantages over surround.vim:
+    1. Better support for customization: (a) For instance, custom surroundings can
+        not only be added, but also deleted and changed in contrast to surround.vim
+        customizations with `let g:surround_{char2nr(' ')}` which only can be added.
+        Consider e.g. the quotation  ` ``text''` which one would like to easily
+        change to a different type of quotation.
+        (b) Also the custom surrounding specifier can be more than one letter which
+        is useful for the zoo of latex surroundings which should not overwrite the
+        standard set of surroundings '',"",(),[],{},<>,||.
+    2. A comprehensive LaTeX customization is already available with many math
+        delimiters such as `m(` and quotations such as `l"`.
+    3. Support for visualrepeat.vim
```

lervag:
@clason First, I'm happy you agree on the "we're not a kitchen sink" attitude. Although, I have to admit the plugin is larger than I ever intended, I still feel I (barely) manage to keep that attitude. I think you are right that it would be convenient with a section that describes plugins that work well with vimtex also in the README file. In the docs I've called that section "Features provided by other plugins" (in the spirit of the forementioned attitude).

@kiryph I'll wait for your next version before reading in full. I skimmed it, and I think it is close to acceptable. However, I will probably suggest (or edit myself) improvements to the wording and text flow. IMHO, the section does not really need to discuss at length why it is "better" than `surround.vim`, because `surround.vim` does not provide the relevant features (in an appropriate manner, at least).

One thing, though: "Note, these are those of the plugin `surround` by Tim Pope which is a possible alternative to sandwich.vim." I don't understand the sentence.

lervag:
I've updated the README as suggested by @clason. Perhaps you could also add relevant links in the PR, @kiryph?

kiryph:
@lervag Before I push a new version, I just want to clear things up.

1. Clason wanted to know why sandwich.vim has advantages over surround.vim and I guess many more user who are familiar with surround.vim could be interested in why they could consider changing plugins.
IMHO the original version does not provide this except for my comment about limited support for latex.
2. >Note, these are those of the plugin surround by Tim Pope which is a possible alternative to sandwich.vim.

   This comment also targets surround.vim user that they do not have to relearn all mappings. I think this sentence can be deleted. This is documented in sandwich.vim anyway. "we're not a kitchen sink" applies also to documentation.

So I guess I will not add the surround.vim vs. sandwich.vim discussion. I think also the notes about the alternative mappings could be removed. The users should simply read the docs of sandwich.vim. I think extensive documentation also scares people away.

clason:
@lervag As long as the first reaction is "Does this really have to be in vimtex, or is there another plugin which already does this well", I'd say you're on message. A section in the README would make this even more obvious :) (No need for long descriptions; a bullet point list of links with "see vimtex.txt for more details and configuration hints" would suffice.)

@kiryph I agree, keeping it short and to the point is a good idea. Maybe something like 

> Deleting and changing surroundings is provided by `vim-surround`. For example, you can
> 1. ...
> 2. ...
> 3. ...
>
> More customisation is possible with `vim-sandwich`, which in addition to the above allows to, e.g.,
> 1. ...
> 2. ...
> 3. ...

The exact mappings are not so relevant here, I think.

lervag:
I agree that a short list similar to @clason's suggestion looks good. However, it would need to be clear that `vim-surround` does not provide alternatives to the `csd`-family of maps from vimtex. Thus one could say that `vim-surround + vimtex ~= vim-sandwich` (when it comes to the surround type of mappings). In other words, it would be useful to mention that `vim-surround` provides useful features that work well with vimtex, and that `vim-sandwich` provides a generalisation of `vim-surround` that is more customizable and that also supports more complex types of surroundings. The latter means that `vim-sandwich` may be used instead of the vimtex mappings, etc...

clason:
I think that would be very helpful -- I see that I am quite confused by what is actually provided by vimtex, vim-surround and vanilla vim... (Which is fine from a user point of view, because it means "it just works" :))

kiryph:
I changed my mind and added surround.vim as another options for surroundings and from my perspective relevant information for latex writer.

If this text is too long for this section, please take what you like it and create a version you think is suitable. I just do not want to spend any time on this anymore. Sorry.

lervag:
I took the liberty of taking parts of what you wrote and combining it in a different way. I think my version works better, and I feel we don't need all the details you added. However, if you disagree, I would be willing to reconsider adding an extra note on how to configure `vim-surround` for LaTeX; I removed this because I don't really find it that relevant, since this is already provided by vimtex in a (IMHO) superior way.

I really appreciate your suggestion and work here. Thanks! I am probably going to test the LaTeX support from vim-sandwich, which I haven't really done so far.

clason:
@kiryph I understand, and appreciate your explanations! I learned quite a bit and will check out `vim-sandwich` again.

Issue #941, opened 2017-08-10
VimtexEventQuit doesn't really work


andymass:
### Explain the issue

This is really simple to reproduce.  1. Create an autocommand `au User VimtexEventQuit echo 'foo' | sleep 2` 2. Open a Tex file. 3. Open a not tex file in the same window.  4. Quit vim. 

It is clear the command is not being called.  

Another bug is if you open a Tex file multiple times the autocommand is added repeatedly, so if you close the tex file now, it is called multiple times.  

It is difficult to use this feature to do cleanup as the docs state.  

### Minimal working example

Please provide a minimal working LaTeX example, e.g.

```tex
\documentclass{minimal}
\begin{document}

Hello World!

\end{document}
```

### Minimal vimrc file

Please provide a minimal vimrc file that reproduces the issue. The following
should often suffice:

```vim
set nocompatible

" Load Vimtex
let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'

" Load other plugins, if necessary
" let &rtp = '~/path/to/other/plugin,' . &rtp

filetype plugin indent on
syntax enable

" Vimtex options go here
```

lervag:
I think I've fixed this now. It is a little bit tricky, actually, to properly handle when Vim quits or deletes buffers. But I think the latest update should be correct in all cases (famous last words, or something). Let me know if you should find another edge case that is wrong.

andymass:
Thanks, will test and get back.  You're a machine.  I'm trying really hard to not let the open issue count get too low :)

lervag:
Yeah, I've noticed. Luckily, I've had some spare time lately to keep on top of the issue count. :)

andymass:
Unfortunately, there are still a few problems.  I'm documenting them here, but I have not fully investigated the cause(s) yet.  I agree it's a tricky problem (with relatively minor impact), so feel free to ignore.

### 1. `VimtexEventQuit` is sometimes fired twice (sometimes even more).
MWE: directory structured as follows:
```
- minivimrc
- minimal.tex
- included.tex
- minimal.tex.latexmain
```

minivimrc:
```vim
set nocompatible
let &rtp = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'
filetype plugin indent on
syntax enable

let g:tex_flavor = 'latex'
augroup Testing
  autocmd!
[...]

PR #951, opened 2017-09-03
Improve documentation for VimtexEventQuit, add additional example

andymass:
Now that the `VimtexEventQuit` feature seems reliable (cf. #941), I think it deserves some better documentation.  This PR aims to clear up potential confusion about when the event is triggered and adds another example which closes any open viewers when the project is quitting.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1841,7 +1841,7 @@ Commands~
                                              *VimtexCompile*
                                              *<plug>(vimtex-compile)*
 :VimtexCompile            If the compiler supports and is set to run in
-                          continuouos mode, then this command works as
+                          continuous mode, then this command works as
                           a compiler toggle. If not, this command will run
                           a single shot compilation.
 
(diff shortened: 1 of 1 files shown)


doc/vimtex.txt
```diff
@@ -2212,12 +2218,29 @@ through |autocmd|s, in particular the |User| autocmd.
...
+  function! CloseViewers()
+    if executable('xdotool') && exists('b:vimtex')
+        \ && exists('b:vimtex.viewer') && b:vimtex.viewer.xwin_id > 0
+      call system('xdotool windowclose '. b:vimtex.viewer.xwin_id)
+    endif
+  endfunction
+
+  augroup vimtex_event_2
+    au!
+    au User VimtexEventQuit     call CloseViewers()
```
lervag: Don't need many spaces between event name and `call CloseViewers` (no need to align here).

doc/vimtex.txt
```diff
@@ -2212,12 +2218,29 @@ through |autocmd|s, in particular the |User| autocmd.
...
+  augroup END
+
+Note: Commands such as |VimtexClean| cannot be used in these autocommands because
```
lervag: This note should be moved to the definition of `VimtexEventQuit`, because it only applies to this specific event.

andymass:
Thanks, updated.

PR #959, opened 2017-09-28
Minor formatting and readability edits

thomasteoh:


doc/vimtex.txt
```diff
@@ -68,7 +68,7 @@ CONTENTS                                                      *vimtex-contents*
     Table of contents                           |vimtex-toc|
     Table of labels                             |vimtex-labels|
     Unite sources                               |vimtex-unite|
-    Include expression (|gf| command)             |vimtex-includeexpr|
+    Include expression (|gf| command)           |vimtex-includeexpr|
```
lervag: Help files are by default showed with the conceal attribute enabled. With conceal, the `|`s are not displayed, and in this case, the current version looks right and your version looks wrong. I don't know any way to make it look right in both cases.

doc/vimtex.txt
```diff
@@ -551,7 +551,7 @@ This feature is explained in more detail later, see |vimtex-imaps|.
    []               |<plug>(vimtex-[])|                             `nxo`
    [[               |<plug>(vimtex-[[)|                             `nxo`
    K                |<plug>(vimtex-doc-package)|                    `n`
-  ---------------------------------------------------------------------~
+  ----------------------------------------------------------------------~
```
lervag: The changes to this table seems to again assume that conceal is turned off. However, the conceal feature is enabled by default, and so I think the formatting should assume that the `|`s and `` ` ``s are concealed here.

doc/vimtex.txt
```diff
@@ -2508,11 +2508,11 @@ change the fold filler, use e.g.: >
...
-  |g:vimtex_fold_enabled|
-  |g:vimtex_fold_manual|
-  |g:vimtex_fold_levelmarker|
-  |g:vimtex_fold_types|
-  |g:vimtex_fold_types_defaults|
+    |g:vimtex_fold_enabled|
+    |g:vimtex_fold_manual|
+    |g:vimtex_fold_levelmarker|
+    |g:vimtex_fold_types|
+    |g:vimtex_fold_types_defaults|
```
lervag: I guess this is to make the similar lists consistent, however, in this case you missed the ones around line 2265.

Issue #949, opened 2017-08-25
Bad indentation in item list


arnegoeteyn:
### Explain the issue

Whenever I am writing a list of \item it gets indented after every new \item. What I would like is
```
\begin{itemize}
    \item
        \item
            \item
\end{itemize}

```
But what I would like to achieve is 
```
\begin{itemize}
    \item
    \item
    \item
\end{itemize}
```

When I dont load the vimtex plugin I get the intended behaviour. The item gets indented at first but as soon as I press space the indentation fixes itself.

### Minimal working example

Please provide a minimal working LaTeX example, e.g.

```tex
\documentclass[12pt]{article}
\begin{document}
\maketitle

\section{Materiaal}
\label{sec:materiaal}
\begin{itemize}
	\item
		\item 
			\item
\end{itemize}



\end{document}

```

### Minimal vimrc file

```vim
call plug#begin('~/.local/share/nvim/plugged')
Plug 'lervag/vimtex'
Plug 'matze/vim-tex-fold'
call plug#end()

```

lervag:
Given your example, what happens if you do `=ie` inside the `itemize` environment? Or `gg=G` from anywhere? Does it yield correct indentation?

This works well for me, also with your supplied minimal example. Could you please consider to use a more minimal vimrc file? E.g. something like this:

```vim
set nocompatible
let &rtp = '~/.vim/bundle/vimtex,' . &rtp
let &rtp = '~/.vim/bundle/vim-tex-fold,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'
filetype plugin indent on
syntax enable
```

Does the problem exist also if you do not load `vim-tex-fold`? (If so, _why_ do you include it in the minimal vimrc file?!)

I assume that you've understood that you must start vim like this to use the minimal vimrc file: `vim -u minivimrc minimal.tex` (or `nvim` or `gvim`, whatever is applicable).

arnegoeteyn:
`=ie `and` gg=G` yield correct indentations. I forgot to remove the` vim-tex-fold` from my minimal vimrc. But removing it doesnt yield correct indentation.

However using your minimal vimrc, I get the correct indentation. Can this be an issue with how the plugin is loaded?

lervag:
I suspect that the problem is your configuration, which is why I ask you to
create a minimal vimrc file where the problem still occurs. You should keep
the settings from your vimrc file that are relevant.

If you do not understand what I mean, let me know and I will try to explain
in more detail.

[...]

PR #961, opened 2017-09-29
Improve indent alignment on ampersands

andymass:
Cf. [comment](https://github.com/lervag/vimtex/issues/949#issuecomment-331747897) on issue #949.

The `indentexpr` does not correctly compute the ampersand position when `set noexpandtab`, because it uses `stridx()`.  Along the way to fixing this, I came up with another annoyance in the following example:

```tex
\documentclass{minimal}
\begin{document}
\begin{align}
      x&= (x+5)(x+2)
          \\ &= x^2 + 5x + 2x + 10      % will not align due to '\\ '
    \\&= x            % vimtex thinks this is an escaped &
\end{align}
\end{document}
```

I generally put `\\` before the `&` instead of at the end of the previous line because it allows one to delete equations more easily.

This PR aims to:
        - Compute offsets correctly when `noexpandtab` is set
        - Allow `\\&` to be aligned
        - Better checking for escaped & with backslashes

Future work: it would be nice if ampersands would align to multiple lines above, not just on the immediately previous line:

```tex
\begin{align}
      x&= (x+5)(x+2)
          + x^2 + 5x + 2x + 10
    \\&= x            % will not align
\end{align}
```

### DIFF
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -43,9 +43,13 @@ function! VimtexIndent(lnum) " {{{1
 
   " Align on ampersands
   if get(g:, 'vimtex_indent_on_ampersands', 1)
-        \ && l:line =~# '^\s*&'
-        \ && l:prev_line =~# '\\\@<!&.*'
-    return indent(a:lnum) + match(l:prev_line, '\\\@<!&') - stridx(l:line, '&')
+        \ && l:line =~# s:amper_align
+        \ && l:prev_line =~# s:ampersand
+    let l:indent_prev = strdisplaywidth(strpart(l:prev_line, 0,
+      \ match(l:prev_line, s:ampersand)))
+    let l:indent_cur = strdisplaywidth(strpart(l:line, 0,
+      \ match(l:line, s:ampersand)))
+    return max([indent(a:lnum) - l:indent_cur + l:indent_prev, 0])
   endif
 
   " Use previous indentation for comments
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! This seems useful, and I will be glad to accept this update. I will review it as soon as I have the time!

indent/tex.vim
```diff
@@ -43,9 +43,13 @@ function! VimtexIndent(lnum) " {{{1
 
   " Align on ampersands
   if get(g:, 'vimtex_indent_on_ampersands', 1)
-        \ && l:line =~# '^\s*&'
-        \ && l:prev_line =~# '\\\@<!&.*'
-    return indent(a:lnum) + match(l:prev_line, '\\\@<!&') - stridx(l:line, '&')
+        \ && l:line =~# '^\W*' . g:vimtex#re#not_bslash . '\m&'
```
lervag: Would it not be better with `\&` instead of `\m&`?

Also, I'm always wondering about stuff like this: does the string concats affect runtime here? Would it be better to define the regex once in a local/script local variable? This might be a good idea also because the same regex is used on line 84, and almost the same on lines 47, 49 and 51.

indent/tex.vim
```diff
@@ -43,9 +43,13 @@ function! VimtexIndent(lnum) " {{{1
...
+        \ && l:prev_line =~# g:vimtex#re#not_bslash . '\m&.*'
+    let l:a = strdisplaywidth(strpart(l:prev_line, 0,
+      \ match(l:prev_line, g:vimtex#re#not_bslash . '\m&')))
+    let l:b = strdisplaywidth(strpart(l:line, 0,
+      \ match(l:line, g:vimtex#re#not_bslash . '\m&')))
+    return max([indent(a:lnum) - l:b + l:a, 0])
```
lervag: How about `l:indent_prev` instead of `l:a` and `l:indent_cur` instead of `l:b`? Or something like that?

lervag:
Are you still working on this? If you want, I could also look into this myself.

andymass:
Should be complete now- sorry for the delay in getting back.  My most recent commit attempts to address your comments, thanks!  Feel free to amend.

I did spend a little bit of time looking into the "future work," but it turned out to be too complex to implement than I have time for at the moment.

lervag:
Thanks, this looks good! I'll open an issue for your future work suggestion. I might have time to look at it. There are no other open issues, and a man needs something to do on his spare time... :)

Issue #964, opened 2017-10-02
Better error handling for latexmk


s4shmz:
First of all, thanks for developing vimtex. The plugin helps my daily works very well.

This is my small ftplugin that I use with vimtex. The function `FilterLatexmkMessages` picks messages of each last-run command called by latexmk.

```vim
setlocal errorformat-=%-G%.%#

augroup FilterLatexmkMessages
  autocmd!
  autocmd QuickfixCmdPost <buffer> call s:FilterLatexmkMessages()
augroup END

function! s:FilterLatexmkMessages()
  let qflist = getqflist()

  "
  " Add log messages to a dictionary. Keys of the dictionary are command rules
  " called by latexmk. If same rule is processed more than once, only the
  " last-run message is stored in the dictionary
  "
  let match_rule="Run number . of rule "
  let log_dict={}
  let log_key=''

  for i in qflist
    if match(i.text, match_rule) != -1
      let log_key = i.text[match(i.text, match_rule)+len(match_rule):]
      let log_dict[log_key] = []
    endif
    if log_key != ''
      call add(log_dict[log_key], i)
    endif
  endfor

  "
  " Reconstruct qflist from the dictionary
  "
  " By this if statement, this function is compatible for single command(latex,
  " pdflatex , bibtex or something else).
  if log_key != ''
    let qflist=[]
    for key in keys(log_dict)
      for i in log_dict[key]
        if ! i.valid
          continue
        endif
        call add(qflist, i)
      endfor
    endfor
    call setqflist(qflist)
  endif
endfunction

" Errorformat for bibtex
setlocal errorformat+=%.%#---line\ %l\ of\ file\ %f
setlocal errorformat+=Sorry---you've\ exceeded\ BibTeX's\ %.%#
setlocal errorformat+=%.%#---this\ can't\ happen
setlocal errorformat+=I\ found\ %.%#---while\ reading\ file\ %f
setlocal errorformat+=refers\ to\ entry\ "%.%#
setlocal errorformat+=%tarning--I\ didn't\ find\ a\ database\ entry\ for\ %.%#
setlocal errorformat+=%tarning--%m
setlocal errorformat+=Aborted\ at\ line\ %l\ of\ file\ %f
setlocal errorformat+=%.%#\ WARN\ -\ %.%#
setlocal errorformat+=%.%#\ ERROR\ -\ %.%#
```

[...]

PR #970, opened 2017-10-10
update errorformat for blg files

s4shmz:
#964 

- `Repeated entry` should be an error.
- A new error is added. (no bibliographystyle)

```latex
\documentclass{article}

\begin{document}

\cite{bib}

%\bibliographystyle{plain}
\bibliography{ref}
\end{document}
```

```
@article{bib,
	author={a},
	title={t},
	journal={j},
	year={2017},
}
```


lervag:
Thanks! I've improved this with slightly better error message. I also try to put the location near where the `\bibliographystyle` command _should_ be (although, of course, this may be "wrong"; still, I think more often than not it might be helpful, at least more helpful than referring to the aux file).

PR #979, opened 2017-10-17
add todonotes package support to match \todo in toc

dopefishh:
The [todonotes](https://ctan.org/pkg/todonotes) package provides handy commands to create todo's that also appear in the document as balloons in the page margin. This PR adds a matcher for the TOC that matches the `\todo{}` commands.


lervag:
This is smart, thanks!

PR #997, opened 2017-11-11
Fix #982: use pure Apple Script to (re)open skim

### LINKED ISSUE #982: Invoking Skim / displayline after compilation fails.
Hello, 

as mentioned in #204, I get an syntax error after `let g:vimtex_view_method = 'skim'` in the `.vimrc` or `ftplugin/tex.vim` and running `\ll`. This is strange, as invoking Skim manually by calling `\lv` works fine.

To make this reproducable, I created a minimal example:

```
\documentclass{article}
\usepackage[backend=bibtex]{biblatex}
\begin{document}
Hello World!
\end{document}

```

Running `\ll` produces:
```
Latexmk: This is Latexmk, John Collins, 19 Jan. 2017, version: 4.52c.
Viewing pdf
Rule 'pdflatex': Rules & subrules not known to be previously run:
   pdflatex
      'pdflatex'
------------
------------
------------
Running 'xelatex  -file-line-error -synctex=1 -interaction=nonstopmode -recorder  "test.tex"'
------------

. . .
. . . 

Transcript written on test.log.
Latexmk: Log file says output to 'test.pdf'
=== TeX engine is 'XeTeX'
Latexmk: Found bibliography file(s) [test-blx.bib]
Latexmk: I have not found a previewer that is already running. 
   So I will start it for 'test.pdf'
------------
------------
Running 'start /Applications/Skim.app/Contents/SharedSupport/displayline  "test.pdf"'
------------
Latexmk: All targets (test.pdf) are up-to-date
For rule 'view', running '&if_source(  )' ...
------------
Running '/Applications/VimR.app/Contents/Frameworks/SwiftNeoVim.framework/Versions/A/PlugIns/NeoVimServer --servername /var/folders/76/w433cm8d655cznw1qmd866gc0000gn/T/nvimhw7Q3i/0 --remote-expr "vimtex#compiler#callback(1)"'
------------
17:22: 
syntax error: A unknown token can’t go after this identifier. (-2740)
```

And indeed, running the shown command (`/Applications/Skim.app/Contents/SharedSupport/displayline  "test.pdf"`) manually, it returns the same error. This is, because displayline also needs a line parameter:

```
$ /Applications/Skim.app/Contents/SharedSupport/displayline "test.pdf"
17:22: syntax error: A unknown token can’t go after this identifier. (-2740)
$ /Applications/Skim.app/Contents/SharedSupport/displayline                                                                                                      
Usage: displayline [-r] [-b] [-g] LINE PDFFILE [TEXSOURCEFILE]
Options:
-r, -revert      Revert the file from disk if it was open
-b, -readingbar  Indicate the line using the reading bar
-g, -background  Do not bring Skim to the foreground
$ /Applications/Skim.app/Contents/SharedSupport/displayline 10 "test.pdf"
$ 
```

So, it appears that after a regular compilation Skim should be invoked via `open`, rather than trying to highlighting a specific line via `displayline`, which fails?

clason:
As discussed in #982, this removes the dependency on the `displayline` bash script distributed with Skim by directly calling the Apple Script command from `displayline` in `skim.view()` and `skim.compiler_callback()`, hardcoding the appropriate options:

* `skim.view()` opens the PDF,  jumps to the current source line, and highlights it;
* `skim.compiler_callback()` just opens the PDF without changing the view.

Both functions check first whether Skim is running and if so, revert the document instead. (This replaces the `pgrep` call in the callback. The code should now also behave properly if multiple Skim windows are open; the previous script always reloaded the currently active window, independent of the contents.)

Some refactoring might be possible to move part of the `osascript` call to a common variable. I have also not touched the error check in line 10, which still depends on `displayline`.


clason:
@tnull Could you test if this works for you?

lervag:
Nice, good work! I think I like this approach better than what we did before.

I think the requirement check now instead can check for `executable('osascript`)` instead of `displayline`.

I would be happy if more people could test this and confirm that it works as expected, since I am not able to test anything myself.

Note that in the version you pushed you've left an `echom`.

clason:
Ah, `osascript` is part of the operating system, and its presence will not tell you if Skim is installed. I don't know if there's a convenient way of testing whether `open -a Skim` will succeed (note that Skim might be installed in `~/Applications/` -- not that I know of a reason why on would do that), so the `displayline` test might still be the simplest one. (As far as I know, this is not an optional part of installing Skim.)


> Note that in the version you pushed you've left an echom.

d'oh!

lervag:
> Ah, osascript is part of the operating system, and its presence will not tell you if Skim is installed. I don't know if there's a convenient way of testing whether open -a Skim will succeed (note that Skim might be installed in ~/Applications/ -- not that I know of a reason why on would do that), so the displayline test might still be the simplest one. (As far as I know, this is not an optional part of installing Skim.)

I see. Then I agree that the `displayline` check might still be relevant, unless someone proposes a better way of checking whether Skim is installed.

It seems we only need someone else to test this before merging. Perhaps @nilsbecker is still around as well?

nilsbecker:
this SO answer has applescript to test for presence of an application: https://stackoverflow.com/questions/6682335/how-can-check-if-particular-application-software-is-installed-in-mac-os 
(it's embedded in a bash script but that would not be necessary)

to test the proposed fix, what do i have to do? can i simply edit a line in my Vundle bundle list in my .vimrc to pull the to-be-merged branch in?

clason:
@nilsbecker Thank you for the link; I'll try it when I'm back on a Mac.

To test, the simplest way is probably to manually edit the `skim.vim` file (or download and replace it: click on "Files changed" -> "View" -> "Raw").

lervag:
@nilsbecker I don't know about Vundle, sorry. But you could go the the vimtex repo in a terminal, then do something like the following:

```
# Go to vimtex dir
cd ~/.vim/bundle/vimtex

git checkout -b clason-skim-reload master
git pull git://github.com/clason/vimtex.git skim-reload
```

When you are finished with testing, you can do the following to clean up and restore:

```
git checkout master
git branch -D clason-skim-reload
```

clason:
I've implemented the approach @nilsbecker linked to.

I've tried taking the name from the `s:skim` dict instead of hardcoding it; this works for the `view` and `callback`, but not the `new` functions, so in the end I went back to the hardcoded name.

lervag:
I don't think it is that bad to use the hardcoded name, since there is no reason to believe it would ever change.

I assume since you've pushed, that you have tested the code yourself and can verify that it does what you expect it to do? A couple of thoughts and comments before merging:

- Did you test if the new "executable" check works, for instance by changing the name from "Skim" to "SkimWrong" (which is not executable)?
- I prefer that you remove the extra blank line between the comment and the conditional block, i.e. line 9 in your new code.

Else I would not mind merging this myself, but it is probably smart to wait some more in case someone else could also test this.

clason:
Re name: I agree, I just thought that was what the dict was for.

Re tests: I checked that it works both if Skim is running and not, for `view` as well as `callback`. (And yes, it prints `Skim not installed` if I change `"Skim"` to, e.g., `"Skin"`.) And yes, it'd be nice to have at least one independent verification.

Re code: I didn't pay attention to code style yet; feel free to modify as you wish. (I ticked the box to permit that.)

tnull:
@clason I tested this, and seems to work very well. I even have the impression that it shifts the viewport less than before?

clason:
@tnull Thank you for testing! I haven't noticed any shifting myself, but haven't tried it yet on all my macOS machines. (It might have something to do with the size of the viewport, in particular whether it fits a whole page or only part.)

nilsbecker:
i have also tested (thanks @lervag for the detailed instructions) and forward searching and pdf update on compilation seem to work fine. there is not yellow bar when the pdf is updated.

lervag:
Ok, the feedback seems to indicate that this works as expected and that it provides an improvement to the current master branch. Unless there are any objections or other comments, I'll merge this later today.

lervag:
Thanks, this has now been merged. Let me know if there should be any related issues!

lervag:
And thanks to @clason for the good work!

PR #1015, opened 2017-12-01
Fix and improve focusing vim on backwards search 

real-or-random:


lervag:
Thanks! I don't quite get this, though, and it would be nice with explanations. That is,

1. Since we are contacting a window for a given PID, why do we need to ignore invisible windows?
2. What is the benefit of calling xdotool in the background?

real-or-random:
I should have explained more. :)

1. For some reason, xdotool finds two windows for a single gvim pid in my environment. I don't know what this second window is but it is invisible. xdotool then uses the first window that it finds and then tries to activate the invisible window, which fails. This is what the change fixes. (An alternative is to activate all windows but I think the method in the commit is more conservative.)

2. I don't know if calling it in the background is useful. When playing around with xdotool, I noticed that the --sync argument is in the wrong place and ignored. I moved it to the right place but then did not see a reason why vim should wait for the window to be activated. (Maybe there is one?) So I went the opposite way, removed --sync and then consequently put the entire call to the background. I don't think the & makes a real difference.



Am 2. Dezember 2017 21:27:55 MEZ schrieb "Karl Yngve Lervåg" <notifications@github.com>:

lervag:
Great, thanks for the explanations! Seems good to me!

PR #1037, opened 2018-01-15
Completion: Autocompletion for acronyms defined through the glossaries package.

x49:
This change allows vimtex to parse completion candidates from acronym definitions like
`\newacronym{foo}{FOO}{foo is an acronym}`
as well as
`\newacronym[longplural={many FOOs}]{foo}{FOO}{foo is an acronym}`
in addition to normal glossary entries.

### DIFF
--- a/autoload/vimtex/complete.vim
+++ b/autoload/vimtex/complete.vim
@@ -621,6 +621,8 @@ endfunction
 
 let s:completer_gls = {
       \ 'patterns' : ['\v\\(gls|Gls|GLS)(pl)?\s*\{[^}]*$'],
+      \ 'keywords' : ['newglossaryentry', 'longnewglossaryentry', 'newacronym', 
+                    \ 'newabbreviation', 'glsxtrnewsymbol'],
       \ 'candidates' : [],
       \}
 
@@ -636,9 +638,9 @@ function! s:completer_gls.parse_glossaries() dict " {{{2
   for l:line in filter(vimtex#parser#tex(b:vimtex.tex, {
         \   'detailed' : 0,
         \   'input_re' : g:vimtex#re#tex_input . '|^\s*\\loadglsentries',
-        \ }), 'v:val =~# ''\\\(newglossaryentry\|newacronym\)''')
+        \ }), 'v:val =~# ''\\\(' . join(self.keywords, '\|') . '\)''')
     let l:entries = matchstr(l:line,
-        \ '\\\(newglossaryentry\|newacronym\)\s*\(\[.*\]\)\=\s*{\zs[^{}]*')
+        \ '\\\(' . join(self.keywords, '\|') . '\)\s*\(\[.*\]\)\=\s*{\zs[^{}]*')
     call add(self.candidates, {
           \ 'word' : l:entries,
           \ 'abbr' : l:entries,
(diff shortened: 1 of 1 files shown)


lervag:
This seems good, thanks! Just some quick questions: I assume `newacronym` is a part of the glossay package? Also, should one update the autocomplete regexes to support other commands that should initiaite completion?

x49:
Yes, exactly. newacronym is defined in the `glossary` package, I'm using it from the `glossaries` package that supercedes that (`glossary` is deprecated). The syntax is identical for both. Acronyms are also displayed with variants of `\gls{}`, so there's no need for a change there. Your comment also made me think to double-check that the `glossaries` package doesn't provide more commands than the two that are implemented now, but the documentation actually mentions these:

```
\newabbreviation[options]{label}{symbol}{description}
\glsxtrnewsymbol[options]{label}{symbol}
\longnewglossaryentry{label}{name={name}}{description}
```

In addition to the ones already covered:

```
\newacronym[options]{label}{symbol}{description}
\newglossaryentry{label}{name={name},description={description},<other options>}
```

Is covering them all with the same regex desirable? It would get pretty unwieldy. If not, how would you split it up? match all commands in the `filter()` call and then only match one command or a subgroup? If we did that, we could also think about setting the word in the candidates list to a more meaningful value, maybe `<name>` for glossary/longglossary, and `<symbol>` for the remaining commands?

lervag:
I would think it should be possible to match all of the commands with the same regexp without it being to clumsy. However, we might need to update the code to prevent the more complex regexp from making it unreadable. I think I would define the regexes as part of the `s:completer_gls` dict and use `join([...], '|')` to make it more readable. Something like that, anyway.

Feel free to suggest an update. Alternatively, provide a good example of all the required completion cases, then I can implement it based on the example. I would also want to update the test cases.

x49:
You're right, it's pretty readable like that. I'll leave the vader tests for you, if that's ok, as I'm not really familier with its use.

Apparently, the description is mandatory for glsxtrnewsymbol (Not really in the docs, but I get a latex error if I omit it). From my understanding that shouldn't be a problem in the test case, anyway? But probably better to leave it in as the new regexp should ignore anything in square brackets.

```
\newglossaryentry{foo}{test}
\newglossaryentry{bar}{test}
\longnewglossaryentry{baz}{name={test}}
\newacronym{qux}{test}
\newabbreviation{quux}{test}
\glsxtrnewsymbol[description={test}]{quuz}{test}
```

lervag:
Thanks! I've added the commands to a test file and made some minor improvements. I merged, although I think we can still discuss what the completion menu keys should be. Let me know what you think.

x49:
Tested it and works well for me! Thanks for your help with this. As for the menu entries, I think ideally it'd be something like
```
<acronym> [acr] <expanded acronym>
<term> [gls] <description of term>
```
I have to think a little bit of how best to capture that, due to the differing syntax for the different glossary/acronym/symbol types. But pragmatically, the way it is implemented now works well, and this would be nice but not totally necessary.

lervag:
No problem, thanks for suggesting the patch and feature improvement!

I agree, it seems useful to include the added information. However, to parse it would require more complex code, because we would need to allow multiple line commands. The current version is very simple because we assume that the relevant stuff is on a single line.

If you can create a very clear overview of the syntax for the different commands, in the sense that I had a working set of examples for testing, then I could probably look into it. If so, please open a new issue.

PR #1044, opened 2018-02-05
Complete environment names

andymass:
After typing `\begin{` or `\end{` we could suggest names of environments.  I was recently writing a document and was initially surprised environments were not suggested to me.

What do you think about this?  I know completion has likely already increased in scope beyond what you originally intended, but this seemed to me like low-hanging fruit since the command completion framework was already in place, and the cwl files already have environments listed.  I simply list the `\begin{name}` at the end of the file, I do not include the `\end{}`.

This PR also repairs the following defects:
    - command candidates become stale and wrong for other documents with different packages after first completion, due to caching
    - lcd is improperly used if there is currently no local directory (lcd causes a window to have a local directory when previously there was none)
    - the cwl conversion script produces non-deterministic ordering of commands due to `key=str.lower`
    - utf8 conversion error in cwl conversion script on latin1 encoded character ü
    - the completion data file class-scrartcl,scrreprt,scrbook is useless here, I created three separate files for the three classes and used an # include statement


lervag:
Thanks! I think this looks good. However, it is quite a lot of changes in one single commit. Would it be possible to split this into a couple of (logical) commits? That would also make it easier for me to review. For instance, updated complete files could all be added in a single commit, which would simplify things greatly (since there is no reason to review these directly). 

However, if you're short on time, I will be able to review it as it stands when I get the time.

In any case, again: Thanks! I very much appreciate your effort, and my first impression is that this is good stuff.

andymass:
You are right, especially with the github interface, and considering so many of the changes are merely due to non-deterministic ordering of the complete command names.

Fortunately, it did not take long to break it up into four commits; first for the `convert-cwl` script, the second two to update the completion data files, and finally the feature code and documentation changes.

autoload/vimtex/complete.vim
```diff
@@ -851,11 +963,12 @@ let s:tex2unicode_list = map([
 
 " }}}1
 
-
```
lervag: I've deliberatly kept two blank lines to more clearly separate the following part from the preceding part. I've tried to do this relatively consistently in the vimtex code. Inspired by python style.

doc/vimtex.txt
```diff
@@ -2359,14 +2360,18 @@ number.  The label completion matches:
...
+                                                 *vimtex-complete-environments*
 
 Command completion is available after `\` and should provide completion
 candidates for relevant LaTeX commands.  The document's preamble is analysed,
-and commands will be completed for the loaded packages. A lot of packages are
-supported, see the path `VIMTEX/autoload/vimtex/complete` for a relevant file
-listing.
+and commands will be completed for the loaded packages as well as those
+defined within the preamble using `\newcommand`, `\let` and `\def`.  Environment
+completion is also available after typing `\begin{` or `\end{`.
```
lervag: The completion for `\end{` should really be deterministic, right? That is, it should match the corresponding `\begin{`... I guess this is some more work to implement, and perhaps should be raised as an issue to further improve things.

doc/vimtex.txt
```diff
@@ -2359,14 +2360,18 @@ number.  The label completion matches:
...
                                                      *vimtex-complete-commands*
+                                                 *vimtex-complete-environments*
 
 Command completion is available after `\` and should provide completion
 candidates for relevant LaTeX commands.  The document's preamble is analysed,
-and commands will be completed for the loaded packages. A lot of packages are
-supported, see the path `VIMTEX/autoload/vimtex/complete` for a relevant file
-listing.
+and commands will be completed for the loaded packages as well as those
+defined within the preamble using `\newcommand`, `\let` and `\def`.  Environment
```
lervag: Should probably also mention `\newenvironment`s?

doc/vimtex.txt
andymass: I didn't mention this in the PR or documentation, but I actually do get the corresponding `\begin{` and show that as the first completion entry, before those from packages (complete.vim:697).

My fear was if parsing fails to find the correct environment name for whatever reason the user won't be showed any alternative entries.  Currently, we're showing both the matching environment and package environments.  Would you like it changed? or should I just update the documentation?

lervag: Ah, that looks good! I didn't notice that. How about this: if you do find the corresponding `begin`, then only complete that, and if not, then show all possibilities.

lervag: I don't think we need to mention it in the docs if it works intuitively.

andymass:
Thanks, updated.  Note that I made a substantial change to improve the `\begin{` matching logic (see complete.vim:695).  Also, `\end{` now suggests packages/newcommands only if finding the matching environment fails.

lervag:
Great! This looks very good! The only thing missing, I think, is a couple of tests in `test/vader/complete.vader`. I think it should be quick to add, but let me know and I could add the tests myself.

andymass:
How about this?

andymass:
Although, these tests seem brittle.  If the completion database were to change the tests would fail.

lervag:
I agree this is brittle, but I think it is OK. It should be easy to update the tests if necessary, and the tests should catch regression errors. Thanks!

PR #1045, opened 2018-02-05
Add documentation pointing to the plugin match-up

andymass:
This adds a brief discussion about enhanced matching provided by the plugin match-up in the "features provided by other plugins" section.  I think no changes to vimtex are necessary.  I have been using the feature for some time and it seems safe to allow the user to enable it if they wish.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -238,6 +238,48 @@ provided by vimtex, but that are instead provided by other plugins.
     [2]: https://github.com/machakann/vim-sandwich
     [3]: http://www.vim.org/scripts/script.php?script_id=3848
 
+  Enhanced matching and higlighting of delimiters~
+    |vimtex| highlights and allows navigation between matching pairs of
+    delimiters including those in math mode, such as `\bigl(` and `\bigr)`, and
+    the `\begin` and `\end` tags of environments.  However, it does not allow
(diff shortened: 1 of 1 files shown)


doc/vimtex.txt
```diff
@@ -238,6 +238,48 @@ provided by vimtex, but that are instead provided by other plugins.
     [2]: https://github.com/machakann/vim-sandwich
     [3]: http://www.vim.org/scripts/script.php?script_id=3848
 
+  Enhanced matching and higlighting of delimiters~
+    |vimtex| highlights and allows navigation between matching pairs of
+    delimiters including those in math mode, such as `\bigl(` and `\bigr)`, and
+    the `\begin` and `\end` tags of environments.  However, it does not allow
+    matching of middle delimiters or `\item`s.
+
+    Alternatively, one may use the plugin |match-up| [1], which offers enhanced
```
lervag: Should be [0], not [1].

lervag:
Except for one very minor comment, I think this looks great. Thanks! I've also started to test matchup on LaTeX documents. So far it looks quite good!

doc/vimtex.txt
andymass: good catch, fixed

lervag:
Great, thanks!

PR #1050, opened 2018-02-12
Fix lcd/cd in last commit

andymass:
Terribly sorry about this, and for catching it only after merge, but I realized I permuted lcd and cd in my PR 😞. The logic should be as follows:

- If *haslocaldir does not exist, assume `lcd` (as done prior to 80b96c1) 
- If window has a local dir, use `lcd`
- If window has no local dir use `cd`


lervag:
No problem! :)

PR #1059, opened 2018-02-21
Extra quotes while preparing commands to run in windows

### LINKED ISSUE #1054: lc got error
When do the clean by `<localleader>lc`, I got error with Vim 8 on Windows 7:

```
Error detected while processing function vimtex#compiler#clean[1]...398[15]..vimtex#process#run[5]..55
line   2:
E484: Can't open file C:\Windows\TEMP\VIo774A.tmp
```

Any friend have any idea on how to fix it? 

Thanks,

image

j-jith:
Added extra quotes while preparing commands to run in windows. Fixes #1054.

### DIFF
--- a/autoload/vimtex/process.vim
+++ b/autoload/vimtex/process.vim
@@ -159,7 +159,7 @@ if has('win32')
       let l:cmd = 'start /b "' . cmd . '"'
     endif
 
-    let self.prepared_cmd = l:cmd
+    let self.prepared_cmd = '"' . l:cmd . '"'
   endfunction
 
   " }}}1


lervag:
Thank you!

vanabel:
@j-jith  @lervag  I see why this fix is not meged, since the above fix will broken the `\lv` command, because in that case we can't run( the double quotes are the error):
```
!"SumatraPDF "/path/of/pdf.pdf""
```
So, there needs another fix. (I also tested that the `\lc` and `\lv` command call the same `l:cmd`)

lervag:
I'm sorry, but as you can probably see from similar issues, e.g. #1062, I'm a little bit frustrated by Windows. I don't know how to fix this kind of thing in a robust manner that "just works" for most people. :)

PR #1069, opened 2018-03-02
Redraw vim after focusing in reverse_goto

andymass:
I use the terminal emulator Konsole.  Since `reverse_goto` was re-written, backwards search from zathura does not update vim's visual file/scroll position, although the terminal is focused, the file is switched, and the real cursor position is moved correctly (just not visually).  I have to actually move the cursor to cause a redraw (I usually just do `jk`).  It's a very minor issue.

I realized recently that `<c-l>` always works properly to update the view.  So, I added a `feedkeys` after the `windowactivate`.  I suppose this is some weird Konsole-only quirk and that most terminals don't need this but it seems harmless in general.  Note `redraw`/`redraw!` do not work because the drawn cursor disappears completely.

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -86,9 +86,12 @@ function! vimtex#view#reverse_goto(line, filename) " {{{1
   normal! zMzvzz
 
   if executable('pstree') && executable('xdotool')
-    let l:xwinids = reverse(split(system('pstree -s -p ' . getpid()), '\D\+'))
+    let l:pids = reverse(split(system('pstree -s -p ' . getpid()), '\D\+'))
 
-    call map(l:xwinids, "system('xdotool search --onlyvisible --pid ' . v:val)[:-2]")
+    let l:xwinids = []
+    call map(copy(l:pids), 'extend(l:xwinids, reverse(split('
+          \ . "system('xdotool search --onlyvisible --pid ' . v:val)"
+          \ . ')))')
     call filter(l:xwinids, '!empty(v:val)')
 
     if !empty(l:xwinids)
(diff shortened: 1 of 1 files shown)


lervag:
Just to be sure: When you supply the `n` flag, this means that `<c-l>` will always be the "default" mapping, right? I.e., custom mappings are not relevant?

andymass:
> When you supply the n flag, this means that <c-l> will always be the "default" mapping, 

Yes, that's correct.

I intended to make this second commit a separate PR but it got placed here, so here are the details:  I use Konsole, and the way it works is with multiple windows all under a single PID.  This means `xdotool search` returns a list instead of a single window id, which vimtex can't handle properly.  Even worse,

From `man xdotool:`
>    Note that the order of windows in the window stack corresponds
>    to the window stacking order, i.e. the bottom-most window will be reported first

This means the window that was taken was always the bottom-most, which is almost always not what one expects.  Currently, each backwards search cycles through a different (visible) Konsole window.  Hence, I use `reverse(` to always use the top-most instead.  This is not perfect, but should be preferable in most cases.

lervag:
> This means the window that was taken was always the bottom-most, which is almost always not what one expects. Currently, each backwards search cycles through a different (visible) Konsole window. Hence, I use reverse( to always use the top-most instead. This is not perfect, but should be preferable in most cases.

Interesting. If I understand correctly, my assumption that the `xdotool search` returns a single ID per PID is not correct, which is why you need the `extends` variant.

I'll check out your branch and check if everything works as expected on my end.

lervag:
Thanks, it works well and seems like a nice improvement!

andymass:
> If I understand correctly, my assumption that the xdotool search returns a single ID per PID is not correct

Yep, apparently it is a list in general:

    $ xdotool search --onlyvisible --pid 3944
    69207024
    69206022

Thanks for including it!

lervag:
Thanks for suggesting the improvement :)

Issue #1063, opened 2018-02-23
zathura compatibility with latexmk setting


chakravala:
On arch linux using latest vim/gvim.

When I launch vimtex, this is my command for it in `.bashrc`
```bash
alias vimtex='vim --servername vim'
```
the configuration is `.vimrc`
```vim
let g:vimtex_view_method = 'zathura'
```

This configuration works fine for me, except that when I launch it and press `\ll` for the first time, zathura is opened and the document is displayed in its previous configuration and then the program freezes for a few seconds, and then it resets the display on zathura with the horizontal scroll reset and a bunch of view settings reset.

However, if I add this to `.vimrc`
```vim
let g:vimtex_compiler_progname = 'latexmk'
```
then the issue goes away and when I press `\ll` the zathura imediately displays the document and does not freeze for a few seconds and **does not** reset my view settings or horizontal scroll config.

**However** now with this setting I can no longer ctrl+click on zathura to communicate back to vimtex where on what line I want the editing cursor to be placed.

How can I make this work using this option? There's probably some simple thing I am overlooking.

Thanks for any assistance.

lervag:
First,

```vim
let g:vimtex_compiler_progname = 'latexmk'
```

does not make any sense. This option is meant to define the name/path to the proper Vim executable for doing remote server/client calls during callbacks.

I suspect that the `.latexmkrc` is relevant here. Could you check if the problem persists if you remove that line from your `.latexmkrc` file?

chakravala:
Then why is it that when I use
```vim
let g:vimtex_compiler_progname = 'latexmk'
```
this solves the issue with zathura resetting my view settings, but I don't really understand what's going on with it. then what is the setting that is supposed to be used for it? It doesn't seem to matter what I enter into the value for that.

Also, the `.latexmkrc` setting does not affect the issue at all, removing it or keeping it doesn't change it. Does it have something to do with how to call pdflatex using synctex?

lervag:
Ok, since `.latexmkrc` is not relevant I'm removing all mentions of it from the thread.

I'm thinking the reason it solves the issue is because the default of `g:vimtex_compiler_progname` is the full path to gvim, which starts with `C:\Program Files`. The problem here is related to escaping during the callbacks from latexmk.

Could you please update the original post to better match the issue template?

chakravala:
[...]

PR #1070, opened 2018-03-02
modified zathura startup behavior

chakravala:
As discussed in #1063, I have made some experimental modifications to the zathura startup behavior.

The following changes make the modifications I wanted available as a config setting via
```vim
let g:vimtex_view_process_start = 0
let g:vimtex_view_xwin_exists = 0
```
(Note that the default is the opposite of this setting, so it is 1,1, this is what I put in `.vimrc`)

And I also changed the timer from 500m to 50m for `xwin_id`, since that is more than is needed.

So this pull request does not change the default behavior for any users, it only provides 2 additional options that skip some things I wanted to skip. The only change that would affect everyone is the change of the `xwin_id` timer to 50m from 500m, but that shouldn't cause any issues, right?

I know you said you don't want these extra settings, but since I was experimenting, I have them here, so if you'd like to merge, that would be great, but I understand you might not want to.

Regards,

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -240,6 +240,8 @@ function! s:init_options() " {{{1
   call s:init_option('vimtex_view_mupdf_options', '')
   call s:init_option('vimtex_view_mupdf_send_keys', '')
   call s:init_option('vimtex_view_zathura_options', '')
+  call s:init_option('vimtex_view_process_start', 1)
+  call s:init_option('vimtex_view_xwin_exists', 1)
 endfunction
 
 " }}}1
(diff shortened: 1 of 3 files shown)


lervag:
No. Thank you for the effort, but you are solving an issue that only you are having. The Zathura viewer works very well for me the way it is now, and this PR would not improve anything for myself or other users who use Zathura (IMHO).

What you _can_ do is to make your own custom viewer. Copy the `zathura.vim` file into your own plugin folders, e.g. `~/.vim/autoload/vimtex/view/my_zathura.vim` (*note* this is a custom path and not related to the vimtex plugin). Change the function names from `#zathura` to `#my_zathura`, then modify the viewer as you like. Now you can use your custom viewer by setting `g:vimtex_view_method = 'my_zathura'`.

PR #1089, opened 2018-03-20
Add an option to disable syntax highlighting.

lgiraldi:
I just added an option to disable vimtex syntax highlighting.

The dependence on TexNewMathZone is a problem if one wants to use a different tex syntax file than the one bundled with vim, see [gi1242/vim-tex-syntax](https://github.com/gi1242/vim-tex-syntax.git) for instance.


lervag:
I agree this option is useful.

PR #1094, opened 2018-04-03
Make b:surround_* suggestions simpler.

doronbehar:
Use `ftplugin/tex.vim` as the best place to put the `b:surround_`
variables.
Add text-object in `ys*` mappings.


kiryph:
Not everyone can edit runtime files of vim, e.g. when working at a place where you do not have super user rights. Consider also work groups where thin-clients are deployed at each desk and you actually work on a central multi-user server with preinstalled vim.

However, it is also possible to place the snippet into `~/.vim/after/ftplugin/tex.vim`.

lervag:
I agree with @kiryph here. But I think it can be simplified, so I'm pushing an update that I think combines @kiryph's concern with your suggestion.

doronbehar:
Great, thanks!

lervag:
No problem :)

Issue #1102, opened 2018-04-11
VimtexClean fails


fuzzybear3965:
# Explain the issue
Trying to run `:VimtexClean` fails on my Windows 10 installation, currently. I've tested using a minimal vimrc and LaTeX file. The error output resembles the screenshot below.

![error](https://user-images.githubusercontent.com/6321578/38589065-ff1a2abc-3cf6-11e8-87a4-977ecd9a1eb1.jpg)

Using the stacktrace in the above screenshot I've traced the source of the error to a [line in `process.vim`](https://github.com/lervag/vimtex/blob/a07cd40cecd339d5a89376d055e96e148a3e0ef2/autoload/vimtex/process.vim#L103). I played around with the `self.prepared_cmd` string and managed to get it to work after a while. The key was to remove the [`start /b`](https://github.com/lervag/vimtex/blob/a07cd40cecd339d5a89376d055e96e148a3e0ef2/autoload/vimtex/process.vim#L159) line from process.vim. This is a poor solution given that this very line is what is supposed to implement background execution. There is almost definitely a better solution, but there is seemingly no proper placement of quotes and spaces which allows this to work.

I've submitted PR #1103 for this.

lervag:
We'll continue this discussion in #1103.

Issue #1062, opened 2018-02-23
Win7 - Latexmk: Stopping because executing following code from command line     $success_cmd = '"C:Program


rjiang9:
When I install vimtex (through vbundle) on a Win7 machine today, I got this error (output of \lo, nothing happens for \ll):

```
Latexmk: Stopping because executing following code from command line
    $success_cmd = '"C:Program
gave an error:
    Can't find string terminator "'" anywhere before EOF at (eval 11) line 1.
```

1. A minimal `vimrc` file, for instance:

```vim
set nocompatible

let &rtp = $HOME. '\\vimfiles\\bundle\\vimtex,' . &rtp
let &rtp .= ','. $HOME . '\\vimfiles\\bundle\\vimtex\after'

filetype plugin indent on
syntax enable

let g:tex_flavor='latex'
let g:vimtex_view_general_viewer = $HOME . '\\VIM\\pdfviewer\\SumatraPDF.exe'
let g:vimtex_view_general_options = '--unique file:@pdf\#src:@line@tex'
let g:vimtex_view_general_options_latexmk = '--unique'
let g:vimtex_compiler_progname = v:progname
```

2. A minimal `.tex` file:
    
```tex
\documentclass{minimal}
\begin{document}
hello you!
\end{document}
```

3. Since this is Win7 machine, when I do `vim --servername VIM -u vimrc mwe.tex`, I got error:

```
  VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Feb  9 2018 13:11:23)
Unknown option argument: "--servername"
More info with: "vim -h"
```

4. A `.latexmkrc` file 

```perl
$pdflatex = 'xelatex --shell-escape %O %S';
```

![image 092](https://user-images.githubusercontent.com/7293118/36578461-9b195df4-181a-11e8-98b2-1d73cdbc577f.jpg)

rjiang9:
Any suggestion would be highly appreciated. 
Thanks.

lervag:
I think this is related to the recent update by #1059, see also #1054. Can you try to revert vimtex to a commit before the one provided by #1059?

lervag:
On windows, instead of `vim --servername ...`, you can do `gvim -u vimrc mwe.tex`.

rjiang9:
@lervag I checkout a old copy of process.vim (the one you committed 2 month ago for removing modeline) and replaced the current one. and it seems I got the same results:
```

Latexmk: Stopping because executing following code from command line
[...]

PR #1103, opened 2018-04-11
fixed :VimtexClean; disables background execution

fuzzybear3965:
Details regarding this PR are in #1102.


lervag:
I think this is related to an earlier PR #1059, which was reverted at a later stage. See also #1062 (which as further links to related issues).

I could accept this, but it will probably break things on other Windows systems...?

PR #1150, opened 2018-06-23
Fix documentation to reflect naming of variables

mgelde:
Just updated vimtex and noticed that the way to globally use xelatex no longer worked. The documentation seems to not have caught up with the code changes.


lervag:
Thanks!

PR #1166, opened 2018-07-15
NCM2: Split omnifunc source based on context

clason:
(This pull request supersedes (because it contains) #1164 and #1165. Commit 94e52c8 is no longer strictly required but may still be sensible. Commit afc8489 is now mainly cosmetic but still useful to avoid extraneous matches when filtering commands. Commit f0fd553 is the actual meat of this PR.)

This pull request adds specific regexp for the ncm2 completion manager. In addition to a full list `g:vimtex#re#ncm2` (which is just a copy of `g:vimtex#re#ncm` for now), it defines the following separate lists:

* `g:vimtex#re#ncm2#cmds`: matches commands, environments, packages, and classes
* `g:vimtex#re#ncm2#bibtex`: matches the argument of citation commands
* `g:vimtex#re#ncm2#labels`: matches the argument of `\*ref` commands
* `g:vimtex#re#ncm2#files`: matches the argument of `\include*` commands

These can be used to define separate ncm2 completion sources with different matching behaviours; see updated documentation. In brief, with the extended example:

* `cmds` get matched on the beginning of `word` (or `abbr`, if this is set to `word`) only
* `bibtex` get matched on beginning of `word` (i.e., bibkey) and fuzzy substring of `abbr` (condensed author list) and `menu` (title)
* `labels` get matched on exact substrings of `word` (the TeX label) and `menu` (the printed label)
* `files` get matched on fuzzy substrings of `word` and `abbr` (full and shortened filename, respectively, although I haven't seen any differences in `v:completed_item` so far -- in fact, might it make sense to show the full filename in the `menu` and use `abbr` only for the filename without path?)

But this can easily and obviously be adjusted to taste (one of the advantages of ncm2 over other plugins). In fact, it might make sense to split the regexps further (e.g., separate packages and classes from commands and environments) to allow further customization.

cc @languitar

### DIFF
--- a/autoload/vimtex/re.vim
+++ b/autoload/vimtex/re.vim
@@ -80,6 +80,47 @@ let g:vimtex#re#ncm = [
       \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
       \]
 
+let g:vimtex#re#ncm2 = [
+      \ '\\[A-Za-z]*',
+      \ '\\[A-Za-z]*cite[A-Za-z]*(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(text|block)cquote\*?(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(for|hy)[A-Za-z]*cquote\*?{[^}]*}(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\[A-Za-z]*ref({[^}]*|range{([^,{}]*(}{)?))',
+      \ '\\hyperref\[[^]]*',
+      \ '\\includegraphics\*?(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(include(only)?|input|subfile){[^}]*',
+      \ '\\\a*(gls|Gls|GLS)(pl)?\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',
+      \ '\\includepdf(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\usepackage(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\documentclass(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\begin(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
+      \]
+let g:vimtex#re#ncm2#cmds = [
+      \ '\\[A-Za-z]*[^{]',
+      \ '\\usepackage(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\documentclass(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\begin(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
+      \]
+let g:vimtex#re#ncm2#bibtex = [
+      \ '\\[A-Za-z]*cite[A-Za-z]*(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(text|block)cquote\*?(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(for|hy)[A-Za-z]*cquote\*?{[^}]*}(\[[^]]*\]){0,2}{[^}]*',
+      \]
+let g:vimtex#re#ncm2#labels = [
+      \ '\\[A-Za-z]*ref({[^}]*|range{([^,{}]*(}{)?))',
+      \ '\\hyperref\[[^]]*',
+      \ '\\\a*(gls|Gls|GLS)(pl)?\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',
+      \]
+let g:vimtex#re#ncm2#files = [
+      \ '\\includegraphics\*?(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(include(only)?|input|subfile){[^}]*',
+      \ '\\includepdf(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
+      \]
+
 let g:vimtex#re#youcompleteme = map(copy(g:vimtex#re#ncm), "'re!' . v:val")
 
 " }}}1
--- a/autoload/vimtex/re.vim
+++ b/autoload/vimtex/re.vim
@@ -80,23 +80,6 @@ let g:vimtex#re#ncm = [
       \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
       \]
 
-let g:vimtex#re#ncm2 = [
-      \ '\\[A-Za-z]*',
-      \ '\\[A-Za-z]*cite[A-Za-z]*(\[[^]]*\]){0,2}{[^}]*',
-      \ '\\(text|block)cquote\*?(\[[^]]*\]){0,2}{[^}]*',
-      \ '\\(for|hy)[A-Za-z]*cquote\*?{[^}]*}(\[[^]]*\]){0,2}{[^}]*',
-      \ '\\[A-Za-z]*ref({[^}]*|range{([^,{}]*(}{)?))',
-      \ '\\hyperref\[[^]]*',
-      \ '\\includegraphics\*?(\[[^]]*\]){0,2}{[^}]*',
-      \ '\\(include(only)?|input|subfile){[^}]*',
-      \ '\\\a*(gls|Gls|GLS)(pl)?\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',
-      \ '\\includepdf(\s*\[[^]]*\])?\s*\{[^}]*',
-      \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
-      \ '\\usepackage(\s*\[[^]]*\])?\s*\{[^}]*',
-      \ '\\documentclass(\s*\[[^]]*\])?\s*\{[^}]*',
-      \ '\\begin(\s*\[[^]]*\])?\s*\{[^}]*',
-      \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
-      \]
 let g:vimtex#re#ncm2#cmds = [
       \ '\\[A-Za-z]*',
       \ '\\usepackage(\s*\[[^]]*\])?\s*\{[^}]*',
@@ -120,6 +103,11 @@ let g:vimtex#re#ncm2#files = [
       \ '\\includepdf(\s*\[[^]]*\])?\s*\{[^}]*',
       \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
       \]
+
+let g:vimtex#re#ncm2 = g:vimtex#re#ncm2#cmds +
+            \ g:vimtex#re#ncm2#bibtex + 
+            \ g:vimtex#re#ncm2#labels + 
+            \ g:vimtex#re#ncm2#files
 
 let g:vimtex#re#youcompleteme = map(copy(g:vimtex#re#ncm), "'re!' . v:val")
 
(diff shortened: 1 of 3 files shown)


clason:
There's still an issue if ncm2 registers sources on ncm2 load (using the `User Ncm2Plugin` as in the ncm2 example), so if the open buffer is not a TeX file, the `g:vimtex#re` strings aren't available and no completion happens even after opening a TeX buffer later on. This can be fixed by 

1. either calling the `register_source` function on `Syntax tex` rather than `User Ncm2Plugin` (which seems to work well and is what I did in the documentation, but will not error out gracefully if ncm2 is not loaded)
2. put the regexes verbatim into the `vimrc`
3. make the regexes available on `vimtex` load, not only on `init()`.

@lervag what do you think? is 3. an option, or is there a (performance) reason for the current behavior?

lervag:
Sorry for taking so long to reply. And I still won't, except to say that I look forward to studying/reviewing this! It looks like interesting, useful and good work!

clason:
No worries; in the meantime I'm dogfooding this via my fork -- and I'm quite happy with it so far :) (Although I did notice that the package completion doesn't trigger on `\RequirePackage{`?)

lervag:
I'll first address your text, then I'll review the code.

> This pull request adds specific regexp for the ncm2 completion manager. In addition to a full list g:vimtex#re#ncm2 (which is just a copy of g:vimtex#re#ncm for now), it defines the following separate lists: ...

> cmds get matched on the beginning of word (or abbr, if this is set to word) only
> bibtex get matched on beginning of word (i.e., bibkey) and fuzzy substring of abbr (condensed author list) and menu (title)
> labels get matched on exact substrings of word (the TeX label) and menu (the printed label)
> files get matched on fuzzy substrings of word and abbr (full and shortened filename, respectively, although I haven't seen any differences in v:completed_item so far -- in fact, might it make sense to show the full filename in the menu and use abbr only for the filename without path?)

I agree this does seem useful, thanks!

> But this can easily and obviously be adjusted to taste (one of the advantages of ncm2 over other plugins). In fact, it might make sense to split the regexps further (e.g., separate packages and classes from commands and environments) to allow further customization.

Yes, possibly. Perhaps one should also name the "full" regex something else to indicate that it includes "everything"? Or, perhaps one should make the full regex a `join()` of the separate lists?

> @lervag what do you think? is 3. an option, or is there a (performance) reason for the current behavior? Also, if we stick with option 1, would Filetype tex be a better event for the autocommand (both seem to work equally well, but I might be missing edge cases)?

Actually, I thought that the regexes should be available on demand due to the autoload script. Are you sure that this: "_so if the open buffer is not a TeX file, the g:vimtex#re strings aren't available and no completion happens even after opening a TeX buffer later on_" is right?

> Although I did notice that the package completion doesn't trigger on `\RequirePackage{`?

That seems right, `\RequirePackage` has not been included. This should be fixed and must be fixed both in the `g:vimtex#re#...` and in the completer code. I've opened a new issue for this: #1172. I'll address that after this PR is completed.

autoload/vimtex/re.vim
```diff
@@ -80,6 +80,47 @@ let g:vimtex#re#ncm = [
...
+      \ '\\includegraphics\*?(\[[^]]*\]){0,2}{[^}]*',
+      \ '\\(include(only)?|input|subfile){[^}]*',
+      \ '\\\a*(gls|Gls|GLS)(pl)?\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',
+      \ '\\includepdf(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\usepackage(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\documentclass(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\begin(\s*\[[^]]*\])?\s*\{[^}]*',
+      \ '\\end(\s*\[[^]]*\])?\s*\{[^}]*',
+      \]
```
lervag: Is this essentially a copy of `g:vimtex#re#ncm`? If so, I think it is better to make that explicit.

However, it could also make sense to make this regex be a combination of the separate `#ncm2` lists.

doc/vimtex.txt
```diff
@@ -2757,13 +2757,72 @@ The following configuration should work well with `vimtex`: >
...
             \ 'mark': 'tex',
             \ 'word_pattern': '\w+',
-            \ 'complete_pattern': g:vimtex#re#ncm,
+            \ 'complete_pattern': g:vimtex#re#ncm2,
+            \ 'on_complete': ['ncm2#on_complete#omni', 'vimtex#complete#omnifunc'],
+            \ })
+  augroup END
+<
+To have autocomplete behave more intelligently like omni-complete, use the
+following setup instead: >
```
lervag: Perhaps rather "To have more intelligent omni-complete like autocompletion, ..."?

clason:
> Yes, possibly. Perhaps one should also name the "full" regex something else to indicate that it includes "everything"? Or, perhaps one should make the full regex a join() of the separate lists?

Yes, that makes sense to avoid duplication and danger adjusting one and not the other. (I used `#ncm2` for consistency with the other plugins so if you're not aware of the possibility of subscoping, you could just use the "standard" name.)

> Actually, I thought that the regexes should be available on demand due to the autoload script. Are you sure that this: "so if the open buffer is not a TeX file, the g:vimtex#re strings aren't available and no completion happens even after opening a TeX buffer later on" is right?

At least they're not available to ncm2 (although you can in fact `:echo` them). Not sure why -- maybe a race condition? -- but the `Filetype tex` seems to take care of that, and postponing registering a source until you actually need it seemed like a good idea as well.

> That seems right, \RequirePackage has not been included. This should be fixed and must be fixed both in the g:vimtex#re#... and in the completer code. I've opened a new issue for this: #1172. I'll address that after this PR is completed.

Indeed, that's a separate issue; I just noticed it in passing and didn't want to forget.

doc/vimtex.txt
```diff
@@ -2757,13 +2757,72 @@ The following configuration should work well with `vimtex`: >
...
+            \ 'complete_length': -1,
+            \ 'scope': ['tex'],
+            \ 'matcher': {'name': 'combine',
+            \             'matchers': [
+            \               {'name': 'prefix', 'key': 'word'},
+            \               {'name': 'abbrfuzzy', 'key': 'abbr'},
+            \               {'name': 'abbrfuzzy', 'key': 'menu'},
+            \             ]},
+            \ 'word_pattern': '\w+',
+            \ 'complete_pattern': g:vimtex#re#ncm2#bibtex,
```
lervag: This feels a little bit too big to put in ones vimrc file. Should we suggest that one puts this in a personal `ftplugin/tex.vim` file? Or would that not work? Is it possible to do something like this instead (pseudocode)?

```vim
" Inside "personal/ftplugin/tex.vim"
if Ncm2PluginLoaded()
  call ncm2#register_source({...})
  call ncm2#register_source({...})
  ...
endif
```

lervag:
> Yes, that makes sense to avoid duplication and danger adjusting one and not the other. (I used #ncm2 for consistency with the other plugins so if you're not aware of the possibility of subscoping, you could just use the "standard" name.)

> At least they're not available to ncm2 (although you can in fact :echo them). Not sure why -- maybe a race condition? -- but the Filetype tex seems to take care of that, and postponing registering a source until you actually need it seemed like a good idea as well.

Strange. I guess it is not a problem to use the `Filetype tex` event, in any case, so let's leave it at that. Although, as commented in the review, perhaps one could also briefly mention the possibility of calling the `ncm2#register...` from ones personal `ftplugin/tex.vim` file, if desired. No need of autocommands then, but then we need to check if ncm2 is loaded...

> Indeed, that's a separate issue; I just noticed it in passing and didn't want to forget. Something else I noticed -- and am not sure whether it's a regex or ncm2 issue -- is that command completion is triggered when typing \\. My regex-fu is too weak to prevent that, though...

Ah, yes. Easy to fix: change `*` to `+` to force word-letter before triggering. Or something else, like `{2,}` or similar, to set number of letters required. However, I think it could be included in #1172, so let's leave this there instead.

doc/vimtex.txt
clason: Sure.

autoload/vimtex/re.vim
```diff
@@ -121,6 +104,11 @@ let g:vimtex#re#ncm2#files = [
       \ '\\includestandalone(\s*\[[^]]*\])?\s*\{[^}]*',
       \]
 
+let g:vimtex#re#ncm2 = g:vimtex#re#ncm2#cmds +
+            \ g:vimtex#re#ncm2#bibtex + 
+            \ g:vimtex#re#ncm2#labels + 
+            \ g:vimtex#re#ncm2#files
+
```
lervag: Finally, how about letting `g:vimtex#re#ncm = copy(g:vimtex#re#ncm2)`?

doc/vimtex.txt
```diff
@@ -2766,8 +2766,8 @@ The following simple configuration should work well with `vimtex`: >
             \ })
   augroup END
 <
-To have autocomplete behave more intelligently like omni-complete, use the
-following setup instead: >
+To have more intelligent autocompletion like omni-complete, use the following
```
lervag: Sorry, but I meant "To have more intelligent omni-complete like autocompletion".

clason:
> Strange.

I don't understand it myself, but it's a fact that the vimtex regexes are not available when ncm2 is loaded. Maybe it's because ncm2 is a python plugin, and these get triggered before viml plugins?

doc/vimtex.txt
clason: That's ambiguous ;) You mean `omni-complete-like autocompletion`?

autoload/vimtex/re.vim
clason: Good idea!

clason:
> > Indeed, that's a separate issue; I just noticed it in passing and didn't want to forget. Something else I noticed -- and am not sure whether it's a regex or ncm2 issue -- is that command completion is triggered when typing \. My regex-fu is too weak to prevent that, though...

> Ah, yes. Easy to fix: change * to + to force word-letter before triggering. Or something else, like {2,} or similar, to set number of letters required. However, I think it could be included in #1172, so let's leave this there instead.

And string literals bite me again: I meant command completion gets triggered on ``\\`` -- I don't mind getting completion right after `\`, although it might be too much noise for some. (I can live without it.)

doc/vimtex.txt
```diff
@@ -2766,7 +2766,7 @@ The following simple configuration should work well with `vimtex`: >
             \ })
   augroup END
 <
-To have more intelligent autocompletion like omni-complete, use the following 
+To have more intelligent, omni-complete-like, autocompletion, use the following
```
lervag: I think this works fine. The initial thought was just to make the sentense flow a little bit better. I'm not 100% sure if it does so now, but I think it is slightly better. Regardless, not worth much more work... :+1:

clason: Too late :)

lervag:
Great. I think we are there now. I have not yet tested ncm2, so I'll allow one or two more days before merging, just in case you should happen to notice something that stopped working after the recent changes.

By the way, off topic: have you had some experience with deoplete, and as such, are you able to give a short comment on how ncm2 works vs deoplete?

clason:
Great! I've been using this for about a week with no complaints so far (although I've been going back and forth over whether I prefer `substr` or `abbrfuzzy` for labels...)

I was using deoplete for a long time myself before switching; initially out of curiosity, but the fuzzier matching sold me. It also feels snappier (although not by a lot). The main design difference seems to be that ncm2 is "opt-in" configuration (although there are a lot of plugins that handle that for specific purposes such as snippet autocompletion -- UltiSnips for me) rather than deoplete's "all-in" approach. That makes it easier to customize on a vimrc (or ftplugin) level, too. So for me there's been a gain in functionality without any losses (everything that used to work in deoplete also works in ncm2, and then some), but I've only really used it for TeX; not sure how the situation is for, e.g., Python or Language Server Protocols (although the plugins are available, and probably easily converted from neovim-completion-manager if not).

languitar:
I was also using deoplete but switched to ncm2 recently because there is a working integration between snipppets from language servers and ultisnips with ncm2. For deoplete, custom hacks are required right now.

lervag:
Thanks for the info, both of you! I'll probably try it out when I get the time.

lervag:
I've merged this now. I've tested and it seems to work quite well.

PR #1176, opened 2018-07-28
Parse $pdf_mode variable in latexmkrc

jipolanco:
This PR adds support for parsing the `$pdf_mode` variable in a latexmkrc file, in a similar way as it was already done with the `$out_dir` variable.

Previously, the value of `$pdf_mode` was ignored by vimtex. In particular, when `$pdf_mode` was set and a TeX directive was *not* given, vimtex would use the default latexmk engine (the value of `g:vimtex_compiler_latexmk_engines['_']`, by default `-pdf`, i.e. pdflatex) instead of `$pdf_mode`. This is probably not the outcome expected by the user, especially if they had not set the default latexmk engine explicitly.

The value of `$pdf_mode` now has priority over TeX directives found in the source files, and a warning is shown if both values are given and are inconsistent (in the same way as it was already done with `build_dir / $out_dir`).

I tried to follow the existent code convention and the `CONTRIBUTING.md` file, but there may be better ways to implement this PR, in particular because I'm not very fluent in vimscript.

I also added a small test associated to the PR. I'm not sure if the location of the test files is the best, so feel free to (re)move them as needed.

### DIFF
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -86,8 +86,8 @@ endfunction
 
 " }}}1
 function! s:compiler.init_pdf_mode_option() abort dict " {{{1
-  "
-  " Check if .latexmkrc sets the pdf_mode - if so this should be respected
+  " If the TeX program directive was not set, and if the pdf_mode is set in a
+  " .latexmkrc file, then deduce the compiler engine from the value of pdf_mode.
 
   " Parse the pdf_mode option. If not found, it is set to -1.
   let l:pdf_mode = s:parse_latexmkrc_option(self.root, 'pdf_mode', 1, -1)
@@ -104,14 +104,16 @@ function! s:compiler.init_pdf_mode_option() abort dict " {{{1
     return
   endif
 
-  if self.tex_program !=# '_' && self.tex_program !=# l:tex_program
+  if self.tex_program ==# '_'  " the TeX program directive was not specified
+      let self.tex_program = l:tex_program
+  elseif self.tex_program !=# l:tex_program
     call vimtex#log#warning(
-          \ 'Setting pdf_mode from latexmkrc overrides tex_program!',
-          \ 'Changed tex_program from: ' . self.tex_program,
-          \ 'Changed tex_program to: ' . l:tex_program)
+          \ 'Value of pdf_mode from latexmkrc is inconsistent with ' .
+          \ 'TeX program directive!',
+          \ 'TeX program: ' . self.tex_program,
+          \ 'pdf_mode:    ' . l:tex_program,
+          \ 'The value of pdf_mode will be ignored.')
   endif
-
-  let self.tex_program = l:tex_program
 endfunction
 
 " }}}1
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -51,6 +51,7 @@ function! s:compiler.init(options) abort dict " {{{1
 
   call self.init_check_requirements()
   call self.init_build_dir_option()
+  call self.init_pdf_mode_option()
 
   call extend(self, deepcopy(s:compiler_{self.backend}))
 
@@ -110,6 +111,36 @@ function! s:compiler.init_build_dir_option() abort dict " {{{1
     endif
     let self.build_dir = l:out_dir
   endif
+endfunction
+
+" }}}1
+function! s:compiler.init_pdf_mode_option() abort dict " {{{1
+  "
+  " Check if .latexmkrc sets the pdf_mode - if so this should be respected
+
+  " Parse the pdf_mode option. If not found, it is set to -1.
+  let l:pdf_mode = self.init_parse_option('pdf_mode', 1, -1)
+
+  " If pdf_mode has a supported value (1: pdflatex, 4: lualatex, 5: xelatex),
+  " override the value of self.tex_program.
+  if l:pdf_mode == 1
+    let l:tex_program = 'pdflatex'
+  elseif l:pdf_mode == 4
+    let l:tex_program = 'lualatex'
+  elseif l:pdf_mode == 5
+    let l:tex_program = 'xelatex'
+  else
+    return
+  endif
+
+  if self.tex_program !=# '_' && self.tex_program !=# l:tex_program
+    call vimtex#log#warning(
+          \ 'Setting pdf_mode from latexmkrc overrides tex_program!',
+          \ 'Changed tex_program from: ' . self.tex_program,
+          \ 'Changed tex_program to: ' . l:tex_program)
+  endif
+
+  let self.tex_program = l:tex_program
 endfunction
 
 " }}}1
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -66,6 +66,35 @@ function! s:compiler.init(options) abort dict " {{{1
 endfunction
 
 " }}}1
+function! s:compiler.init_parse_option(option, is_integer, default) abort dict " {{{1
+  "
+  " Parse option from .latexmkrc, returning its value if the option was
+  " present or a default value if not.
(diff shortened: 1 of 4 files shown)


autoload/vimtex/compiler/latexmk.vim
```diff
@@ -66,12 +67,15 @@ function! s:compiler.init(options) abort dict " {{{1
 endfunction
 
 " }}}1
-function! s:compiler.init_build_dir_option() abort dict " {{{1
-  "
-  " Check if .latexmkrc sets the build_dir - if so this should be respected
+function! s:compiler.init_parse_option(option, is_integer, default) abort dict " {{{1
```
lervag: I like that you refactor the common code into a new function, but it should not be part of the `s:compiler` dict. It is a "pure" function, and so I think this function should be renamed to `s:parse_latexmkrc_option` and moved to the bottom of the file.

autoload/vimtex/compiler/latexmk.vim
```diff
@@ -81,19 +85,62 @@ function! s:compiler.init_build_dir_option() abort dict " {{{1
...
-                \ 'Setting out_dir from latexmkrc overrides build_dir!',
-                \ 'Changed build_dir from: ' . self.build_dir,
-                \ 'Changed build_dir to: ' . l:out_dir[1])
-        endif
-        let self.build_dir = l:out_dir[1]
-        return
-      endif
+      let l:match = matchlist(readfile(l:file), l:pattern)
+      if len(l:match) > 1
+        return l:match[1]
```
lervag: Why not `l:match[0]`?

lervag:
Oh, and regarding test files: I think it is good that you include it. My test setup is not fully consistent, and I've often chosen to use `issue/#` for most tests, even when they are new features. However, I think your approach is good and I should probably have done the same more consistently in the past.

clason:
For what it's worth, I agree that there should be a strict global to local hierarchy where the more local directives take precedence, e.g.,

1. `.vimrc`
2. global `.latexmkrc`
3. local `.latexmkrc`
4. `TEX program` directive

(One thing to keep in mind is that one might collaborate with others who do not use the same toolchain  -- vimtex and/or latexmk -- and things should work for everyone, with the `TEX program` being the most widely used.)

autoload/vimtex/compiler/latexmk.vim
jipolanco: Because `l:match[0]` contains the whole matched pattern, while `l:match[1]` contains the first capture group, which is what we're after.

As an example, if the `latexmkrc` file contains the line

    $pdf_mode = 4;  # compile with lualatex

then `l:match[0]` is `$pdf_mode = 4;  # compile with lualatex` and `l:match[1]` is `4`.

jipolanco:
I just pushed a new version that will prioritise TeX program directives over the value of `$pdf_mode`. The parser function was also renamed as proposed by @lervag.

andymass:
Sorry for my ignorance on this, but can you explain why vimtex need to know the $pdf_mode?  I had assumed if there is no TeX-Program directive, latexmk would be called without flags, so whatever configuration contained within the latexmkrc should just work, same as if you had called latexmk from the command line.

clason:
@andymass If I understand things correctly, vimtex adds a `-pdf` flag to `latexmk` by default, overriding the `$pdf_mode`. (In fact, it should be discussed whether a better fix wouldn't in fact be not to set any flags at all and to completely leave latexmk configuration to the global/local `.latexmkrc` fields -- arguably, that would be more in the "non-features" spirit of `vimtex`...)

jipolanco:
@clason Yes that's exactly it. If vimtex didn't add the `-pdf` flag, and if `$pdf_mode` wasn't set, then latexmk would not generate a pdf at all, instead generating a dvi file with `latex`. I guess that generating a pdf instead of a dvi file is a reasonable default behaviour for vimtex, so adding the `-pdf` flag would be justified. I also agree that a better fix would be to just remove the `-pdf` flag if the user has explicitly set `$pdf_mode`.

clason:
@jipolanco Actually, I would go one step further and say that generating a pdf (i.e., calling `pdflatex`) would (now) be the reasonable default for latexmk.

@lervag Is there any specific reason you originally included the latexmk configuration in the vimtex config rather than directly in the (global) latexmkrc?

lervag:
@andymass and @jipolanco Yes, @clason understands things correctly.

@clason Personally, I would not mind "simplifying" and making most or all latexmk settings configured though a global or local `latexmkrc` file, and then simply run `latexmk [file]` from vimtex. However, I think lot of users expect things to "just work" and will open issues to ask for things that don't, even if this is clearly written in the docs. Thus, I've made some choices under the assumption that there would be far fewer issues if I made a sensible default that works for most people. And `-pdf` is (IMHO) a very sensible default that works for most people.

I would not be surprised if we could simplify things for the better. At the same time, I think things mostly work as expected as it is, and especially after merging this PR.

Note that if you read `:h vimtex-latexmk`, you may find this text:

> As stated, one may customize the `latexmk` options through
> |g:vimtex_compiler_latexmk|.  However, one may also configure `latexmk`
> explicitly through a global `~/.latexmkrc` file, or a project specific
> `.latexmkrc` file. It is important to know that command line arguments have
> priority, so one may want to use custom options if one wants to specify
> particular things in a configuration file.

One might improve this to indicate that the `g:vimtex_compiler_latexmk_engines` is also an important configuration option.

autoload/vimtex/compiler/latexmk.vim
lervag: Ah, yes, of course. Thanks :)

clason:
@lervag I understand that motivation. Personally (for what it's worth), I'd punt all compiler configuration to the respective configuration files (as I believe you do for `arara`) -- people who use vim as their editor (and vimtex as their plugin) should be trusted to know how to configure latexmk. (At most, you'd get one issue where this can be explained, and Google will do the rest.)

As a compromise, one could set the latexmk defaults *only* if no global or local latexmkrc is found. This would set up people who want things to "just work" and allow others who know what they're doing to set things up without interference. (For example, I always force `-shell-escape`, and I only use `lualatex` for a few specific documents using corporate fonts.)

lervag:
@clason I think we both agree and disagree on this subject.

> people who use vim as their editor (and vimtex as their plugin) should be trusted to know how to configure latexmk. (At most, you'd get one issue where this can be explained, and Google will do the rest.)

Yes, I would have hoped so. But my experience is that I have spent quite some time helping people with previously solved issues (even if it is only to refer them to the previous discussion). My impression is that a lot of Vim users are relatively inexperienced, and there seems to be real value in trying to make things "just work" for most people.

E.g., if I were to remove `-synctex=1`, then some people would surely open an issue complaining that synctex does not work. I would explain the problem pretty easily, but I would not be surprised if duplicate issues were raised. Similar for `-file-line-error` (necessary for parsing some of the errors/warnings). Further, the `-pvc` flag which is implicitly used with the `'continuous'` flag does also make sense to me.

However, if you think that the current version makes things inconsistent or "wrong" for experienced/expert users, then I would be willing to consider to change them. If so, please open a new issue to continue this discussion. Else I think we should get back to the current topic.

----

Regardless of the meta-discussion with @clason, I think this PR is still useful as it solves an inconsistency.

lervag:
@jipolanco What do you think about my previous comment?

> One thing, though: Does it make sense to give a warning here? I'm thinking perhaps one should give a warning only if the conflict is between a local rc file vs TeX directive. Perhaps one could add a return value to the new function so that it both returns the parsed value and if the rc file is local or "global"?

jipolanco:
@lervag Sorry for the delay. Yes, I agree with you. It makes sense to show a warning only when the latexmkrc file is local to the project. I'll try to modify the code accordingly and update the PR.

lervag:
Great, thanks!

jipolanco:
Ok, so I implemented this based on what you proposed. It's probably not the most elegant possible solution, but it seems to work correctly :)

lervag:
Thanks! I think this looks very good. I'm merging now.

Issue #1177, opened 2018-07-29
How to autofocus on skim after \lv


znculee:
Thanks for your great plugin. I'm using OSX10.13 with skim and macvim. Everything works fine, but I want the skim be focused after pressing \lv. Is there any configuration can make that?

lervag:
I'm sorry, but I don't really know. Perhaps @clason is able to help here?

lervag:
Thanks, we'll continue this discussion in #1178.

PR #1178, opened 2018-07-29
activate skim after VimtexView

znculee:
As I said in [Issue #1177](https://github.com/lervag/vimtex/issues/1177), activating skim after pressing \lv should be the default setting. Because skim have been able to update automatically after saving, the skim should always be expected to be focused after pressing \lv for forward search. I will pull a request for this setting.

### DIFF
--- a/autoload/vimtex/view/skim.vim
+++ b/autoload/vimtex/view/skim.vim
@@ -41,10 +41,6 @@ function! s:skim.view(file) dict " {{{1
   endif
   if vimtex#view#common#not_readable(outfile) | return | endif
 
-  let l:activate = ''
-  if g:vimtex_view_skim_activate
-    let l:activate = '-e ''activate'''
-  endif
   let l:cmd = join([
         \ 'osascript',
         \ '-e ''set theLine to ' . line('.') . ' as integer''',
@@ -58,8 +54,8 @@ function! s:skim.view(file) dict " {{{1
         \ '-e ''end try''',
         \ '-e ''open theFile''',
         \ '-e ''tell front document to go to TeX line theLine from theSource',
-        \ '     showing reading bar true''',
-        \ l:activate,
+        \ g:vimtex_view_skim_reading_bar ? 'showing reading bar true''' : '''',
+        \ g:vimtex_view_skim_activate ? '-e ''activate''' : '',
         \ '-e ''end tell''',
         \])
 
@@ -94,7 +90,6 @@ function! s:skim.compiler_callback(status) dict " {{{1
         \ '-e ''open theFile''',
         \ '-e ''end tell''',
         \])
-
 
   let self.process = vimtex#process#start(l:cmd)
 
--- a/autoload/vimtex/view/skim.vim
+++ b/autoload/vimtex/view/skim.vim
@@ -41,23 +41,42 @@ function! s:skim.view(file) dict " {{{1
   endif
   if vimtex#view#common#not_readable(outfile) | return | endif
 
-  let l:cmd = join([
-        \ 'osascript',
-        \ '-e ''set theLine to ' . line('.') . ' as integer''',
-        \ '-e ''set theFile to POSIX file "' . outfile . '"''',
-        \ '-e ''set thePath to POSIX path of (theFile as alias)''',
-        \ '-e ''set theSource to POSIX file "' . expand('%:p') . '"''',
-        \ '-e ''tell application "Skim"''',
-        \ '-e ''try''',
-        \ '-e ''set theDocs to get documents whose path is thePath''',
-        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
-        \ '-e ''end try''',
-        \ '-e ''open theFile''',
-        \ '-e ''tell front document to go to TeX line theLine from theSource',
-        \ '     showing reading bar true''',
-        \ '-e ''activate''',
-        \ '-e ''end tell''',
-        \])
+  if !g:vimtex_view_skim_activate
+    let l:cmd = join([
+          \ 'osascript',
+          \ '-e ''set theLine to ' . line('.') . ' as integer''',
+          \ '-e ''set theFile to POSIX file "' . outfile . '"''',
+          \ '-e ''set thePath to POSIX path of (theFile as alias)''',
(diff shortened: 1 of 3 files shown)


clason:
I strongly disagree; in the vast majority of cases, I want to continue typing even after manually updating the PDF view. At the very least, this should be made configurable.

autoload/vimtex/view/skim.vim
```diff
@@ -41,23 +41,42 @@ function! s:skim.view(file) dict " {{{1
...
+          \ '-e ''set theSource to POSIX file "' . expand('%:p') . '"''',
+          \ '-e ''tell application "Skim"''',
+          \ '-e ''try''',
+          \ '-e ''set theDocs to get documents whose path is thePath''',
+          \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
+          \ '-e ''end try''',
+          \ '-e ''open theFile''',
+          \ '-e ''tell front document to go to TeX line theLine from theSource',
+          \ '     showing reading bar true''',
+          \ '-e ''activate''',
```
clason: Thanks for making this configurable! I think it might be possible to keep a single `join` command with a string variable `s:activate` which is either empty or contains `-e ''activate''`, depending on the value of `g:vimtex_view_skim_activate`:

```
    let l:activate = ''
    if g:vimtex_view_skim_activate
       let l:activate = '-e ''activate'''
    endif

    let l:cmd = join([
          \ 'osascript',
          \ '-e ''set theLine to ' . line('.') . ' as integer''',
          \ '-e ''set theFile to POSIX file "' . outfile . '"''',
          \ '-e ''set thePath to POSIX path of (theFile as alias)''',
          \ '-e ''set theSource to POSIX file "' . expand('%:p') . '"''',
          \ '-e ''tell application "Skim"''',
          \ '-e ''try''',
          \ '-e ''set theDocs to get documents whose path is thePath''',
          \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
          \ '-e ''end try''',
          \ '-e ''open theFile''',
          \ '-e ''tell front document to go to TeX line theLine from theSource',
          \ '     showing reading bar true''',
         \ l:activate,
          \ '-e ''end tell''',
          \])
```


 That should make this easier to maintain.

doc/vimtex.txt
```diff
@@ -2053,6 +2053,11 @@ Options~
 
   Default value: ''
 
+*g:vimtex_view_skim_activate*
+  Use this option to enable/disable activate skim after VimtexView.
```
clason: Maybe more explicitly `Set this option to 1 make Skim have focus after activation in addition to being moved to the foreground.`?

znculee:
@clason @lervag I have updated the code and the README file as your recommendations. It is worth to note that skim only will be activate after \lv but not \ll. Because the activation of skim is only for the forward search. Mostly, we might prefer to keep typing after starting the complier.

znculee:
@clason When I testing forward search, I found the forward only works in the macvim without fullscreen mode. The forward search will not work in CLI vim (without clientserver) and macvim with fullscreen. Anyway, this is not very related with this pull request.

doc/vimtex.txt
```diff
@@ -2054,7 +2054,8 @@ Options~
   Default value: ''
 
 *g:vimtex_view_skim_activate*
-  Use this option to enable/disable activate skim after VimtexView.
+  Set this option to 1 make Skim have focus after commond :VimtexView in
```
clason: typo: "command"

clason:
@znculee Indeed, the "silent" update on compiler callback (`ll`) is by design.

The problem with forward search in fullscreen sounds like a macvim bug (I'm using neovim/vimR, so I can't test).

Now that you're starting to configure the skim forward search, maybe it would make sense to also make the reading bar (the line `\  '     showing reading bar true'''`) configurable? (Not necessarily as part of this pull request, though.)

doc/vimtex.txt
lervag: I would change to: "Set this option to 1 __to__ make Skim have focus after **command** |:VimtexView| in". Note the `|`s as well.

autoload/vimtex/view/skim.vim
```diff
@@ -55,6 +59,7 @@ function! s:skim.view(file) dict " {{{1
         \ '-e ''open theFile''',
         \ '-e ''tell front document to go to TeX line theLine from theSource',
         \ '     showing reading bar true''',
+        \ l:activate,
```
lervag: You could also combine 44-47 like this:

```vim
    \ g:vimtex_view_skim_activate ? '-e ''activate''' : '',
```

I find this way equally readable and more condensed.

autoload/vimtex/view/skim.vim
```diff
@@ -89,7 +94,8 @@ function! s:skim.compiler_callback(status) dict " {{{1
         \ '-e ''open theFile''',
         \ '-e ''end tell''',
         \])
-  
+
+
```
lervag: I would prefer to keep a single line here.

lervag:
I think it is better to suggest an option for showing the reading bar as a separate issue. If anyone wants it, they could raise the issue.

Else, I think this PR is close to ready after making the suggested updates.

znculee:
@clason @lervag I have updated the code as your suggestions, and also make the reading bar configurable. The default setting of the reading bar is the same as original implementation. Indeed, sometimes I also think the reading bar is annoying. Thank you for your great comments.

doc/vimtex.txt
```diff
@@ -2054,11 +2054,16 @@ Options~
...
 
 *g:vimtex_view_skim_activate*
-  Set this option to 1 make Skim have focus after commond :VimtexView in
+  Set this option to 1 to make Skim have focus after command |:VimtexView| in
   addition to being moved to the foreground.
 
   Default value: 0
 
+*g:vimtex_view_skim_reading_bar*
+  Set this option to 0 to hidden reading bar after command |:VimtexView|.
```
clason: `Set this option to 1 to highlight current line in PDF after command |:VimtexView|.`?

znculee:
@lervag I have updated the doc as your suggestion.

lervag:
I think this looks good now. I'll merge. Thanks for the contribution, @znculee!

Issue #553, opened 2016-09-14
pgrep is not executable


nowox:
I am running on Cygwin where there is no `pgrep` alternative. So I get the warning `vimtex warning: pgrep is not executable` which is not just a warning because `vimtex#latexmk was not initialized` and and I cannot use any mapping such as `<leader>ll`. 

How can I make `vimtex` run on cygwin?

lervag:
Good question. I don't know. I don't use Windows and I don't use cygwin. Is it not possible to install pgrep in cygwin? A [quick google search](https://www.google.no/search?q=pgrep+in+cygwin&btnI=Im+Feeling+Lucky&gws_rd=cr&ei=P7nZV8DkF8eRsgH1wLzoBw) lead me [here](http://unix.stackexchange.com/questions/19359/alternative-to-pkill-or-pgrep-f-on-cygwin), which seems to indicate that you can install pgrep through the [procps](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Fprocps%2Fprocps-3.2.8-5&grep=pgrep) package.

lervag:
I closed this, because 1) it seems you can install pgrep, and 2) if not, then I can't help you. You will have to figure this out yourself. If you do find a solution, then I would be interested in pull requests or suggestions for how to make vimtex work for cygwin, either through improved documentation or new code.

lgalke:
Can confirm that it worked after installing procps cygwin package. 
I needed to issue `apt-cyg download procps` before `apt-cyg install procps`.
Great experience in conjunction with Sumatra.pdf, the only windows viewer that does not completely lock open files for writing.

PR #1184, opened 2018-08-02
Added win32unix compatibility

jjwallman:
pgrep is not executable on win32unix (see [this issue](https://github.com/lervag/vimtex/issues/553)).

The relevant tests only tested has('win32') and so did not identify this condition.


lervag:
Thanks!

Issue #1106, opened 2018-04-11
Add Table of Figures/Tables


lervag:
Cf. #1085.

Note: The direction is now going to "Table of Contents/TODOs/Labels/Figures/Tables", i.e. 5 different things. I'm thinking we should change the default mappings to make all of these lists more available. Perhaps something like a choice menu? I would be happy to get some ideas/suggestions/comments on this.

clason:
Would a three-letter mapping be possible: `<leader>lt[cdlft]`?

lervag:
Yes, I think it could. I also think it could be nice with a mini-menu for assisting with the final letter. That is, mapping `<localleader>lt` to a function that displays a menu and asks for the final letter. Perhaps, if possible, only showing the menu if there is a short delay.

Note: It goes without saying that I will add `<plug>` mappings for all the various commands for those who want that, which would allow to keep the old behaviour if desired.

lervag:
I've opened a pull request that should address this issue, and as such, I am closing this now.

Issue #1101, opened 2018-04-09
Separate list of TODOs


lervag:
Cf. #1085: @kiryph propose to add a dedicated table of todos similar to the table of contents. The following features should be considered:

- [ ] `:VimtexToT` and `<plug>(vimtex-tot)` (perhaps also a default mapping?)
- [ ] Denite source
- [ ] Option to hide TODOs from ToC

----

Should not work on this until #1107 has been resolved.

dbmrq:
As the one who suggested adding TODOs to the ToC, I just want to add that a dedicated TODO list would be great, it just seemed easier to use the ToC at the time.


let g:VimtexImportante = {
      \ 're' : g:vimtex#re#not_bslash . '\%\c\s*IMPORTANTE\s*:?\s*\zs.*',
      \ 'in_preamble' : 1,
      \}

function! g:VimtexImportante.get_entry(context) abort dict
  return {
        \ 'title'  : 'IMPORTANTE: ' .
            \ matchstr(a:context.file, '\(\/.*\).*\/\zs\a.*\.\a\a\a') .
            \ ' - line ' . a:context.lnum,
        \ 'number' : '[!]',
        \ 'file'   : a:context.file,
        \ 'line'   : a:context.lnum,
        \ 'level'  : a:context.max_level,
        \ 'rank'   : a:context.lnum_total,
        \ }
endfunction

let g:vimtex_toc_custom_matchers = [g:VimtexImportante]

dbmrq:
As the one who suggested adding TODOs to the ToC, I just want to add that I agree a dedicated TODO list would be great. It just seemed easier to use the ToC at the time.

It would also be nice if it were possible to add other custom words to be matched.

I have the following custom ToC matcher, for instance:

```vim
let g:VimtexImportant = {
      \ 're' : g:vimtex#re#not_bslash . '\%\c\s*IMPORTANT\s*:?\s*\zs.*',
      \ 'in_preamble' : 1,
      \}

function! g:VimtexImportant.get_entry(context) abort dict
  return {
        \ 'title'  : 'IMPORTANT: ' .
            \ matchstr(a:context.file, '\(\/.*\).*\/\zs\a.*\.\a\a\a') .
            \ ' - line ' . a:context.lnum,
        \ 'number' : '[!]',
        \ 'file'   : a:context.file,
        \ 'line'   : a:context.lnum,
        \ 'level'  : a:context.max_level,
        \ 'rank'   : a:context.lnum_total,
[...]

PR #1191, opened 2018-08-13
Better TOC feature

lervag:
This PR is still work-in-progress! It should address #1106 and #1101, and as such, I'm closing both of those issues now.

Remaining work:
- [x] Allow custom maps/commands to prefiltered content in the ToC
- [x] Allow custom words for todo items (see [here](https://github.com/lervag/vimtex/issues/1101#issuecomment-380199414))
- [x] Clean up code (and git commits)
- [x] Rename option `g:vimtex_toc_types` to `..._layers`
- [x] Update documentation, including some info on the API with respect to denite/unite, etc
- [x] Add/Update denite/unite/etc source
- [x] Possibly ensure that the active layers persist when closing/reopening the TOC window

The current idea is to have one TOC to unite them all, and to allow to customize the default content, as well as simple keybindings to toggle what is displayed. I've dubbed the various content types for "layers". Right now, there is an option `g:vimtex_toc_layers` that is not documented, but the default value should make it quite clear how it works.

In #1106, it was suggested to add table of Figures/Tables, but I am wondering if this is not already provided through the `label` type. If anyone still thinks there should be a layer for Figures/Tables, then please explain how this should work as opposed to the layer for labels.

Perhaps it would also be nice with a simple way to map the TOC window with different layers activated, although I think for most people, it should be enough to be able to customize the default layers and that these are easy to toggle within the TOC window.


lervag:
I think this feature is ready to be tested by others. I would be happy if perhaps one of @clason, @languitar or @dbmrq would be willing to check out the branch and test it.

If you test, please note the change of option names: All `g:vimtex_index_...` options are now `g:vimtex_toc_...`.

Perhaps I should add a deprecation warning for this in the init code?

dbmrq:
Awesome! I’ll try it in the next couple of days and let you know. :)

clason:
Nice! From a brief test, toggling keys messes up the jumping (I assume that the hiding the keys doesn't recalculate the correspondence between line number in the list and the target).

Also, I would be very sad if the old shortcuts for list of labels and list of contents were to disappear -- for longer documents such as lecture notes or books, a combined list is too long to be useful, and jumping to a section and jumping to a label are (for me) very different use cases (and so are based on a different mental model, so to speak; having the list match the model saves some time). So a big vote from me for prefiltered shortcuts (including list of todos, which is quite useful)!

(I can live without a table of figures/tables, although using the table of labels for that purpose as well presupposes coauthors that stick to a convention for labels, which is not always the case...)

lervag:
Thanks for the reviews! I've updated the todo with allowing maps to prefiltered tocs.

> Toggling keys messes up the jumping

I do not quite understand; could you explain what you mean?

> I can live without a table of figures/tables, although using the table of
labels for that purpose as well presupposes coauthors that stick to
a convention for labels, which is not always the case...

I did not understand this either. What would a table of figures/tables give that is not provided by the table of labels?

clason:
> I do not quite understand; could you explain what you mean?

Sorry, I should have been more precise. I'll try to come up with a minimal reproducing example (the document I tested with was a large multi-file project), but the gist of it is if I open the ToC and then turn off, say, the contents layer and select one of the labels, vimtex jumps not to the label but to a different location (which *might* correspond to the entry that was in this position before filtering, but that's purely a guess).

> I did not understand this either. What would a table of figures/tables give that is not provided by the table of labels?

Well, not every float might have a label (in which case it wouldn't show up in the table of labels), or you might be working with coauthors who don't stick to the assumed convention for labels (e.g., `fig:X` for figures, `tab:Y` for tables) and so can't easily be found in a list of 100+ labels for a long document. So a separate layer for floats obtained from parsing the `aux` files would be of use, but it's definitely more of a niche case than for the other layers (and additional work, so it shouldn't be part of this PR).

Actually, the more I think about it, the less I can see myself ever using the combined list. Besides being often too long, if you have labels for chapters or sections, they would either show up in addition to the chapter and section itself, leading to duplicate entries or -- as now -- not show up at all, leading to missing labels, which is worse in my view. (If I open a list of labels, I expect to see *all* labels, and not what the labels point to.) On the other hand, the former would make sense if the additional contents layer should be interpreted as structuring the table of labels/todos to it.

So my request would be to have the label layer show *all* labels exactly as they appear in the TeX, while the content layer is interpreted as now (and doesn't strip duplicates). A separate feature request (not part of this PR) would be to print the number next to the label (the same way that the omnicomplete function already does) instead of the ToC counter.

And while I'm at it: Would it be possible to toggle the help text (with an option to default to off)? It now takes quite a bit of space (and most of it is cut off anyway -- might I suggest putting the key *before* the text and, if possible, omitting the `;;`?), so after getting used to it (which I expect to be in a short time), I'd prefer to use the space for content.

(And just to be clear since this is turning into a lot of kvetching: 💯 for the feature itself, I think this will be really useful!)

clason:
Yep, seems to work fine now -- thanks! (And sorry I missed the option.)

Yes, a single additional layer for floats (figures, tables, algorithms(?)) was exactly what I was thinking of.

And the proposed way of creating custom mappings sounds perfect.

lervag:
I'll add the layer for floats to the list of remaining work. One question: What should the text be in the ToC? My current idea is something like this:

1. Match `\begin{(figure|table)}` to detect a float
2. Search for `\label` and/or `\caption` to get the ToC string

I think I want to finish this PR first. I would be happy if you could open a new issue for the float layer.

My goal now is to finish the mapping feature tomorrow, which should make the PR ready for merge, unless you (or anyone else) find a bug or have suggestions for improvements.

lervag:
Ok, I've pushed an update that should allow custom mappings, see `:help vimtex-toc-custom-maps`. Note, you will need to "force pull"; e.g.

```
git fetch
git checkout better-toc
git reset --hard origin/better-toc
```

I think it is ready to merge, but I'll wait a couple of days in case I've missed a bug or something similar.

dbmrq:
I just gave it another try and everything seems to be working really great! 🍻

A small detail for the todo part:  I think it would be better if the keywords were only matched if they ended with a non-word character (`\W`) or something like that. Right now if I have a comment that says "todowhatever" it will show on the list as "todo: whatever". It's also not possible to use keywords that contain each other… I use both "important" and "importante" (important in Portuguese), for instance, and the last one shows on the list as "important: e".

That's just nitpicking though, this is already great. :)

clason:
Very nice!

Some things I noticed:
* The documentation still claims that `+/-` in/decreases `tocdepth`, but it only changes the width of the number column. (Being able to change `tocdepth` on the fly would be useful!)
* Adding new keywords `g:vimtex_toc_todo_keywords` doesn't seem to work fully for me; `\cctodo` isn't parsed even when setting `let g:vimtex_toc_todo_keywords = ['todo', 'cctodo']`
* This is arguable, but the list of labels is missing the section labels. I know that the list of contents does something similar, but since they're parsed, it's not completely equivalent, and at the least this should be noted in the documentation.
* Again, this may be by design, but the input layer shows only `\input`, not `\include`d files. Actually, I think this layer might be even more useful if it also included `\includegraphics`.
* For completeness (let me know if I should open a new issue for this or if you rather not spend time on it),  the label layer could show instead of the containing section numbers the actual number the label refers to.

clason:
> I'll look into it, but I would say this belongs to a separate issue. If I see something "easy" to fix, then I'll update it now. Just to be clear: If +/- would change tocdepth, it would also change which entries to display, right? I.e., a \subsubsubsection could be hidden if the tocdepth is to small?

Yes, that is how I understood the documentation. (To be clear, I was just pointing out that the documentation doesn't match the implementation; the easy fix is of course to change the documentation.)

> Currently, this option only relates to todos in comments, i.e. % TODO: Text, not to the todopackage integration. This will require more work. Perhaps the best approach is to match any command that matches \\\w*todo\w*\>?

That's weird, because it currently definitely matches `\todo` (and from your comments on another issue, I thought this was intended?) 

> I would prefer to work on this after the current PR is finished, so please open
a new issue. Thanks!

Of course!

lervag:
> That's weird, because it currently definitely matches \todo (and from your comments on another issue, I thought this was intended?) It might be easier to make todo_keywords take full regex rather than just keywords?

I think the confusion here is that I have two parsers for TODOs: One parser that parses comments, and one that parses the `\todo` from the todonotes package. The former was updated to allow more keywords, the latter has not been updated and will only match `\todo{`. I think it does make sense to match something like `\cctodo` and similar variants as well, since it seems like a common use of the todonotes package. I'll do this by updating the regexp from `\\todo` to `\\\w*todo\w*`.

clason:
Ah! That explains it, of course. The updated regex is fine with me; if somebody localizes their `todonotes` commands, they can speak up for themselves (@dbmrq ?) 😃

lervag:
@clason You wrote:

> This is arguable, but the list of labels is missing the section labels. I know that the list of contents does something similar, but since they're parsed, it's not completely equivalent, and at the least this should be noted in the documentation.

I don't see exactly what you mean with "list of labels is missing the section labels". For me, the list of labels show all lines with `\label{...}`.

Also note, in latest version I think I've fixed tocdepth and the todo issue.

clason:
What I meant was that the list of labels doesn't show the section *labels* (it does show the headings, but only if the content layer is turned on).

And yes, `tocdepth` (and `\cctodo`... so many todos... 😞 ) works fine now -- thanks!

lervag:
I've once again forced pushed. I think this should be ready for merge, but again I'll allow a day or so for comments.

@clason On my side, your example seems to indicate a bug. The line `\section{one}\label{sec:one}` only shows a single line in the ToC. I think the reason is due to a limitation in the parser, in that it parses everything line by line, and it only allows a single match per line. So, in this case the label is matched first, and the section is thus not matched.

Personally, I think it is better to write the labels on the next line. Do you have any good reason why that is not good?

languitar:
Sorry, I didn't have time to test this so far.

> Personally, I think it is better to write the labels on the next line. Do you have any good reason why that is not good?

There is a chktex warning that at least recommend to put caption and labels on the same line to ensure consistent hyperlinks. There's a package (hypcap) that relaxes this constraint, but without that, in some cases, this seems to be necessary.

lervag:
> There is a chktex warning that at least recommend to put caption and labels on the same line to ensure consistent hyperlinks. There's a package (hypcap) that relaxes this constraint, but without that, in some cases, this seems to be necessary.

I see. I've never experienced an issue with this, and I have worked with LaTeX documents for 15+ years. I don't even add a comment after the section command, although I would expect that to fix any unlikely issues.

An important point from my side is that it might be a difficult job to improve the vimtex ToC parser to allow multiple ToC entries per line. So, I propose this: If you think this is an important feature, please open a new issue for it. In the meantime I will merge the current PR.

clason:
@lervag Thank you! I see the difficulty with the parser, and I don't feel strongly enough to argue for more work on your part here. (Although as @languitar mentioned, there are also recommendations for putting the labels on the same line; I myself are not consistent here.)

I feel really bad about bringing this up, but would it be possible extend the mappings to allow for specifying further options? Specifically, I would like to open the table of labels with `tocdepth` set to zero, because only then I get the old coloring depending on label prefix and the shown numbers are weirdly inconsistent (some show the part, some the chapter, some the section).

clason:
Playing around with the `tocdepth`, I now keep getting errors when opening the ToC with only the label layer selected:
```
Error detected while processing function 73[34]..79[38]..80[15]..83[7]..84[3]..85:
line   29:
E684: list index out of range: 0
E116: Invalid arguments for function nr2char
E15: Invalid expression: nr2char(number[0] + 64)
```
It's a bit random; I'll see if I can find a consistent way of reproducing it.

clason:
>  Specifically, I would like to open the table of labels with tocdepth set to zero

Actually, I shouldn't mess with `tocdepth` but just toggle the numbering (`s`). So what I want is to configure the mappings so that numbering is on *only* if the content layer is shown.

lervag:
@clason Thank you for more constructive feedback! It is appreciated! However, to keep things short I won't include your quotes and I'll try to be brief.

1. I'll look into more customizable mappings.

2. There is actually not a dedicated parser for `\input` and `\include`; these are handled specially due to the way the vimtex parser for TeX files work. However, I agree we should also include `\input{file.tikz}`, and so I'll add it.

3. Not quite sure about relative vs absolute; for me, it already shows relative paths, I think. Could you create a minimal example?

4. Did you manage to reproduce the error with the tocdepth thing?

5. The other layers already ignore the `tocdepth`.

clason:
No problem!

Re 3: I see the issue now: the paths are relative to the current working directory, not the current (or main) file. So if I start, say, `gVim` from the a directory and open `:e` a file in a subdirectory, the include layer shows `subdirectory/files`. Not sure it merits spending a lot of time on it, then.

Re 4: Somewhat: Open a toc showing only(!) the label layer and press `-` or `+` a couple of times until an error message happens. This will persist (even after vim restarts!) unless you open a contents (or include?) layer and adapt tocdepth there first. (The persistence makes it a bit hard to get a clean set of steps for reproducing.)

lervag:
> the paths are relative to the current working directory, not the current (or main) file

Ah, yes. This is natural to me, because I like to keep the current path to the current file's path. But it should be set to something that feels natural to most people... what would that be?

Btw, I think I've fixed the bug with the input file for tikz files. Please check.

clason:
Thanks! The tikz files now show up. From some brief tests, it also seems to help with the errors? 

About the file path: I agree that relative to the current file path is the most natural; the issue is that if you launch a GUI vim (gVim, vimR) from a launcher and *then* open a file, the current path is always the home directory, no matter what file you open later. But as I said, a minor thing.

I think if you can link `s` and `C` (i.e., toggling contents on/off automatically toggles numbers, and the default is off), I'm very happy with the current state.

lervag:
> About the file path: I agree that relative to the current file path is the most natural; the issue is that if you launch a GUI vim (gVim, vimR) from a launcher and then open a file, the current path is always the home directory, no matter what file you open later. But as I said, a minor thing.

This depends on the `autochdir` setting, which is off by default. I'm a weirdo who prefers to keep it on. :)

lervag:
Ok, I've pushed a new major update. I've made it possible to create full custom ToCs pretty easily. See the updated docs for an example, `:help vimtex-toc-custom-maps`.

lervag:
Due to the big update, I'll once again allow some days before merge. But now I think I've met most of the issues that you've raised, and I am getting pretty happy about the current implementation.

I might have missed something, so please let me know what you think!

clason:
Wow, thanks! So if I want to have separate mappings for each layer, I'd put
```vim
let g:toc_label = vimtex#toc#new({
         \ 'layers' : ['label'],
         \ 'show_numbers' : 0,
         \})
let g:toc_contents = vimtex#toc#new({
         \ 'layers' : ['contents'],
         \ 'show_numbers' : 1,
         \})
```
etc. in my `.vimrc`? That somehow doesn't work -- I get an error message when starting nvim, and calling `vimtex#toc#new(...)` manually does nothing.

lervag:
@clason Not quite. You should ensure that the `vimtex#toc#new` is run after the TeX buffer has been loaded, as indicated in the help with the `autocmd`, i.e.:

```vim
augroup vimtex_customization
  autocmd!
  autocmd FileType tex call CreateTocs()
augroup END

function CreateTocs()
  let g:toc_label = vimtex#toc#new({
           \ 'layers' : ['label'],
           \ 'show_numbers' : 0,
           \})
  let g:toc_contents = vimtex#toc#new({
           \ 'layers' : ['contents'],
           \ 'show_numbers' : 1,
           \})

  nnoremap <silent> ... :call g:toc_label.open()<cr>
  nnoremap <silent> ... :call g:toc_content.open()<cr>
endfunction
```

Another possibility is to add the content of `CreateTocs()` to a `ftplugin/tex.vim` file (I personally prefer to keep small `ftplugin/<filetype>.vim` for filetype specific customizations, see [my `.vim/personal/ftplugin` files](https://github.com/lervag/dotvim/tree/master/personal/ftplugin).

clason:
Yes, I see. It's a bit more cumbersome than before (I personally prefer to keep all customization in my vimrc for easy synchronization across my machines), but it works. (Although I'm a bit worried that this will impact startup time, but probably unreasonably so.) 

I noticed that selecting entries from the `include` layer jumps *into* TeX and TikZ (and, less usefully, PNG) files instead of the location where they are included -- except for bib files, for which nothing happens?

clason:
I really hate to write this, but the performance of the new ToC is significantly worse than before. It's not so noticeable for a paper with a single TeX file and a few includes, but for a book with more than a dozen included TeX chapters, jumping to a new location takes several(!) seconds where it used to be less than a second. (Opening a ToC is also slow, but it wasn't much faster before.)

dbmrq:
I also noticed that the ToC is significantly slower, even when just opening it for a larger project. I think the new features are worth it, so I wasn't going to say anything, but since @clason mentioned it I thought it could be useful to add that this is my experience too.

clason:
@lervag I know you want to be done with this, so might I suggest just leaving in the legacy toc and label lists (for now) and adding the new "smart" list as a separate mapping? This way, those who want performance (or hate having their cheese moved) can stick to the old and those who want the features can use the new. (And it's clear that new features -- if any -- are only implemented for the new list.)

lervag:
I have not noticed this slowdown. Have you found an easy way to verify it? For me, it is still O(100ms) to open...

There is really nothing in the recent refactoring that should make things slower, so I'll fix this before I merge.

clason:
It seems to be a direct function of the number of included files and/or labels, so it's not easy to reproduce except with an actual project of that length. Jumping is also only really slow if the target is in a different file.

clason:
Something I also noticed: The toc parser parses `\include`d files even if the include statement is commented out.

clason:
> Less easy (and going out on a limb here), it might help to refactor the l.pat expression to do hierarchical matching (i.e., don't match the full expression at once, but first check whether it's a comment, and if not, check whether it's a command, and only then match the regex).

Indeed, the following change in `state.vim` helps **significantly** with opening files (but not so much the toc):
```vim
function! s:vimtex.parse_graphicspath() abort dict " {{{1
  let self.graphicspath = []

  let l:pat = '\v\\graphicspath\s*\{\s*\{\s*\zs.*\ze\s*\}\s*\}\s*$'

  for l:line in self.preamble
      let l:paths = matchstr(l:line, g:vimtex#re#not_comment)
      if !empty(l:paths)
          let l:paths = matchstr(l:paths, g:vimtex#re#not_bslash)
          if !empty(l:paths)
              let l:paths = matchstr(l:paths, l:pat)
              if !empty(l:paths)
                  let l:paths = split(l:paths, '\s*}\s*{\s*')
                  let self.graphicspath = l:paths
                  break
              endif
          endif
      endif
  endfor
endfunction
```

lervag:
Thanks for the profiling, @clason! I've updated something close to suggested; could you please see if it is better now?

Note: I'm not able to reproduce the issue with slow speed, neither for opening nor for activating. This makes it difficult to work on improvements, unfortunately.

If you by chance are able to share a project where this happens, and your relevant config, then I could investigate it further? Another alternative is if you could find some publicly available documents where the same issue occurs.

lervag:
I don't see what I should do with the file `1802.03347`. I can't gunzip it, I can't untar it, etc. I guess I'm just stupid, but...

For your suggested changes: The first one may be good, but the second one is wrong. The matchers may match commented lines, e.g. the todo matcher (i.e. `% TODO: ...`).

clason:
>I don't see what I should do with the file 1802.03347. I can't gunzip it, I can't untar it, etc. I guess I'm just stupid, but...

No, arXiv is stupid... `tar xzf 1802.03347` works for me.

> For your suggested changes: The first one may be good, but the second one is wrong. The matchers may match commented lines, e.g. the todo matcher (i.e. % TODO: ...).

I realized that... In any case, either of these just worked for contents; commented out labels and includes still appeared in the tables. (The `re.vim` change seems to have prevented parsing commented out includes, though.)

lervag:
`tar xzf 1802.03347` does not work for me. Not for the other file either. :(

But OK, in the meantime, I just want to see if I understand things correctly. The following four issues are the ones I'm aware off that are active now:

1. Activating a TOC entry now works as expected (i.e. fast). If slow, then it is similarly slow if opening the corresponding file separately, so the slowness is due to vimtex being slow to load. This would be a separate issue.

2. Opening the ToC may take some time (O(1s)) for large documents.

3. Commented out labels and includes are not ignored.

4. Include entries go to the included file, not to the include line.

Please let me know if I've missed any comments!

-----

1. If this is still an issue that you think could or should be improved, then please open a new issue for this and provide an example of a file that you feel should open faster.

2. This is of course computer dependent. I'm on a new Carbon X1 Gen. 6, which is pretty fast. So that might explain why I don't see this issue. I've tested on my PhD thesis, which has 3400 lines of LaTeX code (only as a single file), and it has a noticable lag, perhaps 1 second. After updating to fix e.g. 3 above, it is slower than before. Not unexpected, since the regexp had to be tighter.

    Note: One could possibly change the `refresh_always` option to 0 (perhaps also by default). Then the `r` key for refreshing would be used to refresh the ToC, but now it will open much faster on consequent calls.

3. Should be fixed now. But at the expense of even slower ToC open. Sorry...

4. Yes, this is a matter of taste, I guess. But perhaps it is better to go to the include line, then do `gf` if one wants to enter the file? Still, it feels like the point of having these entries is to be able to open the files...

lervag:
One possiblity for speedup is to add a couple of restrictions to the author:

1. TODO comments should come on separate lines, i.e. match `^\s*%'.

2. `\todo`, `\include`, `\bibinput` and similar commands should be attached to beginning of line.

If 1. and 2. could be assumed, then the parsing would be much lighter and it would be faster.

clason:
Weird. Did you try renaming it? But no matter, the files aren't really important.

1. Correct. And the slowness -- as far as it is slower than before -- is due to parsing. So if 2. gets taken care of, this should be be fine, too.

2. Correct. And the slowness is due to matching regexes in two places. (The machine I'm testing on is a 4.2 Ghz Core i7, so not exactly a slouch, either.) The project I'm harping on about has 15k lines of TeX in 26 files, but I think it's more the number of entries than the number of lines. I'll play with `refresh_always`; maybe that's a reasonable workaround.

3. Correct first, fast second 😄 

4. I'm coming around to your point of view: Opening the included files is more useful, but then there's no need to match graphics files (pdf, png, jpg, etc.). But bib files should work 😉 

5. I think assuming `\include` and `\bibinput`/`\addbibresource` being a single line is reasonable, but `TODO` and `\todo` can be anywhere (since for the latter, wrong placement can insert spurious white space -- just like `\footnote`).

clason:
I just tested the changes; the speed matches what I observed in my own changes (~5s to open ToCs, but selecting entries is reasonably fast -- fast enough for me).

But now a lot of labels don't show up? (I think only those on a line by themselves appear, ones like
```tex
\begin{equation}\label{eq:whatever}
```
don't.)

dbmrq:
I just tried searching for my todos with vimgrep and it was pretty slow too, so I guess that's not Vimtex's fault.

I don't love the idea of assuming todo's will be on their own lines, but I also tried that with vimgrep and it really was a lot faster, so it's definitely tempting.

Another option would be to do what @clason already suggested and let users define their own regex patterns for the todos, instead of just keywords. Then people who often work on large projects could make that choice themselves.

Also, I just tried the latest version with this setting:

   let g:vimtex_toc_todo_keywords = ['TODO', 'FIXME', 'IMPORTANT', 'IMPORTANTE']

And `% IMPORTANTE: …` still shows as "IMPORTANT: E: …" for me. :/

lervag:
I've updated according to the latest comments. There are two remaining issues: _speed_ and _include for bib files_.

> I'm coming around to your point of view: Opening the included files is more useful, but then there's no need to match graphics files (pdf, png, jpg, etc.). **But bib files should work wink**

For me, bib files work as expected. Could you provide a minimal example that fails?

I'm working on the speed now. I thought of an optimization before falling asleep last night that might help.

clason:
> For me, bib files work as expected. Could you provide a minimal example that fails?

It fails if multiple bib files are included in a single command:
```tex
\bibliography{bib1,bib2}
```
(it also only shows `bib2` in the table). Not sure that's fully legal, though (although it works), so I wouldn't worry about it that much. Might be worth documenting, anyway, in case someone else stumbles over this.

lervag:
Right, so, I've tweaked and squeezed as much as I can. I acknowledge that it might still be slow for very large files. But it should not be too bad now, I think.

I'll also look into the bibliography issue. If I'm right, that's the last thing that remains. It should not be too difficult. When solved, then I propose that this is ready for merging. As always, feel free to raise concerns! :)

clason:
That's impressive -- from my (admittedly brief) tests, I don't see much difference from the old toc (opening files *and* tocs) anymore!

I think requiring each bibliography to be a single statement is reasonable -- but I'd suggest also matching biblatex's `\addbibresource` for that (which takes the full filename including the `.bib` extension, by the way).

clason:
Two things I noticed: 

1. The todo matcher doesn't match `\todo[inline]{fix me}`.

2. Something seems to be screwing up the numbering in the contents layer -- I think it's now ignoring all sectioning *with* labels?

lervag:
`\todo[inline]{fix me}` works for me.

Reg. the numbering: Are you still putting the `\label` on the same line as the `\section` command?

lervag:
Now the todo matcher should work as expected, e.g. for

```tex
\todo[inline]{
  ...
  }
```

Further, we parse all listed bibs in the `\bibliography` commands, the `\section` commands are parsed before the `\labels`, and we can have both of them on the same line.

I've tried to do some more minor optimizations, but I don't think they've made any difference.

clason:
Thank you! I had already resigned myself to go through all my TeX files and adapt the style...

As far as I can tell, that only leaves

1. parsing `\addbibresource` for bib includes and
2. the list of todos cuts of the last character of the argument of `\*todo{...}` (i.e., would show only `..`).

lervag:
What is wrong with the `\addbibresource` command? It should work.

I'll fix the cut off for the TODOs immediately.

clason:
Well, it doesn't show up in the list of includes:
```tex
\addbibresource{quasilinearelliptic.bib}
```
but there's no bib file in the table of includes.

clason:
Thanks, `\addbibresource` works perfectly now!

Regarding the other commands: I have not used them yet (as far as I remember), but I think they go into the preamble as well. "Local" resources are provided via
```tex
\begin{refsection}[section.bib]
...
\end{refsection}
```
But I think this is something of a niche use (if different -- normally you can just use the standard `\addbibresource` for `refsection`s as well. I wouldn't worry about it for now and wait for someone to actually need that.

clason:
> So, it seems we might be finished, then. 

Indeed (phew!) :smiley: Impressive work!

I'll test-drive the branch for my daily work the next two days, and promise I will only report major bugs.

lervag:
Thanks!

I'm happy that I did this. I think I'll even use this myself (I've never used the old ToC for anything). :)

lervag:
Ok, so I'm merging this now. It seems to work well. As always, feel free to open issues if something unexpected breaks.

PR #1201, opened 2018-08-30
Add message when lacking documentation, add redraw

andymass:
Forgot I had these small tweaks in my repository for a while:

  - with master, when I go to view documentation and come back to vim, I have to redraw the screen due to the use of `silent execute ! ... &`
  - it's nice to have feedback that a command is not found when pressing `K`.


lervag:
Thanks!

Issue #1200, opened 2018-08-24
RFC: Support external tools for formatting code e.g. latexindent.pl


kiryph:
There are several suggestions for formatting latex source code or more generally 

> plaintext [...] which exports to some non-plaintext system. 

**Update**  I mainly have in mind a manuscript written in latex, not the source code of a latex package, also not specifically tikz or pstricks pictures.

The formatting style should make collaborative writing easier,  being VCS-friendly (ideally commits should be limited to actual changes, not unnecessary re-flows of paragraphs), diff friendly (cli with limited line lengths) and of course should be easy to read.

## Suggestions for formatting

- [Wikibook: LaTeX/Collaborative Writing of LaTeX Documents](https://en.wikibooks.org/wiki/LaTeX/Collaborative_Writing_of_LaTeX_Documents)
  > ### Directives for using LaTeX with version control systems
  > 
  > Avoid 'ineffective' modifications.
  > Do not change line breaks without good reason.
  > Turn off automatic line wrapping of your LaTeX editor.
  > Start each new sentence in a new line.
  > Split long sentences into several lines so that each line has at most about 80 characters.
  > [...]

- Article: [VCS friendly, patchable, document line wrapping by Cristopher Lemmer](http://dustycloud.org/blog/vcs-friendly-patchable-document-line-wrapping/) (add additional indentation for split sentences)
  > ```
  > If you do enough work in any sort of free software environment, you get used
  >   to doing lots of writing of documentation or all sorts of other things in
  >   some plaintext system which exports to some non-plaintext system.
  > One way or another you have to decide: are you going to wrap your lines with
  >   newlines?
  > And of course the answer should be "yes" because lines that trail all the way
  >   off the edge of your terminal is a sin against the plaintext gods, who are
  >   deceptively mighty, and whose wrath is to be feared (and blessings to be
  >   embraced).
  > So okay, of course one line per paragraph is off the table.
  > So what do you do?
  > ```

- Article: [Semantic linefeeds by Brandon Rhodes](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) with quote from [1974 by Brian Kernighan](https://en.wikipedia.org/wiki/Brian_Kernighan) 

  > ### Hints for Preparing Documents
  > 
  > Most documents go through several versions (always more than you expected) before they are finally finished. Accordingly, you should do whatever possible to make the job of changing them easy.
  > 
  > First, when you do the purely mechanical operations of typing, type so subsequent editing will be easy.
[...]

PR #1204, opened 2018-09-02
Better amp indent (fixes #1196)

### LINKED ISSUE #1196: bad indent with interaction of command{ + alignment char
I think it has been this way for some time but I didn't investigate the precise conditions before.  It seems like the indent following command-open-brace is computed incorrectly when there are ampersands.  When using `let g:vimtex_indent_on_ampersands=0` brace indent does work (except of course ampersands aren't aligned then).

```vim
set nocompatible
let &rtp  = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'
filetype plugin indent on
syntax enable
```

MWE, and behavior after re-indenting (`gg=G`):
```tex
\begin{document}
\begin{environ*}
	A &= 10
	\\&= \command{
	something
}
\end{environ*}
\end{document}
``` 

Expected:
```tex
\begin{document}
\begin{environ*}
	A &= 10
	\\&= \command{
	        something
        }
\end{environ*}
\end{document}
```

lervag:
This should solve #1196. I think it generally improves the indendation code as well. It seems to work well here. Please let me know what you think, @andymass.

### DIFF
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -29,21 +29,13 @@ endfunction
 function! VimtexIndent(lnum) abort " {{{1
   let s:sw = exists('*shiftwidth') ? shiftwidth() : &shiftwidth
 
-  let l:prev_lnum = s:get_prev_line(prevnonblank(a:lnum - 1), 0)
+  let [l:prev_lnum, l:prev_line] = s:get_prev_lnum(prevnonblank(a:lnum - 1))
   if l:prev_lnum == 0 | return indent(a:lnum) | endif
-
-  " Get current and previous line and remove comments
-  let l:line = substitute(getline(a:lnum), '\\\@<!%.*', '', '')
-  let l:prev_line = substitute(getline(l:prev_lnum),   '\\\@<!%.*', '', '')
+  let l:line = s:clean_line(getline(a:lnum))
 
   " Check for verbatim modes
   if s:is_verbatim(l:line, a:lnum)
     return empty(l:line) ? indent(l:prev_lnum) : indent(a:lnum)
-  endif
-
-  " Align on ampersands
-  if s:indent_amps.check(a:lnum, l:line, l:prev_line)
-    return s:indent_amps.indent
   endif
 
   " Use previous indentation for comments
@@ -51,36 +43,43 @@ function! VimtexIndent(lnum) abort " {{{1
     return indent(a:lnum)
   endif
 
-  " Ensure previous line does not start with ampersand
-  let l:prev_lnum = s:get_prev_line(l:prev_lnum,
-        \ get(g:, 'vimtex_indent_on_ampersands', 1))
-  if l:prev_lnum == 0 | return 0 | endif
-  let l:prev_line = substitute(getline(l:prev_lnum), '\\\@<!%.*', '', '')
+  " Align on ampersands
+  let [l:ind, l:do_indent, l:prev_lnum, l:prev_line]
+        \ = s:indent_amps.check(a:lnum, l:line, l:prev_lnum, l:prev_line)
+  if l:do_indent
+    return l:ind
+  endif
+  " unsilent echo printf('%2s %2s %2s', l:ind, l:prev_lnum, a:lnum) l:line "\n"
 
   " Indent environments, delimiters, and tikz
-  let l:ind = indent(l:prev_lnum)
   let l:ind += s:indent_envs(l:line, l:prev_line)
   let l:ind += s:indent_delims(l:line, a:lnum, l:prev_line, l:prev_lnum)
   let l:ind += s:indent_conditionals(l:line, a:lnum, l:prev_line, l:prev_lnum)
   let l:ind += s:indent_tikz(l:prev_lnum, l:prev_line)
+
   return l:ind
 endfunction
 
 "}}}
 
-function! s:get_prev_line(lnum, skip_amps) abort " {{{1
+function! s:get_prev_lnum(lnum) abort " {{{1
   let l:lnum = a:lnum
-  let l:prev = getline(l:lnum)
+  let l:line = getline(l:lnum)
 
-  while l:lnum != 0
-        \ && (l:prev =~# '^\s*%'
-        \     || s:is_verbatim(l:prev, l:lnum)
-        \     || (a:skip_amps && l:prev =~# s:indent_amps.re_align))
+  while l:lnum != 0 && (l:line =~# '^\s*%' || s:is_verbatim(l:line, l:lnum))
     let l:lnum = prevnonblank(l:lnum - 1)
-    let l:prev = getline(l:lnum)
+    let l:line = getline(l:lnum)
   endwhile
 
-  return l:lnum
+  return [
+        \ l:lnum,
+        \ l:lnum > 0 ? s:clean_line(l:line) : '',
+        \]
+endfunction
+
+" }}}1
+function! s:clean_line(line) abort " {{{1
+  return substitute(a:line, '\s*\\\@<!%.*', '', '')
 endfunction
 
 " }}}1
@@ -93,22 +92,83 @@ endfunction
 
 let s:indent_amps = {}
 let s:indent_amps.indent = 0
+let s:indent_amps.env = 0
+let s:indent_amps.continued = 0
+let s:indent_amps.prev_lnum = 0
+let s:indent_amps.prev_line = ''
 let s:indent_amps.re_amp = g:vimtex#re#not_bslash . '\&'
 let s:indent_amps.re_align = '^[ \t\\]*' . s:indent_amps.re_amp
-function! s:indent_amps.check(lnum, cline, pline) abort dict " {{{1
-  if get(g:, 'vimtex_indent_on_ampersands', 1)
-        \ && a:cline =~# self.re_align
-        \ && a:pline =~# self.re_amp
-
-    let l:pre = strdisplaywidth(strpart(a:pline, 0, match(a:pline, self.re_amp)))
-    let l:cur = strdisplaywidth(strpart(a:cline, 0, match(a:cline, self.re_amp)))
-    let self.indent = max([indent(a:lnum) - l:cur + l:pre, 0])
-
-    return 1
+function! s:indent_amps.check(lnum, cline, plnum, pline) abort dict " {{{1
+  if !get(g:, 'vimtex_indent_on_ampersands', 1)
+    return [a:plnum > 0 ? indent(a:plnum) : 0, 0, a:plnum, a:pline]
   endif
 
-  let self.indent = 0
-  return 0
+  if !self.env
+    if a:pline =~# self.re_amp && a:cline =~# self.re_amp
+      let self.env = vimtex#env#is_inside('\w\+')[0]
+      let self.indent = strdisplaywidth(
+            \ strpart(a:pline, 0, match(a:pline, self.re_amp)))
+      let [self.prev_lnum, self.prev_line] = self.get_prev_nonamp(a:plnum)
+    endif
+  endif
+
+  " Check if the ampersand environment is closed
+  if self.env
+    let l:env = vimtex#env#is_inside('\w\+')[0]
+    if l:env != self.env || a:cline =~# '^\s*\\end'
+      let self.env = ''
+      let self.indent = 0
+      let l:plnum = self.prev_lnum
+      let l:pline = self.prev_line
+      let self.prev_lnum = 0
+      let self.prev_line = ''
+      return [l:plnum > 0 ? indent(l:plnum) : 0, 0, l:plnum, l:pline]
+    endif
+  endif
+
+  " Check if we should indent directly
+  if self.env && a:cline =~# self.re_align
+    let l:ind_diff =
+          \   strdisplaywidth(strpart(a:cline, 0, match(a:cline, self.re_amp)))
+          \ - strdisplaywidth(strpart(a:cline, 0, match(a:cline, '\S')))
+    return [self.indent - l:ind_diff, 1, a:plnum, a:pline]
+  endif
+
+  " Get indent (either continued or from previous nonamped line
+  if self.get_continued_indent(a:cline)
+    let l:ind = self.indent + s:sw
+    let l:plnum = a:plnum
+    let l:pline = a:pline
+  else
+    let [l:plnum, l:pline] = self.get_prev_nonamp(a:plnum)
+    let l:ind = l:plnum > 0 ? indent(l:plnum) : 0
+  endif
+
+  return [l:ind, 0, l:plnum, l:pline]
+endfunction
+
+" }}}1
+function! s:indent_amps.get_continued_indent(line) abort dict " {{{1
+  if self.env && !self.continued && a:line !~# self.re_amp
+    let self.continued = 1
+    return 1
+  else
+    let self.continued = 0
+  endif
+endfunction
+
+" }}}1
+function! s:indent_amps.get_prev_nonamp(lnum) abort dict " {{{1
+  let [l:lnum, l:line] = s:get_prev_lnum(a:lnum)
+
+  while l:lnum > 1
+    if l:line !~# s:indent_amps.re_align
+      break
+    endif
+    let [l:lnum, l:line] = s:get_prev_lnum(l:lnum-1)
+  endwhile
+
+  return [l:lnum, l:line]
 endfunction
 
 " }}}1
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -100,8 +100,7 @@ function! s:indent_amps.check(lnum, cline, plnum, pline) abort dict " {{{1
   let self.prev_ind = a:plnum > 0 ? indent(a:plnum) : 0
   if !get(g:, 'vimtex_indent_on_ampersands', 1) | return self.prev_ind | endif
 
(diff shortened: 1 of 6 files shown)


andymass:
Thanks!  I really appreciate the work you put in to improve the indentation- I know this is a very tricky area of the plugin.  Regarding the `\\ &=` form that I use (which worked prior to this PR), I can live with it and I would be happy to address this myself at some point if you are willing to accept changes.

I haven't had any time to dedicate to testing this PR but I am actively using it in my writing.
How do you feel about this case?
```tex
\[
    \somecommand{\someother{
            hello  % too much indent?
    }}
\]
```

lervag:
> Regarding the \\ &= form that I use (which worked prior to this PR), I can live with it and I would be happy to address this myself at some point if you are willing to accept changes.

I would definitely be willing to accept changes, as long as they make sense. :)

> How do you feel about this case?

The current implementation uses "simple" math to decide the indents, i.e. it counts the number of "openers" (e.g. in your example, there are two opening braces `{`). This gives double indentation in your example, but it also ensures that things are consistent. If I were to change this so that it could recognize that there should be only one indent, then I would need to make a more complex state machine. There are a lot of edge cases to consider, for instance:

```tex
% Case 1: This seems "easy"
  \begin{env1}
    \cmd1{\cmd2{
      Text
    }}
  \end{env1}

% Case 2a: Not so easy
  \begin{env1}
    \cmd1{\cmd2{
      Text
    }
  }
\end{env1}

% Case 2b: Not so easy, possible solution
  \begin{env1}
    \cmd1{\cmd2{
      Text
    }
    }
  \end{env1}
```

I think the example should indicate what is difficult about what you are asking. In addition, the state machine requires knowledge of the opening part. Say you are indenting from inside an already indented part, e.g.

```tex
    Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy
    eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam
    voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
    clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit
    amet.
    % How many times to deindent?
  }}
```

The current implementation does not have these problems, and as such, it is consistent (afaik).

lervag:
Btw, I do not quite understand your `\\ &` syntax. Do you think we should "ignore" the leading `\\` when indenting? It would be helpful with a more complex example for this. Is this what you want?

```tex
% Assuming shiftwidth=2 (and no tabs)
\begin{environ*}
  f(x) &= 10
     \\&= \command{
           something
         }
     \\&= g(x)
\end{environ*}
```

Also, I don't really think the PR should break anything wrt. this syntax, but
it should also not quite work before this PR. For me, the amp indent is
supposed to make things like this work:

```tex
% Assuming shiftwidth=2 (and no tabs)
\begin{tabular}
  column1 & column2
          & column3 \\
  column1 & column2
          & column3
\end{tabular}
```

andymass:
Thanks for the detailed explanation!   I can definitely live with the restriction that `{{` creates a double indent by avoiding introducing two commands on one line.

> Do you think we should "ignore" the leading \\\\ [...] Is this what you want?

Yes, precisely.  I think `\\` should be the same as white space when before ampersand.

Here are some examples, which I hope can make things clearer.  Each snippet is after performing `gg=G` on the branch mentioned, `shiftwidth=2`.

### on vimtex master

Ignoring leading `\\` was introduced in #961.  I know it is not commonplace in the LaTeX world but I do this because I find it makes adding/deleting equations easier.  It worked very well for examples like this:

```tex
\begin{environ*}
  f(x) &= 10
     \\&= something
     \\&= something else
\end{environ*}
```

Note that this _never_ worked, and I'm totally okay with it, because it fails gracefully:

```tex
\begin{environ*}
  f(x) &= 10
  + something
  \\&= something else
\end{environ*}
```

### on vimtex master

This example was the subject of #1196; it fails badly:
```tex
\begin{environ*}
  f(x) &= 10
     \\&= \command{
  something
}
\\&= g(x)
\end{environ*}
```

### on better-amp-indent

Ignoring of `\\` broke:

```tex
\begin{environ*}
  f(x) &= 10
       \\&= something
       \\&= something else
\end{environ*}
```

### on better-amp-indent

Although this fixed the issue of the closing brace going to column zero, it causes the ampersand to be placed slightly too much to the right, due to the `\\` .  I wonder if it's not a simple miscalculation of where the `&` is in the string.

```tex
\begin{environ*}
  f(x) &= 10
       \\&= \command{
           something
         }
       \\&= g(x)
\end{environ*}
```

lervag:
Ok, I think this should be fixed now. Please test.

andymass:
Sorry about this, but I found a weird issue.  For me, running the following causes the line `something` to oscillate between two indent states.  I can cause it on other lines as well.

Do `vim -u minvimrc indent.tex`.

minvimrc:
```vim
set nocompatible
let &rtp = '~/.vim/bundle/vimtex,' . &rtp
let &rtp .= ',~/.vim/bundle/vimtex/after'
filetype plugin indent on
syntax enable

set sw=2 ts=2

au VimEnter * normal! 4ggk==j
call timer_start(500, {tid->execute('normal! ==')}, {'repeat': -1})
nnoremap q :qall!<cr>
```
indent.tex:
```tex
\begin{environ*}
  f(x) &= 10
     \\&= \command{
           something
         }
\\&= g(x)
\end{environ*}
```

lervag:
Ah, yes. This is a result of the current implementation being state dependent. I.e., the result of `==` depends on the result of doing `==` on the previous line.

Again, this is a hard problem. Solving it by using a state worked well and was not too hard, but it has this exact limitation. If I were to solve this issue, then I would have to redesign the method to a completely new logic. That is, I would have to detect if the indentation of the current line depends on ampersands, either explicitly (if there is amps in the beginning) or implicitly (if it is a continuation line). If so, we would have to search backwards to determine the correct indentation level.

With state, I can set the correct level when first identifying the amps, and then use that for preceding lines. This makes it easy to continue applying the correct indentation level even if there are multiple continued lines in the middle.

So, the question is: Can we live with the current method, or should I spend more time to make a more robust stateless method?

languitar:
I just found this issue and i have another option: currently, vimtex has the concept that some tasks are better handled by other plugins. And for indentation or code formatting, one could also go that way. I am frequently using https://github.com/sbdchd/neoformat, which works pretty well for LaTeX with `latexindent` as the backend.

lervag:
Thanks, @languitar. See also #1200. I think this is a useful idea and will probably work in this direction in the future. But I still think the current PR will improve the current indentation method.

I also think it makes sense that vimtex provides a sensible indentation method by itself, even though it may be better to rely on a more dedicated method for an improved version.

lervag:
Yes, I think your basic problem here is due to the "state machine" implementation, and as such, I think your examples and description qualifies for a winning argument that I should go for the stateless approach.

Note that this will take some more time, because I will need to redo the logic. I will need to change from saving state during indents, to looking further backwards when the indent can not be fully specified from the previous line.

In this regard, it would be useful to add a restriction, that is: If I could say that the ampersand indent style is limited to certain environments, then it would be easy to first locate the current environment, and then use this information to set the current indent. For this, I would need a good list of environments where `&`s are allowed. I would appreciate some help in making this list. Obvious environments include `split` and `align\w*`, as well as `tabular`. (I would prefer to implement a good list instead of adding a new option.)

andymass:
Running this on the texstudio completion files gives a pretty big list.

    grep '\\begin{\w\+\*\?}.*array' *.cwl | sed -n 's/.*\\begin{\(\w*\*\?\)}.*/\1/p' | sort -u | tr '\n' '|'
    grep '\\begin{\w\+\*\?}.*tabular' *.cwl | sed -n 's/.*\\begin{\(\w*\*\?\)}.*/\1/p' | sort -u | tr '\n' '|'

```text
align|align*|alignat|alignat*|aligned|alignedat|bmatrix|bmatrix*|Bmatrix|Bmatrix*|bsmallmatrix|bsmallmatrix*|Bsmallmatrix|Bsmallmatrix*|cases|cases*|dcases|dcases*|drcases|drcases*|empheq|eqnarray|eqnarray*|flalign|flalign*|lgathered|matrix|matrix*|multlined|numcases|pmatrix|pmatrix*|psmallmatrix|psmallmatrix*|rcases|rcases*|rgathered|smallmatrix|smallmatrix*|split|spreadlines|subnumcases|tikzcd|vmatrix|vmatrix*|Vmatrix|Vmatrix*|vsmallmatrix|vsmallmatrix*|Vsmallmatrix|Vsmallmatrix*
```
```text
longtable|longtabu|mpsupertabular|mpsupertabular*|supertabular|supertabular*|tabu|tabular*|tabularx|tabulary|tikztimingtable
```

It's probably sufficient to drop the stars and go on name alone.  From a cursory look I noticed `tabular|Beqnarray|mpxtabular|xtabular` are missing the `tabular`/`align` flag so there are probably more which are unlabeled, but I think these should cover maybe 95% of cases.  

```text
align|alignat|aligned|alignedat|Beqnarray|bmatrix|Bmatrix|bsmallmatrix|Bsmallmatrix|cases|dcases|drcases|empheq|eqnarray|flalign|lgathered|longtable|longtabu|matrix|mpsupertabular|mpxtabular|multlined|numcases|pmatrix|psmallmatrix|rcases|rgathered|smallmatrix|split|spreadlines|subnumcases|supertabular|tabu|tabular|tabularx|tabulary|tikzcd|tikztimingtable|vmatrix|Vmatrix|vsmallmatrix|Vsmallmatrix|xtabular
```

lervag:
Ok, I've updated and rebased. Please test. I think this should be getting close to optimal.

andymass:
Very nice work, thanks! It seems almost perfect for all the cases we've mentioned.  There is one problem I noticed;

```tex
\begin{environ*}
  f(x) &= 10
       &= \begin{matrix}
         10
\end{matrix}
\end{environ*}
```

I think \begin ... \end should work basically the same as \command{ ... }.

This brings up an interesting question, what should happen if the inner environment has ampersands?  This is the behavior currently:

```tex
\begin{environ*}
  f(x) &= 10
       &= \begin{matrix}
  10 & 20 \\
  30 & 40
\end{matrix}
\end{environ*}
```

An ideal solution would probably align the ampersands to the inner environment separately from the outer one, although an acceptable solution might be to just not align the inner ones.  I don't know which is easier given the current implementation.

lervag:
Haha, of course you find the only weak point I know of... :)

Well, this is actually again not trivial. So, the current implementation uses the following assumption: All lines above the current line have correct indentation. I think this assumption makes sense. However, in the current implementation I only consider `\begin`s that are attached to the beginning of the line. This is why the first example fails.

I think the correct algorithm needs to somehow count `\begin...\end` pairs starting at the current line and back, in order to determine the correct nested block that determines the current ampersands.

If you agree this sounds correct, then I'll try to implement it during the weekend.

andymass:
> I think the correct algorithm needs to somehow count \begin...\end pairs starting at the current line and back, in order to determine the correct nested block that determines the current ampersands.
>
> If you agree this sounds correct, then I'll try to implement it during the weekend.

I see the complication.. I think this proposal sounds fine.  Hopefully it is not too difficult.  Would it not be enough to just look at whatever the current environment surrounding the line is?

I did notice some other problems btw:

1)
```tex
\[
  x &= 20
    &= 10
  \]

  \[
 x &= 10
   &= 20
 \]
```
2)
```tex
\[
    x &= 20
   \\ &= 10
\]

\[
   x &= 10           % why is this indented less?
  \\ &= 20
\]
```

lervag:
Before I consider your last mentioned problems: Is this even valid LaTeX? That is, can you use ampersands directly in the displayed math environment?

lervag:
Now I think it should work well also in displaymath and with nested environments. I've assumed that one always puts the `\begin{...}` as the last part of a line, and that one always puts the `\end{...}` at the beginning (and similar for `\[` and `\]`). I think this should not be very inconvenient for anyone.

andymass:
> Is this even valid LaTeX? That is, can you use ampersands directly in the displayed math environment?

No, you're absolutely right- it's invalid. 

These changes seem very good.  I will continue to try them out during my writing and let you know if there are any major issues.

lervag:
Sounds good. Actually, I think I'll merge this, because it seems to work fine for me for some time. Feel free to open new issues or continue this thread if you find other/more problems.

andymass:
Thank you!  I'm happy with the improvements and appreciate the effort.

clason:
I hate to be *that* guy, but this PR has introduced some regressions for me. First, it's noticeable slower than before (really noticeable with a 16kwords/2klines file, factor >10x), but that's probably unavoidable (can't please everyone, and I've already played my "please improve indentation performance card"...) but I also see some misalignments of

```tex
\begin{theorem}                                                                   
    \begin{equation*}                                                             
        \min_{x\in U} F(x)                           
\end{equation*}                                                                   
\end{theorem}   
```
(produced by indendation), which then leads to further misindentation of `\end{document}` (because no longer special cased?). 

(I did profile; most of selftime (10s) is spent in `s:indent_amps.parse_context` inside the while `l:lnum >= 1` loop -- which seems to be quadratic in the number of lines! -- and in particular the regexes, as well as `s:indent_amps.check` (also 10s), most of which in `call self.parse_context(...)`. If I ask really nicely, would it be possible to make the smarter `&` alignment -- and hence calling these expensive functions -- configurable? I'd be fine with disabling `&` alignment completely if that was the tradeoff.)

clason:
It's a work in progress with a co-author, so I'm not really comfortable sharing it at the moment. But here are the profile logs:
[master.log](https://github.com/lervag/vimtex/files/2463378/master.log)
[bb62ab259.log](https://github.com/lervag/vimtex/files/2463379/bb62ab259.log)
The interesting block is
```
653748              0.639251   while l:lnum >= 1
652886              1.501407     let l:line = getline(l:lnum)

652886              3.867070     if l:line =~# '^\s*\\\%(end\|\]\)'
125299              0.225606       let l:depth += 1                             125299              0.075438     endif
                                                                                652886              3.954085     if l:line =~# '\\\%(begin\S*\|\[\)$'
74505              0.144586       let l:depth -= 1
74505              0.106524       if l:depth == l:init_depth - 1                  977              0.001780         let self.init_lnum = l:lnum
  977              0.001631         let self.init_line = l:line
  977              0.002454         let self.init_ind = indent(l:lnum)
  977              0.000876         break
                                  endif
73528              0.030657     endif
651909              1.298461     if self.amp_ind < 0 && l:depth == 1 && l:line  =~# self.re_amp
  145              0.009739       let self.amp_ind = strdisplaywidth( strpart(l:  line, 0, match(l:line, self.re_amp)))
  145              0.000163     endif

651909              1.528101     let l:lnum = prevnonblank(l:lnum - 1)
651909              0.483660   endwhile
```
Note the number of calls; the function itself is only called 1839 times. So there might indeed be a bug.

The misalignment does seem related since it goes away when I comment out the new indent code. Getting a minimal example is difficult since the behavior is preserved when yanking and pasting to a new buffer, but *not* when copy-pasting the text via the clipboard. Very weird.

(Sorry I didn't notice that the option is already there! I was looking at the `VimtexIndent` function and assumed the whole ampersand indentation block there would be guarded by the option, not that the function itself would check it and terminate early.)

lervag:
I think the issue is that at least one (perhaps more) of the regexes do not handle end-of-line spaces. This should be easy to verify: If you do `:%s/\s\+$//` in your document, then, if I'm right, things should work as expected. Could you check if I'm right?

clason:
That fixes some but not all of the alignment issues, but doesn't impact the performance much. Curiously, I see the same problem if I select just this environment and selectively indent with `=`. Here's an example that indents correctly if (and only if) I delete the trailing spaces:
```tex
    \begin{equation*}
        \left \{
            \begin{aligned}
                -\Delta \theta &= u \quad \text{in } \Omega, \\
                \theta &=0 \quad \text{on } \Gamma.
            \end{aligned}
        \right.
    \end{equation*}
```
Here's one that doesn't even without trailing spaces:
```tex
    \begin{equation} \label{eq:state2}
        \left\{
            \begin{aligned}
                -\div [a_M(z)\nabla y + b(\nabla y)]  &= u \quad \text{in } \Omega, \\
                y &=0 \quad \text{on } \Gamma.
            \end{aligned}
        \right.
\end{equation}
```
(note in both cases the initial indentation, which is from a surrounding environment). If I remove the space between the `{equation}` and the `\label`, the indentation is correct. (If I disable indent on ampersands, the indentation is correct even with the space.)

lervag:
Ok, I see. I'm a bit swamped right now, so it would be very helpful if you could open a new issue for this, or perhaps 2:

1. For the misalignments (i.e. error in alignment), and

2. For the slow down.

This makes it easier for me to address the issues quickly, and as such, it is very helpful. I also think it makes it better wrt. involving other people (who can see the issues directly).

As said, I think I should be able to solve these issues when I get the time.

clason:
No problem! (Disabling amp indentation is a perfectly feasible workaround for me in the meantime.) 

I suspect both issues have the same cause (the stateless implementation misses `\begin`/`\end` of environments due to trailing spaces after the closing `}`, and this 1. throws off the count and 2. makes it look to far ahead/behind. But this just means you can close two issues with one commit :smile:

lervag:
Thanks! :)

PR #1226, opened 2018-09-21
Autoclose the QuickFix window (default: off)

Aster89:
The integer option `g:vimtex_quickfix_autoclose_after_keystrokes` sets
the initial number for a countdown of cumulative movements, edits, etc,
at the end of which the QuickFix window automatically closes if the user
has not jumped into it. The countdown is only paused: if the user jumps
back out of the QuickFix window, the countdown continues. The described
behavior does not depend on `g:vimtex_quickfix_mode`. The new option is
off by default (`g:vimtex_quickfix_autoclose_after_keystrokes == -1`).

The countdown is delegated to the new function
`s:qf_check_editing_persistence`, which is called by an `autocmd`
defined in an `augroup`, the latter created whenever the QuickFix window
is opened and deleted by the aforementioned function itself when the
countdown reaches zero (_i.e._ when the QuickFix autocloses).

The triggering events are maybe redundant and should be trimmed.

### DIFF
--- a/autoload/vimtex/qf.vim
+++ b/autoload/vimtex/qf.vim
@@ -85,6 +85,12 @@ function! vimtex#qf#open(force) abort " {{{1
     if g:vimtex_quickfix_mode == 2
       call s:window_restore()
     endif
+    if g:vimtex_quickfix_autoclose_after_keystrokes >= 0
+      augroup vimtex_qf_autoclose
+        autocmd!
+        autocmd CursorMoved,CursorMovedI * call s:qf_autoclose_check()
+      augroup END
+    endif
     redraw
   endif
 endfunction
@@ -176,6 +182,27 @@ endfunction
 
 function! s:qf_has_errors() abort " {{{1
   return len(filter(getqflist(), 'v:val.type ==# ''E''')) > 0
+endfunction
+
+" }}}1
+"
+function! s:qf_autoclose_check() abort " {{{1
+  if get(s:, 'keystroke_counter') == 0
+    let s:keystroke_counter = g:vimtex_quickfix_autoclose_after_keystrokes
+  endif
+  redir => l:bufstring
+  silent! ls!
+  redir END
+  if filter(split(l:bufstring, '\n'), 'v:val =~# ''%a- .*Quickfix''') == []
+    let s:keystroke_counter -= 1
+  else
+    let s:keystroke_counter = g:vimtex_quickfix_autoclose_after_keystrokes + 1
+  endif
+  if s:keystroke_counter == 0
+    cclose
+    autocmd! vimtex_qf_autoclose
+    augroup! vimtex_qf_autoclose
+  endif
 endfunction
 
 " }}}1
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -189,6 +189,7 @@ function! vimtex#init_options() " {{{1
   call s:init_option('vimtex_quickfix_mode', '2')
   call s:init_option('vimtex_quickfix_open_on_warning', '1')
   call s:init_option('vimtex_quickfix_blgparser', {})
+  call s:init_option('vimtex_quickfix_autoclose_after_keystrokes', '0')
 
   call s:init_option('vimtex_texcount_custom_arg', '')
 
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1559,6 +1559,17 @@ Options~
 
   Default value: 2
 
+*g:vimtex_quickfix_autoclose_after_keystrokes*
+  If set to value greater than zero, then the quickfix window will close after
+  this number of motions (i.e. |CursorMoved| and |CursorMovedI| events). This
+  is most useful if one sets |g:vimtex_quickfix_mode| to 2, in which case this
+  option allows one to continue editing and removing the distraction of the
+  quickfix window automatically.
+
+  Note: The count is reset when the quickfix window is entered.
+
+  Default value: 0
+
 *g:vimtex_quickfix_open_on_warning*
   Control whether or not to automatically open the |quickfix| window in case
   there are warning messages and no error messages.


lervag:
Some relevant comments are here:
https://github.com/Aster89/vimtex/commit/907d69d4c0e2233a4de076a8b1631cd3fba4a765#commitcomment-30597425

lervag:
A reply to your last comment on the commit.

> Yes, %a- *Quick part is for that; maybe the minus is not needed, but I
don't think it's bad, just to be surer that the filter detects the qf
window.

Ok!

> I'd like the counter to be global since the only thing that I think
should pause it should be that the qf win is the active win, otherwise the
counter should go down to zero whatever I do. I mean, if the treshold is 5,
2 movements in 1 window + move to other window + 2 movements = 5, and the
qf win must close at the next movement (next = that's why I've added 1 to
my line 190).

Well, all of this should still work with a script local variable.

> Version 7.4.2008 of what? Vim?

Yes.

> `if !exists('g:vimtex_keystrokes_counter') || g:vimtex_keystrokes_counter == 0`
> is to check if the countdown hasn't started yet or it has reached the zero at
least once. I could simplify things up by setting a zero default value for
that variable, which btw should not be set by the user (if set, it would
have effect only for the first occurrence in a session).

Yes, I understood that. This is exactly why I suggest to make it a script local variable.

> What does if get(s:, 'vimtex_keystrokes_counter') == 0 mean?

`get(s:, 'vimtex_keystrokes_counter')` gets the variable from the script local variable list and uses 0 if it doesn't exist, see `:help get()`.

> And can you tell me more about the autocmd! problem?

Not sure what you are talking about...

autoload/vimtex/qf.vim
```diff
@@ -85,6 +85,12 @@ function! vimtex#qf#open(force) abort " {{{1
     if g:vimtex_quickfix_mode == 2
       call s:window_restore()
     endif
+    if g:vimtex_quickfix_autoclose_after_keystrokes >=0
+      augroup autoCloseQf
+        autocmd!
+        autocmd CursorMoved,CursorMovedI,TextChanged,TextChangedI,TextChangedP,TextYankPost * call s:qf_check_editing_persistence()
```
lervag: I think we should restrict to `CursorMoved,CursorMovedI` here, as you proposed earlier.

autoload/vimtex/qf.vim
```diff
@@ -179,3 +185,22 @@ function! s:qf_has_errors() abort " {{{1
 endfunction
 
 " }}}1
+"
+function! s:qf_check_editing_persistence() abort " {{{1
+  if get(s:, 'vimtex_keystrokes_counter') == 0
```
lervag: Since this is now script local, I propose to shorten the name to `keystroke_counter` (also without the plural `s`).

autoload/vimtex/qf.vim
```diff
@@ -179,3 +185,22 @@ function! s:qf_has_errors() abort " {{{1
...
+    let s:vimtex_keystrokes_counter = g:vimtex_quickfix_autoclose_after_keystrokes + 1
+  endif
+  redir => l:bufstring
+  silent! ls!
+  redir END
```
lervag: I did mention `execute()`, but I think we should keep this as is because vimtex should work on Vim version 7.4.

autoload/vimtex/qf.vim
```diff
@@ -85,6 +85,12 @@ function! vimtex#qf#open(force) abort " {{{1
     if g:vimtex_quickfix_mode == 2
       call s:window_restore()
     endif
+    if g:vimtex_quickfix_autoclose_after_keystrokes >=0
+      augroup autoCloseQf
```
lervag: I would prefer if we change the group name to `vimtex_qf_autoclose`, which is more in line with other vimtex related autocmd groups.

autoload/vimtex/qf.vim
Aster89: Yessir, you're right :+1:

autoload/vimtex/qf.vim
Aster89: No dedicated option defaulting to only `CursorMoved,CursorMovedI`?

autoload/vimtex/qf.vim
```diff
@@ -85,6 +85,12 @@ function! vimtex#qf#open(force) abort " {{{1
     if g:vimtex_quickfix_mode == 2
       call s:window_restore()
     endif
+    if g:vimtex_quickfix_autoclose_after_keystrokes >=0
+      augroup vimtex_qf_autoclose
+        autocmd!
+        autocmd CursorMoved,CursorMovedI * call s:qf_check_editing_persistence()
```
lervag: Sorry, one more comment: I think the function name should more resemble the option name. How about `s:qf_autoclose_check_keystrokes` or just `s:qf_autoclose_check`?

autoload/vimtex/qf.vim
lervag: I would prefer to keep the option count low. If we need more options, then I would rather use a dictionary with multiple keys to control this feature. But I would think that we could find a sensible setting that works for all?

lervag:
Good work! Now we need

1. Documentation
2. Possibly change function name
3. Settle the autocmd event question

autoload/vimtex/qf.vim
Aster89: Ok, then let's leave it with two opts. If one they I'll find uncomfortable for this (unlikely!) I'll let you know.

Aster89:
I've forgot the tag, wait!

lervag:
> I've forgot the tag, wait!
Mmm, but it's not tracked....

Not sure what you mean. Should I wait, or not?

Aster89:
I'm sorry, I've been busy for a while. Anyway I meant  the output file of `ctags` without which I can't <kbd>Tab</kbd>-complete `:help g:vimtex_quickfix_auto`. But maybe this is automatically generated?

lervag:
I propose the following help text. I also suggest using 0 for disabled, but I might have misunderstood something.

```
*g:vimtex_quickfix_autoclose_after_keystrokes*
  If set to value larger than zero, then the quickfix window will close after
  this number of motions (i.e. |CursorMoved| and |CursorMovedI| events). This
  is most useful if one sets |g:vimtex_quickfix_mode| to 2, in which case this
  option allows one to continue editing and removing the distraction of the
  quickfix window automatically.

  Default value: 0
```

Aster89:
Yes, it is somehow intended
> > I'd like the counter to be global since the only thing that I think
> > should pause it should be that the qf win is the active win, otherwise the
> > counter should go down to zero whatever I do. I mean, if the treshold is 5,
> > 2 movements in 1 window + move to other window + 2 movements = 5, and the
> > qf win must close at the next movement (next = that's why I've added 1 to
> > my line 190).

After all, we should decide what to do in the basic case of option set to 0. Imho it should mean that the user can't do anythink to keep the qf open *except* jumping to it. So 1 movement is required to trigger the autoclose.

Aster89:
Indeed if you reread the help text (to which I'll certainly add the `This is most useful` part), I considered the option integer value as "the number of movemens *different* from _jumping to the qf_ allowed before autoclose is triggered". It came natural to me, but if you want I'll remove that `+1`.

Aster89:
> Indeed if you reread the help text (to which I'll certainly add the `This is most useful` part), I considered the option integer value as "the number of movemens _different_ from _jumping to the qf_ allowed before autoclose is triggered". It came natural to me, but if you want I'll remove that `+1`.

Ow, I just realized the missin `jumping to`!

Aster89:
I finally agree with you.

Should we mention in the help that, if the user jumps to the qf when the counter has reached 1, then hitting enter on a waring/error and, in turn, jumping to the source file will close the qf?

lervag:
> Should we mention in the help that, if the user jumps to the qf when the counter has reached 1, then hitting enter on a waring/error and, in turn, jumping to the source file will close the qf?

How about resetting the count if one jumps into the qf window? Would that make more sense?

Or, perhaps even more logical: How about disabling the autoclose alltogether? I.e., workflow:

* I'm working on my manuscript, making a minor error so the qf window opens.
* I ignore the error, continue work and the qf autocloses.
* Now I make a new minor error, the qf window opens again, and I choose to fix them.
* Jump to the qf window, select an error, then try to fix it.
* Now, I think it could be natural if the qf window in this case _stays open_. That is, by entering it, I am sort of signalling that I'm interested in the contents.

What do you think?

lervag:
Ok, then I think a good compromise is to reset. This should also be mentioned in the docs, e.g. add a sentence: "The count is reset when the quickfix window is entered."

Aster89:
@lervag , I've done the edits, note that the `+ 1` in the `else` is to take into account movement of jumping back to a non-qf window, so that the countdown starts from the moment the cursor *is* in the non-qf win and not *goes* to the non-qf win. I find it more consistent. Do you agree?

lervag:
Yes, I agree! Good work! I'll take another quick look later, then rebase and merge.

lervag:
Thanks, this has now been merged. Good work, I appreciate your effort!

PR #1240, opened 2018-10-09
Fix `csc` command broken when new name is very short

lacygoill:
Consider this minimal `vimrc`:

    set rtp^=/path/to/vimtex
    set rtp+=/path/to/vimtex/after
    filetype plugin indent on

And this tex file:

    \usepackage[utf8]{inputenc}

Press `csc` to change the command while your cursor is on the first character of the line.  
Insert `a` while the plugin asks for your input, and press Enter to validate.  
The following error is raised:

    > aError detected while processing function vimtex#cmd#change[31]..vimtex#pos#set_cursor:
    line    1:
    E474: Invalid argument

gif

The issue seems to be influenced by several factors, including the cursor position on the command name, and the length of the user's input. Depending on those factors, the error may or may not be raised.  
However, if you prefix `csc` with a count, the error is always raised.  
I think it's due to the `vimtex#pos#set_cursor()` function which calls the `cursor()` function.  
Sometimes, the byte index of the column position in the list passed as an argument may be negative.

This PR tries to make sure it's reset to `0` when that happens.  
In my limited testing, it seems to fix the issue.  

I use `map()` to reset any negative number to `0`, because:

 - I had various errors, sometimes the negative number was in the 2nd position in the list, sometimes in the 3rd (I could be wrong about the exact positions, but they vary)
 - I don't think any negative number makes sense in a list describing a position in a buffer

### DIFF
--- a/autoload/vimtex/pos.vim
+++ b/autoload/vimtex/pos.vim
@@ -5,7 +5,7 @@
 "
 
 function! vimtex#pos#set_cursor(...) " {{{1
-  call cursor(s:parse_args(a:000))
+  call cursor(map(s:parse_args(a:000), 'v:val < 0 ? 0 : v:val'))
 endfunction
 
 " }}}1


andymass:
I think `csc` could be smarter in general about where it places the cursor if the cursor is on the `\command` portion.  Consider,

    \command{text}
    \comm█nd{text}     % to show the cursor

after `csc` `a<cr>`:

    \a{te█t}

In this case the faulty position computation gave 0.  I would probably expect the cursor to land here:

    \█{text}

lacygoill:
@andymass I think I can get the behavior you describe by replacing this line in `vimtex#cmd#change()`:

    let l:save_pos[2] += strlen(l:new_name) - strlen(l:old_name) + 1

With this `if` block:

    if strlen(l:new_name) < strlen(l:old_name)
      let l:save_pos[2] = searchpos('\\\k*{\k.\{-}}', 'bcnW')[1] + strlen(l:new_name)
    endif

Incidentally, it also fixes the bug of the original post, so there would be no need to change `vimtex#pos#set_cursor()` anymore.

Before:
the code computes the new column position by adding the difference between the lengths of the new command name and the old one.

After:
the code leave the new column position if the new name is longer than the old one.
So, if you are on the 5th character of a command name which contains 10 characters, and you replace it with a new command name which contains 20 characters, you are still on the 5th character.
It's a design decision, so I'm not sure everybody would agree with such a position, but I think that's what I would expect.

OTOH, if the new command name is shorter, then the code tries to position the cursor on the end of the latter.

I haven't tested it on anything other than the example you provided, so I'm not sure it's reliable.  
The regex and the flags passed to `searchpos()` may need to be changed...

lervag:
I think this was fixed by my latest commit. Thanks for the suggested fix, @lacygoill! I think it fixes both this issue and the concern by @andymass. I've made a minor modification to ensure that the cursor stays put if possible.

Issue #1238, opened 2018-10-08
Unexpected repetition of `csc` with the dot command after an undo


lacygoill:
Consider this minimal vimrc:

    set rtp^=~/.vim/plugged/vim-repeat/
    set rtp^=~/.vim/plugged/vimtex
    set rtp+=~/.vim/plugged/vimtex/after/
    filetype plugin indent on

And the following tex file:

    \usepackage[utf8]{inputenc}

Position the cursor at the end of `usepackage` (i.e. its last `e` character).
Press `dd` to delete the line, then `u` to undo.
Press `csc` to change the command name.
Provide the name `foo` and press Enter to validate.
Press `u` to undo, then `.` to repeat the command.

I would expect the substitution of the command to be repeated, and get:

    \foo[utf8]{inputenc}

But what actually happens is that the line is deleted:

![gif_broken](https://user-images.githubusercontent.com/8505073/46631347-699fd880-cb47-11e8-9ba0-2fb4f8ca5709.gif)

I think this is due to the fact that the `vimtex#cmd#change()` function in `autoload/vimtex/cmd.vim` calls the `repeat#set()` function at its end.  
`repeat#set()` doesn't seem reliable after an undo. [An issue](https://github.com/tpope/vim-repeat/issues/63#issuecomment-390281441) has been opened on `vim-repeat`'s bug tracker, where its author says that they prefer to avoid it in the future, in favor of an operator function:

> The direction I am moving is avoid repeat.vim wherever I can in favor of an operator func, even if I have to shoehorn it in.

`repeat#set()` can lead to other similar issues. I remember using `repeat#set()` to repeat a custom function, for which `.` could sometimes repeat the last but one edit, instead of the last edit, even without any undo.

Anyway, I agree with tpope, and I've removed all occurrences of `repeat#set()` in my configuration files, in favor of operator functions, as they don't suffer from these issues. Besides, it removes a dependency.

I've looked at my plugins:

    :vim /repeat#set(/gj ~/.vim/**/*.vim ~/.vim/**/*.snippets ~/.vim/vimrc | cw

And noticed that `vimtex` calls `repeat#set()` in 12 different locations.
The current issue is for one of them.

To eliminate `repeat#set()`, you could use a “pseudo” operator function.
pseudo because the user would never be able to provide an arbitrary text-object, like they can for the other operator functions.
Instead, you would always pass it the motion `l`.
I think that's what tpope meant when he used the word “shoehorn”.
[...]

PR #1247, opened 2018-10-16
Remove `vim-repeat` dependency to fix issue #1238

lacygoill:
## Introduction

This PR tries to eliminate `vim-repeat` as a dependency, in order to fix the [issue 1238](https://github.com/lervag/vimtex/issues/1238).  It affects these `<plug>` key sequences used as a lhs in some mappings:

| <plug> key sequence                            | mapped by default to | installed in file         |
| -----------------------------------------------|----------------------|---------------------------|
| `<plug>(vimtex-cmd-change)`                    | `csc`                | autoload/vimtex/cmd.vim   |
| `<plug>(vimtex-cmd-delete)`                    | `dsc`                | autoload/vimtex/cmd.vim   |
| `<plug>(vimtex-cmd-create)`                    | `F7`                 | autoload/vimtex/cmd.vim   |
| `<plug>(vimtex-cmd-toggle-star)`               | `tsc`                | autoload/vimtex/cmd.vim   |
| <br>                                           |                      |                           |
| `<plug>(vimtex-delim-toggle-modifier)`         |  `tsd`               | autoload/vimtex/delim.vim |
| `<plug>(vimtex-delim-toggle-modifier-reverse)` |  `tsD`               | autoload/vimtex/delim.vim |
| `<plug>(vimtex-delim-delete)`                  |  `dsd`               | autoload/vimtex/delim.vim |
| `<plug>(vimtex-delim-change-math)`             |  `csd`               | autoload/vimtex/delim.vim |
| <br>                                           |                      |                           |
| `<plug>(vimtex-env-delete)`                    | `dse`                | autoload/vimtex/env.vim   |
| `<plug>(vimtex-env-delete-math)`               | `ds$`                | autoload/vimtex/env.vim   |
| `<plug>(vimtex-env-toggle-star)`               | `tse`                | autoload/vimtex/env.vim   |
| `<plug>(vimtex-env-change)`                    | `cse`                | autoload/vimtex/env.vim   |
| `<plug>(vimtex-env-change-math)`               | `cs$`                | autoload/vimtex/env.vim   |

---

I've tried to respect the style of the original code (indentation level, explicit `l:` scope, full name of commands, folding markers ...), but I may have missed something.

There may be too much code, and even though I made some manual tests, I'm not sure it's completely reliable, so feel free to close the PR if it can't be merged.

Some questions/answers about how the changes work in the following.

## Why executing `:normal` to press `g@l`?

IOW, why this:

```vim
nnoremap <silent><buffer> <plug>(vimtex-cmd-delete)
    \ :<c-u>call <sid>setup_operator('delete')<bar>normal! g@l<cr>
                                                   ^^^^^^^^^^^
```

Instead of this:

```vim
nnoremap <silent><buffer> <plug>(vimtex-cmd-delete)
    \ :<c-u>call <sid>setup_operator('delete')<cr>g@l
                                                  ^^^
```

Here, both versions work, but the second one won't work if `s:normal_mode()` calls `input()`. A MWE:

```vim
nno  cd  :<c-u>call Func()<cr>bbb
fu! Func() abort
    call input('>')
endfu
```

Press `cd`, and you'll see that `bbb` has been consumed by `input()` and inserted on the command-line, instead of being pressed.  But if you rewrite the mapping like this, it works as expected:

```vim
nno  cd  :<c-u>call Func()<bar>norm! bbb<cr>
fu! Func() abort
    call input('> ')
endfu
```

To stay consistent, I've always used the first version, even when it was not necessary, so that if you change the code later and add an `input()` somewhere, it won't consume `g@l`.

## Why changing `vimtex#cmd#create_ask()`?

Before, this function was called by pressing `F7` or the dot command. In both cases, it asked the user for a command name.

However, when the user pressed `.`, they didn't have to type anything on the input prompt because `repeat#set()` wrote the last command name input by the user inside the typeahead buffer, followed by a CR, and so `input()` consumed the latter automatically. As a result, the same function `create_ask()` could behave differently depending on how it was called without inspecting any variable.

I could write the command name in the typeahead buffer, but I would need to do it only when the user pressed the dot command. I don't know how to do that without remapping `.` which would lead to re-write `vim-repeat` and defeat the purpose of this PR.

So, I had to change the signature of `create_ask()` to pass it a second flag, that it can inspect to decide whether it should prompt the user for a command name. Note that not only does it fix the issue `#1238` but it also removes the brief display of the input prompt which may be distracting to some users.

## Do the commands support a count?

No they don't.

I could have added support for a count, but:

 - It would have added a lot of code/complexity.
 - In the [latest commit](https://github.com/lervag/vimtex/commit/34e2423391daf614123f618910eaaedcc60a10cd), you have prefixed some of the `:call` used in mappings with `C-u`, which kills the repetition of the called functions, and the latters don't inspect `v:count[1]`; so I assume you didn't intend these commands to support a count.
 - I've searched for the pattern `count` in `:h vimtex`, and couldn't find a match where the documentation was saying that the commands are meant to support a count.
 - Even if it's not strictly equivalent, one can still press `.` to repeat a command a few times.

## Why removing `vimtex#echo#input()` from `vimtex#cmd#change()`?

`vimtex#cmd#change()` must be called by an operator function (`s:opfunc()`) to be made dot repeatable. And, inside an opfunc, I don't know how to print a message, then prompt the user for some input, without the message being erased. The only way I found to maintain the old behavior of the command, was to move `input()` outside the opfunc.

MWE:

```vim
nno cd :set opfunc=Func<cr>g@l
fu! Func(_)
    echo 'question'
    let answer = input('>')
endfu
" Press `cd`.
" You'll only see the prompt (`>`), not the message (`question`).
```

I think the issue comes from Vim which, inside an operator function, automatically redraws before any command which prints a message (like `:echo`, or `:call input()`). I don't know how to prevent it. `:h echo-redraw` doesn't help here. Adding a `:redraw` before the call to `input()` doesn't prevent Vim from erasing the previous message.

Just as an experiment, I also tried to print 2 messages with 2 `:echo` from a timer:

```vim
" ✔
:echo 'foo' | echo 'bar'
"    → foo
"      bar

" ✘
:call timer_start(0, {-> execute('echo "foo" | echo "bar"', '')})
"    → bar
```

And it suffers from the same issue. Same thing from an autocmd:

```vim
augroup test_echo
    au!
    au CursorHold * call Func() | au! test_echo
augroup END
fu! Func()
    echo 'foo'
    echo 'bar'
endfu
" ✘ wait a few seconds, only 'bar' is printed
```

## Why removing `vimtex#env#change()` from `vimtex#env#change_prompt()`?

For the same reason as before, I need `input()` to be called from outside an opfunc. So, I call `vimtex#env#change_prompt()` from `s:setup_operator()` which is called before the opfunc, then the latter can call `vimtex#env#change()` to make the change.

## Why checking `a:0` in `vimtex#env#change()`?

Originally, `vimtex#env#change()` could be called by `vimtex#env#toggle_star()` and be passed 3 arguments. It's still the case. However, it could also be called by `vimtex#env#change_prompt()`. It's not the case anymore, because again, `input()` must be called from outside an opfunc. So, `s:setup_operator()` calls `vimtex#env#change_prompt()` to get the user's input, then later `vimtex#env#change()` is called to make the change. However, I can't pass the 3 arguments that `vimtex#env#change()` expected originally. I could pass the user input (`a:new`), but not the other 2 arguments (`a:open`, `a:close`), because, contrary to the input which can be re-used when `.` is pressed, those must be recomputed for every repeat.

To summarize, when `vimtex#env#change()` is called by `vimtex#env#toggle_star()`, it can be passed its original 3 arguments, but not when it's called by `vimtex#env#change_prompt()`. That's why the function must check `a:0` to determine from where it was called and get its arguments accordingly.

### DIFF
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -6,27 +6,27 @@
 
 function! vimtex#cmd#init_buffer() " {{{1
   nnoremap <silent><buffer> <plug>(vimtex-cmd-delete)
-        \ :<c-u>call vimtex#cmd#delete()<cr>
+        \ :<c-u>call <sid>setup_operator('delete')<bar>normal! g@l<cr>
 
   nnoremap <silent><buffer> <plug>(vimtex-cmd-change)
-        \ :<c-u>call vimtex#cmd#change()<cr>
+        \ :<c-u>call <sid>setup_operator('change')<bar>normal! g@l<cr>
 
   inoremap <silent><buffer> <plug>(vimtex-cmd-create)
         \ <c-r>=vimtex#cmd#create_insert()<cr>
 
   nnoremap <silent><buffer> <plug>(vimtex-cmd-create)
-        \ :<c-u>call vimtex#cmd#create_ask(0)<cr>
+        \ :<c-u>call <sid>setup_operator('create_ask')<bar>normal! g@l<cr>
 
   xnoremap <silent><buffer> <plug>(vimtex-cmd-create)
         \ :<c-u>call vimtex#cmd#create_ask(1)<cr>
 
   nnoremap <silent><buffer> <plug>(vimtex-cmd-toggle-star)
-        \ :<c-u>call vimtex#cmd#toggle_star()<cr>
+        \ :<c-u>call <sid>setup_operator('toggle_star')<bar>normal! g@l<cr>
 endfunction
 
 " }}}1
 
-function! vimtex#cmd#change(...) " {{{1
+function! vimtex#cmd#change(new_name) " {{{1
   let l:cmd = vimtex#cmd#get_current()
   if empty(l:cmd) | return | endif
 
@@ -35,14 +35,7 @@ function! vimtex#cmd#change(...) " {{{1
   let l:cnum = l:cmd.pos_start.cnum
 
   " Get new command name
-  if a:0 > 0
-    let l:new_name = a:1
-  else
-    let l:new_name = vimtex#echo#input({
-          \ 'info' : ['Change command: ', ['VimtexWarning', l:old_name]],
-          \})
-  endif
-  let l:new_name = substitute(l:new_name, '^\\', '', '')
+  let l:new_name = substitute(a:new_name, '^\\', '', '')
   if empty(l:new_name) | return | endif
 
   " Update current position
@@ -61,10 +54,8 @@ function! vimtex#cmd#change(...) " {{{1
         \ . l:new_name
         \ . strpart(l:line, l:cnum + strlen(l:old_name) - 1))
 
-  " Restore cursor position and create repeat hook
+  " Restore cursor position
   cal vimtex#pos#set_cursor(l:save_pos)
-  silent! call repeat#set(
-        \ "\<plug>(vimtex-cmd-change)" . l:new_name . '', v:count)
 endfunction
 
 function! vimtex#cmd#delete(...) " {{{1
@@ -110,9 +101,6 @@ function! vimtex#cmd#delete(...) " {{{1
     endif
   endif
   cal vimtex#pos#set_cursor(l:save_pos)
-
-  " Create repeat hook
-  silent! call repeat#set("\<plug>(vimtex-cmd-delete)", v:count)
 endfunction
 
 function! vimtex#cmd#delete_all(...) " {{{1
@@ -157,11 +145,15 @@ endfunction
 
 " }}}1
 function! vimtex#cmd#create_ask(visualmode) " {{{1
-  let l:cmd = vimtex#echo#input({
-        \ 'info' :
-        \   ['Change command: ', ['VimtexWarning', '(empty to cancel)']],
-        \})
-  let l:cmd = substitute(l:cmd, '^\\', '', '')
+  if a:visualmode
+    let l:cmd = vimtex#echo#input({
+          \ 'info' :
+          \   ['Change command: ', ['VimtexWarning', '(empty to cancel)']],
+          \})
+    let l:cmd = substitute(l:cmd, '^\\', '', '')
+  else
+    let l:cmd = s:cmd_name_create_ask
+  endif
   if empty(l:cmd) | return | endif
 
   " Avoid autoindent (disable indentkeys)
@@ -192,8 +184,6 @@ function! vimtex#cmd#create_ask(visualmode) " {{{1
     let l:save_reg = getreg('"')
     let l:pos[2] += strlen(l:cmd) + 2
     execute 'normal! ciw\' . l:cmd . '{"}'
-    silent! call repeat#set(
-          \ "\<plug>(vimtex-cmd-create)" . l:cmd . '', v:count)
     call setreg('"', l:save_reg)
     call vimtex#pos#set_cursor(l:pos)
   endif
@@ -231,9 +221,38 @@ function! vimtex#cmd#toggle_star() " {{{1
         \ . l:new_name
         \ . strpart(l:line, l:cnum + strlen(l:old_name) - 1))
 
-  " Restore cursor position and create repeat hook
+  " Restore cursor position
   cal vimtex#pos#set_cursor(l:save_pos)
-  silent! call repeat#set("\<plug>(vimtex-cmd-toggle-star)", v:count)
+endfunction
+
+" }}}1
+
+function! s:setup_operator(operator) abort " {{{1
+  let &opfunc = s:snr().'opfunc'
+  let s:operator = a:operator
+  if s:operator isnot# 'change' && s:operator isnot# 'create_ask' | return | endif
+  if s:operator is# 'change' && empty(vimtex#cmd#get_current()) | return | endif
+  let s:cmd_name_{a:operator} = vimtex#echo#input({
+        \ 'info' : ['Change command: ',
+        \   ['VimtexWarning',
+        \     a:operator is# 'change' ? vimtex#cmd#get_current().name : '(empty to cancel)']],
+        \})
+  let s:cmd_name_{a:operator} = substitute(s:cmd_name_{a:operator}, '^\\', '', '')
+endfunction
+
+" }}}1
+function! s:opfunc(_) abort " {{{1
+  execute 'call vimtex#cmd#'.{
+        \   'change': 'change(get(s:, "cmd_name_change", ""))',
+        \   'create_ask': 'create_ask(0)',
+        \   'delete': 'delete()',
+        \   'toggle_star': 'toggle_star()',
+        \ }[s:operator]
+endfunction
+
+" }}}1
+function! s:snr() " {{{1
+  return matchstr(expand('<sfile>'), '<SNR>\d\+_')
 endfunction
 
 " }}}1
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -230,12 +230,12 @@ endfunction
 function! s:setup_operator(operator) abort " {{{1
   let &opfunc = s:snr() . 'opfunc'
   let s:operator = a:operator
-  if s:operator isnot# 'change' && s:operator isnot# 'create_ask' | return | endif
-  if s:operator is# 'change' && empty(vimtex#cmd#get_current()) | return | endif
+  if s:operator !=# 'change' && s:operator !=# 'create_ask' | return | endif
+  if s:operator ==# 'change' && empty(vimtex#cmd#get_current()) | return | endif
   let s:cmd_name_{a:operator} = vimtex#echo#input({
         \ 'info' : ['Change command: ',
         \   ['VimtexWarning',
-        \     a:operator is# 'change' ? vimtex#cmd#get_current().name : '(empty to cancel)']],
+        \     a:operator ==# 'change' ? vimtex#cmd#get_current().name : '(empty to cancel)']],
         \})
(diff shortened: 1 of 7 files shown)


andymass:
> but the second one won't work if s:normal_mode() calls input().

Would `inputsave()`/`inputrestore()` help here?  (I haven't thought too hard about it though).

> I've searched for the pattern count in :h vimtex

The `tsd`/`tsD` are supposed to accept a count:

https://github.com/lervag/vimtex/blob/34e2423391daf614123f618910eaaedcc60a10cd/doc/vimtex.txt#L2364

I wonder, since these particular maps do not take user input, would adding the count back be possible?  Maybe,

       \ :<c-u>call <sid>setup_operator('...')<bar>execute 'normal!' (v:count ? v:count : '') . 'g@l'<cr>

or 

       \ :<c-u>call <sid>setup_operator('...')<bar>normal! <c-r>=v:count ? v:count : ''<cr>g@l<cr>

lacygoill:
@andymass Thank you very much for the comments.

> Would inputsave()/inputrestore() help here? (I haven't thought too hard about it though).

It would help, but I tried to minimize the number of lines of code.
Using `:normal! g@l` should be more future-proof without adding 2 lines of code in each opfunc.

> The tsd/tsD are supposed to accept a count:

Ah yes you're right, I missed this part or didn't see the link between it and `tsd`/`tsD`.

> I wonder, since these particular maps do not take user input, would adding the count back be possible? Maybe,
>
>    `\ :<c-u>call <sid>setup_operator('...')<bar>execute 'normal!' (v:count ? v:count : '') . 'g@l'<cr>`
>
> or
>
>    `\ :<c-u>call <sid>setup_operator('...')<bar>normal! <c-r>=v:count ? v:count : ''<cr>g@l<cr>`

I think they should both work, I have commited a second commit using the second version, as it's shorter, and squashed the 2 commits.

lervag:
@lacygoill Thank you for this, I think you have done a pretty good job here. I took the liberty of doing some minor modification of your formatting, as it made it easier for me to read and understand what you've written. I hope you don't mind. For future updates, please note that you should keep a paragraph without line breaks for optimal text flow on github.

After the latest commit and input from @andymass, I can't see anything wrong with your arguments. But the tests are failing. I investigated some, and I've attached a patch that should fix the failing tests (copy the diff to a file, then use `git apply`). Finally, I will also need to do a review of the code before merging.

If I understand correctly, you are claiming that this PR should make vimtex completely independent of vim-repeat, right?

```diff
diff --git a/test/vader/cmd.vader b/test/vader/cmd.vader
index 1797a85..62e88ef 100644
--- a/test/vader/cmd.vader
+++ b/test/vader/cmd.vader
@@ -11,8 +11,8 @@ Expect tex (Cmd: verify):
   foobar
   foo
 
-Do (Cmd: dsc):
-  dsc
+Execute (Cmd: dsc):
+  execute "normal dsc"
 
 Expect tex (Cmd: verify):
   foo
diff --git a/test/vader/delims.vader b/test/vader/delims.vader
index 080c2e5..1c0f581 100644
--- a/test/vader/delims.vader
+++ b/test/vader/delims.vader
@@ -4,8 +4,8 @@ Given tex (Delims: toggle in complicated math):
          & 1 \\
   \end{smallmatrix}\bigr)$
 
-Do (Toggle):
-  3jtsd
+Execute (Toggle):
+  execute 'normal 3jtsd'
 
 Expect tex (Verify):
   $(\begin{smallmatrix}
diff --git a/test/vader/env.vader b/test/vader/env.vader
index 823ed95..d7dfdc3 100644
--- a/test/vader/env.vader
+++ b/test/vader/env.vader
@@ -15,8 +15,8 @@ Given tex (Environments: foo and bar):
     Bar
   \end{bar}
 
-Do (Change env: foo -> baz):
-  csebaz\<cr>
+Execute (Change env: foo -> baz):
+  execute "normal csebaz\<cr>"
 
 Expect tex (Verify):
   \begin{baz}
@@ -27,8 +27,8 @@ Expect tex (Verify):
     Bar
   \end{bar}
 
-Do (Change env: foo -> bar (with completion)):
-  cse\<c-z>\<c-z>\<c-z>\<cr>
+Execute (Change env: foo -> bar (with completion)):
+  execute "normal cse\<c-z>\<c-z>\<c-z>\<cr>"
 
 Expect tex (Verify):
   \begin{bar}
@@ -47,8 +47,8 @@ Given tex (Environments: test and center):
     \begin{center} a \end{center}
   \end{test}
 
-Do (Delete env: test and center):
-  dsedse
+Execute (Delete env: test and center):
+  execute 'normal dsedse'
 
 Expect tex (Verify):
      a 
```

autoload/vimtex/cmd.vim
```diff
@@ -231,9 +221,38 @@ function! vimtex#cmd#toggle_star() " {{{1
...
-  " Restore cursor position and create repeat hook
+  " Restore cursor position
   cal vimtex#pos#set_cursor(l:save_pos)
-  silent! call repeat#set("\<plug>(vimtex-cmd-toggle-star)", v:count)
+endfunction
+
+" }}}1
+
+function! s:setup_operator(operator) abort " {{{1
+  let &opfunc = s:snr().'opfunc'
```
lervag: 1. I prefer to keep space before/after `.` (makes it more distinguishable from dictionary dots).

2. I assume that `s:snr()` is the standard way of accessing script local functions when we've lost the scope. Just curious, are you sure we need it, i.e., that we can't just do `s:opfunc`?

autoload/vimtex/cmd.vim
```diff
@@ -231,9 +221,38 @@ function! vimtex#cmd#toggle_star() " {{{1
...
+  let s:operator = a:operator
+  if s:operator isnot# 'change' && s:operator isnot# 'create_ask' | return | endif
+  if s:operator is# 'change' && empty(vimtex#cmd#get_current()) | return | endif
+  let s:cmd_name_{a:operator} = vimtex#echo#input({
+        \ 'info' : ['Change command: ',
+        \   ['VimtexWarning',
+        \     a:operator is# 'change' ? vimtex#cmd#get_current().name : '(empty to cancel)']],
+        \})
+  let s:cmd_name_{a:operator} = substitute(s:cmd_name_{a:operator}, '^\\', '', '')
+endfunction
```
lervag: One idea for minor improvement is to make `s:setup_operator` and `s:opfunc` methods of a "class" `s:operator`. Call it semi-OO, if you want. This way, there would be less `s:` variables, since `s:setup_operator` could save state to itself. Something like this:

```vim
let s:operator = {}

function! s:operator.setup(operator) dict abort
  "...
endfunction

function! s:operator.opfunc(_) dict abort
￼ execute 'call vimtex#cmd#' . {
￼       \   'change': 'change(get(s:, "cmd_name_change", ""))',
￼       \   'create_ask': 'create_ask(0)',
￼       \   'delete': 'delete()',
￼       \   'toggle_star': 'toggle_star()',
￼       \ }[self.operator]
endfunction
```

Another benefit of this is that it would make it slightly easier to read the code when e.g. `vimtex#cmd#change` uses a script local variable, since it would use `s:operator.<var>`, which would instantly be read as something that is coupled to the operator code.

autoload/vimtex/cmd.vim
```diff
@@ -231,9 +221,38 @@ function! vimtex#cmd#toggle_star() " {{{1
...
   cal vimtex#pos#set_cursor(l:save_pos)
-  silent! call repeat#set("\<plug>(vimtex-cmd-toggle-star)", v:count)
+endfunction
+
+" }}}1
+
+function! s:setup_operator(operator) abort " {{{1
+  let &opfunc = s:snr().'opfunc'
+  let s:operator = a:operator
+  if s:operator isnot# 'change' && s:operator isnot# 'create_ask' | return | endif
```
lervag: Is there a good reason to use `is` and `isnot` instead of `==` and `!=`?

autoload/vimtex/delim.vim
```diff
@@ -202,6 +199,46 @@ endfunction
...
+          \   ['Change surrounding delimiter: ', ['VimtexWarning', l:name]],
+          \ 'complete' : 'customlist,vimtex#delim#change_input_complete',
+          \})
+  endif
+endfunction
+
+" }}}1
+function! s:opfunc(_) abort " {{{1
+  if s:operator is# 'toggle_modifier'
+    call call('vimtex#delim#toggle_modifier', [s:toggle_modifier_dir])
```
lervag: Can you not just call this directly?

autoload/vimtex/delim.vim
```diff
@@ -202,6 +199,46 @@ endfunction
...
+  else
+  execute 'call vimtex#delim#'
```
lervag: Wrong indentation.

lervag:
Great, thank you for the updates!

lervag:
Thank you for the useful input! I'll look into the possibility of making this slightly more maintable, but it seems that the gain might be lost in the extra burden.

Also, I was not aware of the distinction between making the dictionary functions directly and between defining them and allocating them wrt. the debugging part. This is useful, and I agree it makes a lot of sense to avoid defining functions that are not named when debugging.

lervag:
Again, thank you! I really appreciate your contribution!

lervag:
I've attempted to make some modifications that made sense to me. I mostly kept the structure you suggested, but I tried to avoid using script local variables as much as possible, keeping them inside the operator wrapper functions.

PR #1254, opened 2018-11-03
Prioritize local mappings over global ones

lacygoill:
Some local mappings have been given the `<nowait>` argument, but not all.

This PR tries to give it to all local mappings, so that in case of a conflict with a global mapping, the local one wins immediately, and the user doesn't have to wait for the timeout to pass.

I found these mappings with this `:vimgrep` command which looks for the pattern `<buffer>`:

    :vim /<buffer>/gj ~/.vim/plugged/vimtex/**/*.vim | cw

Then I filtered the results and removed the lines which matched the patterns `<nowait>`, `autocmd`, or `<buffer>\s*\%(<plug>\|<sid>\)` (because the only mappings which really seem to matter are the ones whose lhs can be pressed by the user):

    :Cfilter! <nowait>\|autocmd\|<buffer>\s*\%(<plug>\|<sid>\)

The `:Cfilter` command is provided by the default `cfilter` package loaded with `:packadd cfilter`.

Feel free to close the PR if `<nowait>` causes an issue.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -399,7 +399,7 @@ function! s:init_default_mappings() " {{{1
     if !hasmapto(a:rhs, a:mode)
           \ && index(get(g:vimtex_mappings_disable, a:mode, []), a:lhs) < 0
           \ && (empty(maparg(a:lhs, a:mode)) || a:0 > 0)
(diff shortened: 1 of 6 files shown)


lervag:
Thank you! I think this looks correct.

PR #1255, opened 2018-11-03
Make functions abort immediately when an error is raised

lacygoill:
Some functions have been given the `abort` argument, but not all.

This PR tries to give it to all functions, so that if an error is raised, the function aborts immediately.

I found these functions with this `:vimgrep` command:

    :noa vim /^\s*fu\%[nction]!\=\%(.*abort\)\@!.*\zs\s*"\s*{{{\|^\s*fu\%[nction]!\=\%(.*abort\)\@!.*\zs/gj ~/.vim/plugged/vimtex/**/*.vim | cw

Feel free to close the PR if adding `abort` causes an issue.

---

I was tempted to submit another PR to add the keycode `<c-u>` in front of all the `:call` commands in mappings, so that if the user presses a count by accident before the lhs, Vim won't translate it into a range which would cause the subsequent function to be called several times, and may have undesired effects.

MWE:

    nno  cd  :call Func()<cr>
    fu! Func()
        echo 'hello'
    endfu
    " press `3cd`, and you'll see 'hello' printed 3 times

I found 35 matches with this `:vimgrep` command:

    :noa vim /\m\cmap.*\%(<c-u>\s*\)\@<!\zs\<call\>/gj ~/.vim/plugged/vimtex/**/*.vim | cw

But I don't know the codebase well enough, and it's possible that some of them expect a range because the functionality they help to implement supports a count, so I didn't submit another PR.

### DIFF
--- a/autoload/health/vimtex.vim
+++ b/autoload/health/vimtex.vim
@@ -63,7 +63,7 @@ endfunction
 
 " }}}1
 
-function! s:check_plugin_clash() " {{{1
+function! s:check_plugin_clash() abort " {{{1
   let l:scriptnames = split(execute('scriptnames'), "\n")
 
   let l:latexbox = !empty(filter(copy(l:scriptnames), "v:val =~# 'latex-box'"))
--- a/autoload/unite/sources/vimtex.vim
+++ b/autoload/unite/sources/vimtex.vim
@@ -17,7 +17,7 @@ let s:source = {
       \ 'hooks' : {},
       \}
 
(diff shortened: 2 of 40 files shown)


lervag:
Again, thank you!

Reg. `<c-u>`: I would not be surprised if there might be some errors present, but I also know that there are several cases where adding `<c-u>` is not correct. Thus one would have to more carefully check each of the 35 cases if one were to perform this update, so that one does not add `<c-u>` where it should not be added.

PR #1267, opened 2018-12-04
Make vimcomplete.bst work with date field

clason:
Biblatex recommends using the `date` field over `year`, which was not recognized by the bibliography style used to generate the omnicomplete menu. This commit adds `date` as a field and uses this if `year` does not exist. (If both don't exist, the same fallback is used as before.)


lervag:
Thanks!

PR #1285, opened 2018-12-17
Add viewer support for cygwin environments

pirofti:
When running Vim inside Cygwin the paths are relative to the Cygwin environment, they are not real Windows paths. Example: `/home/pirofti/foo.pdf` versus `C:\cygwin\home\pirofti\foo.pdf`. 

PDF viewers are usually installed in Windows not in Cygwin. So when the viewer is invoked with the PDF path, Vimtex sends a Cygwin path instead of a Windows path and the viewer does not know how to parse that.

This patch checks if we are running inside Cygwin by checking for the specific `cygpath` utility that is also used to generate a Windows path from a Cygwin path.

### DIFF
--- a/autoload/vimtex/view/general.vim
+++ b/autoload/vimtex/view/general.vim
@@ -54,6 +54,10 @@ function! s:general.view(file) dict abort " {{{1
   let l:cmd = substitute(l:cmd, '@col', col('.'), 'g')
   let l:cmd = substitute(l:cmd, '@tex',
         \ vimtex#util#shellescape(expand('%:p')), 'g')
+  " Check if we are inside Cygwin and call the viewer with a Windows path
+  if executable('cygpath')
+    let outfile = substitute(system('cygpath -aw '.outfile), '\n', '', 'g')
+  endif
   let l:cmd = substitute(l:cmd, '@pdf', vimtex#util#shellescape(outfile), 'g')
 
   " Start the view process


lervag:
I'm not quite sure this is correctly handled. Could you provide a minimal example (vimrc file and minimal latex file), and then some example output from `\li`?

Some questions:
- Is it only for the PDF viewer that we want to use the `cygpath` tool?
- If not, would it not make sense to use `cygpath` on the vimtex.root variable?

pirofti:
I am not sure either :)

This is only for the viewer.

`sumatrapdf /home/pirofti/foo.pdf` makes no sense to the Windows binary. Instead, my patch feeds it the true Windows path.

The viewer is the only one affected. Compilation works. This is only for `\lv`.

If you still want a minimal example, let me know and I'll try to produce one in the morning.

Thanks for looking into this,
Paul Irofti


În 17 decembrie 2018 23:18:46 EET, "Karl Yngve Lervåg" <notifications@github.com> a scris:
>I'm not quite sure this is correctly handled. Could you provide a
>minimal example (vimrc file and minimal latex file), and then some
>example output from `\li`?
>
>Some questions:
>- Is it only for the PDF viewer that we want to use the `cygpath` tool?
>- If not, would it not make sense to use `cygpath` on the vimtex.root
>variable?

PR #1303, opened 2019-01-09
Rename amsmath texMathZones

kiryph:
according to amsmath.vba.gz
version 1c ASTRO-ONLY from Jun 29, 2018

This also removes the duplicate for equation.


lervag:
Thanks!

Issue #1318, opened 2019-02-02
Visual mode is interrupted while a "VimtexCompileOutput" window is open


ishitaku5522:
## Issue
Visual mode is interrupted on CursorMoved while a "VimtexCompileOutput" window is open.

## How to reproduce
1. Open a .tex file
1. Press \\lo to open VimtexCompileOutput window
1. Move cusor to the window of .tex file
1. Press v to enter visual mode
1. Press l to move cusor

## Expected behavior
Keeping visual mode active even while a "VimtexCompileOutput" window is open.

## Cause of this issue
Changing window focus in s:output.update() interrupts visual mode:
https://github.com/lervag/vimtex/blob/master/autoload/vimtex/compiler.vim#L203

## Solution
Not update ouput window while visual mode is active. I think interruption of visual mode is not natural behavior for users and  comfortable text editing is more important than updating output.

See: https://github.com/lervag/vimtex/pull/1319

lervag:
Thanks for posting and for submitting a fix! We'll continue the discussion in #1319.

PR #1319, opened 2019-02-02
Disable update of ouput window while visual mode is active

ishitaku5522:
Fix https://github.com/lervag/vimtex/issues/1318

### DIFF
--- a/autoload/vimtex/compiler.vim
+++ b/autoload/vimtex/compiler.vim
@@ -174,7 +174,7 @@ function! vimtex#compiler#output() abort " {{{1
       return
     endif
 
-    if mode() ==# 'v' || mode() ==# 'V' || mode() ==# ''
+    if mode() ==? 'v' || mode() ==# ''
       return
     endif
 
--- a/autoload/vimtex/compiler.vim
+++ b/autoload/vimtex/compiler.vim
@@ -174,6 +174,10 @@ function! vimtex#compiler#output() abort " {{{1
       return
     endif
 
+    if mode() ==# 'v' || mode() ==# 'V' || mode() ==# ''
+      return
+    endif
+
     " Try to enforce a file read
     execute 'checktime' self.name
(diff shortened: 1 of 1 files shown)


autoload/vimtex/compiler.vim
```diff
@@ -174,6 +174,10 @@ function! vimtex#compiler#output() abort " {{{1
       return
     endif
 
+    if mode() ==# 'v' || mode() ==# 'V' || mode() ==# ''
```
lervag: Why are you also checking `mode() ==# ''` here?

lervag: Also, you can combine the first two by using `==?`, that is: `mode() ==? 'v'`.

lervag:
Thanks! I'll merge this after we've settled my comments.

autoload/vimtex/compiler.vim
ishitaku5522: Thank you for your review.

> Why are you also checking `mode() ==# ''` here?

On Github, it looks empty string but this is actually "^V" (input by Ctrl+V Ctrl+V) which means Ctrl+V for visual block mode.
You can see it in raw souce code:   
https://raw.githubusercontent.com/ishitaku5522/vimtex/no_output_update_on_visual/autoload/vimtex/compiler.vim

> Also, you can combine the first two by using `==?`, that is: `mode() ==? 'v'`.

I updated.

ishitaku5522: Is "\<C-v>" better for you?

lervag: Yes, I think `"\<c-v>` is better (I prefer to not capitalize the letters, though). Thanks!

ishitaku5522:
Thank you for your waiting. I updated!

lervag:
Thank you for noticing, reporting and fixing the issue! :)

PR #1337, opened 2019-03-09
Declare mathpar to be a math zone

madgen:
The mathpar environment is defined in the mathpartir package for
typesetting inference rules and it is reasonably popular in programming
language circles. So it is a good default to include with vimtex.


lervag:
Seems reasonable. Thanks.

PR #1339, opened 2019-03-17
Add refcheck to g:vimtex_quickfix_latexlog

kiryph:
and order packages alphabetically


kiryph:
IMHO packages which add only warnings, such as refcheck, onlyamsmath, nag, widows-and-orphans, chkfloat, do not need a special switch. I also assume the messages follow a similar pattern.
I think it is better to add a general package switch for these packages.

lervag:
Sorry for the late replies! Been very busy lately!

It seems this PR was basically solved by #1341. Or is there anything more here that should be addressed?

kiryph:
I think having a general switch which covers the refcheck warning should be fine.

Nothing to be done here.

PR #1341, opened 2019-03-19
Add to 'efm' general package warnings

kiryph:
when using parser 'latexlog'. E.g. warnings by the package 'refcheck' are now shown.

However, others are still missing, e.g. following multi-line warning by onlyamsmath

```
Package onlyamsmath Warning: Environment eqnarray or eqnarray* used, please use
(onlyamsmath) only the environments provided by the amsmath
(onlyamsmath) package on input line 18.
```

### DIFF
--- a/autoload/vimtex/qf/latexlog.vim
+++ b/autoload/vimtex/qf/latexlog.vim
@@ -91,7 +91,9 @@ function! s:qf.set_errorformat() abort dict "{{{1
   "
   let l:default = self.config.packages.default
   if get(self.config.packages, 'natbib', l:default)
-    setlocal errorformat+=%+WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l%.
+    setlocal errorformat+=%+WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l.
+  else
+    setlocal errorformat+=%-WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l.
   endif
 
   if get(self.config.packages, 'biblatex', l:default)
@@ -126,6 +128,10 @@ function! s:qf.set_errorformat() abort dict "{{{1
   if get(self.config.packages, 'titlesec', l:default)
     setlocal errorformat+=%+WPackage\ titlesec\ Warning:\ %m
     setlocal errorformat+=%-C(titlesec)%m
+  endif
+
+  if get(self.config.packages, 'general', l:default)
+    setlocal errorformat+=%+WPackage\ %.%#\ Warning:\ %m\ on\ input\ line\ %l.
   endif
 
   " Ignore unmatched lines
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1656,6 +1656,15 @@ OPTIONS                                                        *vimtex-options*
   and have to be specifically disabled. However, it is possible to use the
   entry `disable` to disable either all warnings, or all package warnings, if
   so desired.
+  Warnings by following packages can be disabled individually:
+  `babel`, `biblatex`, `fixltx2e`, `hyperref`, `natbib`, `scrreprt`, and `titlesec`.
+  The entry `general` under packages can be used to turn off messages of
+  packages matching the pattern >
+
+    Package .* Warning: %m on input line %l.
+<
+  Note: Messages by `natbib` follow also this pattern but have to be turned
+        off with the entry `natbib`.
 
   Some examples may be helpful: >
 
@@ -1685,12 +1694,13 @@ OPTIONS                                                        *vimtex-options*
           \ 'font' : 1,
           \ 'packages' : {
           \   'default' : 1,
-          \   'natbib' : 1,
-          \   'biblatex' : 1,
+          \   'general' : 1,
           \   'babel' : 1,
-          \   'hyperref' : 1,
-          \   'scrreprt' : 1,
+          \   'biblatex' : 1,
           \   'fixltx2e' : 1,
+          \   'hyperref' : 1,
+          \   'natbib' : 1,
+          \   'scrreprt' : 1,
           \   'titlesec' : 1,
           \ },
           \}


autoload/vimtex/qf/latexlog.vim
```diff
@@ -91,7 +91,9 @@ function! s:qf.set_errorformat() abort dict "{{{1
   "
   let l:default = self.config.packages.default
   if get(self.config.packages, 'natbib', l:default)
-    setlocal errorformat+=%+WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l%.
+    setlocal errorformat+=%+WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l.
+  else
+    setlocal errorformat+=%-WPackage\ natbib\ Warning:\ %m\ on\ input\ line\ %l.
```
lervag: Do we need this? I thought the last line, `setlocal errorformat+=%-G%.%#`, ignores everything that is not otherwise matched...

lervag:
Please ignore my first message, as I did not quite understand what you wanted to achieve first. I think it looks good. Regarding the missing entries: As you might have noticed, most of the package related warnings have to be added manually. I'll probably add things if there are issues raised, and I'll accept PRs.

autoload/vimtex/qf/latexlog.vim
kiryph: Just to clarify: the new general rule would catch the natbib warning as well. By adding the `%-WPackage ...` entry, the line is explicitly excluded when the user has added to vimrc: 

    let g:vimtex_quickfix_latexlog = {'packages' : {'natbib' : 0} }

Otherwise, this line would have no effect.

I guess you got it now.

kiryph:
Regarding the missing warning by `onlyamsmath`:

I have tried to add a general rule for one-/two-/three line messages
```diff
+  if get(self.config.packages, 'general', l:default)
+    " one liner
+    setlocal errorformat+=%+WPackage\ %s\ Warning:\ %m\ on\ input\ line\ %l.
+    " two liner
+    setlocal errorformat+=%+WPackage\ %s\ Warning:\ %m%.
+    setlocal errorformat+=%-Z(%s)\ %m\ on\ input\ line\ %l.
+    " three liner
+    setlocal errorformat+=%+WPackage\ %s\ Warning:\ %m%.
+    setlocal errorformat+=%-C(%s)\ %m%.
+    setlocal errorformat+=%-Z(%s)\ %m\ on\ input\ line\ %l.
+  endif
+
``` 
However, I was not sure if this has side effects. It would catch the `onlyamsmath` warning:

Example:
```tex
\documentclass{article}
\usepackage{refcheck}
\usepackage{natbib}
\usepackage[warning]{onlyamsmath}

\begin{document}

\citet{super} % Triggers natbib warning; do not show when using vimtex by adding 
% let g:vimtex_quickfix_latexlog = {'packages' : {'natbib' : 0} } to your vimrc
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod
tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At
vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren,
no sea takimata sanctus est Lorem ipsum dolor sit amet.

\begin{equation}
   a=b\label{eq:your_equation_label} % Triggers refcheck warning
\end{equation}

$$a=b$$ %% triggers onlyamsmath warning

\end{document}
```

Quickfix list:
```
log-msg-line-no-with-linebreak.tex|15 warning| Package refcheck Warning: Unused label `eq:your_equation_label' on input line 15.
log-msg-line-no-with-linebreak.tex|18 warning| Package onlyamsmath Warning: Environment $$ $$ used, please use  only the environments provided by the amsmath package
```

kiryph:
It would be easy and possible sensible to add warning messages from packages which do not add a line number, e.g.
```
Package widows-and-orphans Warning: Display widow on page 2
Package widows-and-orphans Warning: Widow on page 3
Package widows-and-orphans Warning: Orphan on page 3
Package widows-and-orphans Warning: Widow on page 7 (second column)
Package widows-and-orphans Warning: Widow on page 8 (first column)
```
Two-liner without line numbers
```
Package nag Warning: Package a4wide is obsolete.
(nag) Use the `a4paper’ class option instead.
Package widows-and-orphans Warning: Orphan on page 5 (second column) and widow
(widows-and-orphans) on page 6 (first column)
Package widows-and-orphans Warning: Orphan and widow on page 8 (first and
(widows-and-orphans) second column)
```

This would be better than not showing them at all.

lervag:
I agree. If you're able to post a PR, I'll review and merge. Else I could consider to do it, but if so, please open a new issue/feature request with specification (or link here).

kiryph:
I would create a PR or maybe two:
1. show two/three line messages with line numbers as well
2. show package warnings messages without line numbers again for 1/2/3 lines.

However, I would like to add a few (vader) tests as well.
Are there already (vader) tests for verifying that parsing the log file works as expected which I could copy and adapt?

I have not yet written tests with vader or for vimtex at all which why this would be helpful.

Afaik, in the folder `test/issues` only example files exist which have to be run manually.

`test/features/error-messages` sounds about the right place but also looks like manual invocation.

Finally, `test/vader` seems not to contain any vader test for latex log file parsing.

I could not find easily in other vim plugins which use vader and might test for log file parsing and the results shown in the quickfix window:
- https://github.com/w0rp/ale/blob/master/test/test_go_to_definition.vader
- https://github.com/Squareys/vim-cmake/blob/master/test/cmake_quickfix.vader

lervag:
All the Vader tests are in the `test/vader` directory. There are none that tests the log parser, so that would also be a valuable contribution.

As inspiration, I think the `input_files.vader` and `get_main.vader` files might be useful, since they show some more complicated setups and test executions. You should be able to run the vader tests with the run script, e.g. `run logparser.vader` (or whatever you'll call the file).

PR #1346, opened 2019-03-24
Add basic fzf.vim integration

T-Rex96:
I adapted the [Denite.vim](https://github.com/lervag/vimtex/blob/master/rplugin/python3/denite/source/vimtex.py) source for use with [fzf.vim](https://github.com/junegunn/fzf.vim). Right now, this is just a basic implementation, which lists all the entries from `vimtex#parser#toc` and lets you jump to the right file and line number. I've never written in vimscript before, so please excuse my beginner mistakes..

Some stuff I'd like to add before this is merged:

- [x] Add colored output 
- [x] Filtering the output, e.g. only show labels or todo items
- [ ] ~Maybe add preview for equations? Does vimtex already have a way to do this?~

### DIFF
--- a/autoload/vimtex/fzf.vim
+++ b/autoload/vimtex/fzf.vim
@@ -4,11 +4,19 @@
 " Email:      karl.yngve@gmail.com
 "
 
-function! vimtex#fzf#run() abort " {{{1
+function! vimtex#fzf#run(...) abort " {{{1
+  " The filter argument may be used to select certain entry types
+  " according to the different "layers" of vimtex-toc:
+      " c:  content: This is the main part and the "real" ToC
+      " t:  todo: This shows TODOs from comments and `\todo{...}` commands
+      " l:  label: This shows `\label{...}` commands
+      " i:  include: This shows included files
+  " The default behavior is to show all entries, e.g. 'ctli'
+
   " The --with-nth 3.. option hides the first two words from the 
   " fzf window which we used to pass on the file name and line number
   call fzf#run({
-      \ 'source': <sid>parse_toc(),
+      \ 'source': <sid>parse_toc(a:0 == 0 ? 'ctli' : a:1),
       \ 'sink': function('vimtex#fzf#open_selection'),
       \ 'options': '--ansi --with-nth 3..',
       \})
@@ -22,7 +30,7 @@ endfunction
 
 " }}}1
 
-function! s:parse_toc() abort " {{{1
+function! s:parse_toc(filter) abort " {{{1
 " Parsing is mostly adapted from the Denite source
 " (see rplugin/python3/denite/source/vimtex.py)
 python3 << EOF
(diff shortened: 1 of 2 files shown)


lervag:
>  Maybe add preview for equations? Does vimtex already have a way to do this?

Not quite sure what you mean by this. I guess the answer to the second question is "no".

lervag:
I do not quite see how you actually use this with fzf.vim. Are you mapping something to `vimtex#fzf#run`?

T-Rex96:
> Thanks! The overall quality seems very good, although I do have some comments. Let me know if any of them are unclear.

Thanks!
> Also: You should mention the feature in the main docs in the NAVIGATION section, and if relevant/necessary, explain how to active/use this.

Yep, will do :)

> > Maybe add preview for equations? Does vimtex already have a way to do this?
> 
> Not quite sure what you mean by this. I guess the answer to the second question is "no".

E.g. when you have something like
```
\begin{equation}
  E = mc^2 \label{eq:einstein} 
\end{equation}
\begin{equation}
  R = \frac{U}{I} \label{eq:ohm} 
\end{equation}
```
the fzf entry could be
```
eq:bla                          E = mc^2
eq:ohm                          R = \frac{U}{I} 
```
> I do not quite see how you actually use this with fzf.vim. Are you mapping something to `vimtex#fzf#run`?
Yeah, I didn't include a default mapping, but you would just do something like
`nnoremap <localleader>lt :call vimtex#fzf#run()<cr>`

lervag:
Reg. preview: I would not mind at all if you added such a feature, but I would not deem it important. Let's merge this PR without that feature first, then possibly open a new one for that feature if you see a good way to implement it.

For the mapping: I propose that we leave it unmapped, but that you explain it in the docs (as was already an established todo).

T-Rex96:
> Reg. preview: I would not mind at all if you added such a feature, but I would not deem it important. Let's merge this PR without that feature first, then possibly open a new one for that feature if you see a good way to implement it.
> 
> For the mapping: I propose that we leave it unmapped, but that you explain it in the docs (as was already an established todo).

Sounds good! I pushed some commits that should address all of your comments. I also added some documentation :)

autoload/vimtex/fzf.vim
```diff
@@ -0,0 +1,92 @@
...
+    execute "normal! " . line . "gg" 
+  else
+    execute printf('edit +%s %s', line, file)
+  endif
+endfunction
+
+" }}}1
+
+function! s:parse_toc(filter) abort " {{{1
+" Parsing is mostly adapted from the Denite source
```
lervag: The comment should be indented. Same with  the following two lines. The python block may stay unindented, though.

lervag:
Thanks! I think this looks very good. I'll merge and apply a couple of minor, syntactical fixes myself.

autoload/vimtex/fzf.vim
T-Rex96: I think python blocks **have** to stay unindented, right? At least when I tried indenting the python block vim gave me a lot of errors.

lervag: Yes, that's true. I only meant that you should have indents for the initial "pure vimscript" code. See my latest commit.

Issue #1350, opened 2019-03-27
No module named 'colorama'


kiryph:
```
:call vimtex#fzf#run('c')
```
raises on my machine following error

```
|| Messages maintainer: Bram Moolenaar <Bram@vim.org>
|| "document.tex" 767L, 21644C
|| Error detected while processing function
|| vimtex#fzf#run[11]
|| <SNR>93_parse_toc[48]
|| Traceback (most recent call last):
||   File "<string>", line 3, in <module>
|| ModuleNotFoundError: No module named 'colorama'
|| Error detected while processing function
|| vimtex#fzf#run[11]
|| fzf#run[36]
|| E605: Exception not caught: Invalid source type
```

New fzf.vim source requires non-standard python module colorama.

The description of colorama says

> Makes ANSI escape character sequences (for producing colored terminal text and cursor positioning) work under MS Windows.

I am on macOS. Do I need this @T-Rex96 ?

I would prefer if we do not need this by default.

By the way this should be documented, e.g. in

```
FZF INTEGRATION                                                   *vimtex-fzf*

                                           https://github.com/junegunn/fzf.vim
                                               https://github.com/junegunn/fzf
|fzf.vim| integrates the general-purpose command-line fuzzy finder |fzf| into vim
and neovim. Similar to the |denite.vim| and |unite.vim| source it may be used to
quickly navigate |vimtex|'s built-in ToC feature. To use it, just define a
mapping to `vimtex#fzf#run()` in your .vimrc, e.g.: >

  nnoremap <localleader>lt :call vimtex#fzf#run()<cr>

You can also choose to only show certain entry "layers", according to this
table (see |vimtex-toc| for detailed explanation of the "layers"):

  `c`:  content
  `t`:  todo
  `l`:  label
  `i`:  include

The default behavior is to show all layers, i.e. 'ctli'. To only show
`content` and `label`s use: >

  :call vimtex#fzf#run('ct')
```

lervag:
I've pushed an update on this. What do you think, @kiryph? Does it work as expected? I.e.: Colorama is not necessary, but without it there are no colors.

T-Rex96:
Oh I thought Colorama was a default python package, my bad. We could also just switch to plain ansi escape sequences for the colors in that case.

kiryph:
@lervag: no, your recent commit does not avoid an error message if one does NOT have colorama.
The full messge is slightly different:

```
|| Error detected while processing function
|| vimtex#fzf#run[11]
[...]

PR #1352, opened 2019-03-28
Add colors via ansi escape codes when colorama is not installed, c.f. Issue #1350

T-Rex96:
I can't test this on Windows right now, but on my Linux install this seems to work fine.


lervag:
Thanks!

PR #1368, opened 2019-04-24
Add up to date suggestion for use vim-polyglot use

juliankrieger:
There are options to disable vim-polyglots latex part, which enabled me to use vimtex :)

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -3869,7 +3869,7 @@ This is a problem, because vim-polyglot [0] includes and enables LaTeX-Box by
 default. LaTeX-Box will not work well with vimtex, e.g. because of conflicting
 mappings.
 
-Suggested solution: Don't use vim-polyglot.
+Suggested solution: use |g:polyglot_disabled = ['latex'] in your .vimrc|
 
 [0]: https://github.com/sheerun/vim-polyglot
 


lervag:
I'm note sure I want to accept this. I strongly dislike polyglot and I believe people should avoid it. Still, I agree that your added information is useful. Perhaps we can reword it?

```diff
-Suggested solution: Don't use vim-polyglot.
+Suggested solution: Avoid it! Or use |g:polyglot_disabled| = ['latex'] in your `.vimrc` file.
```

juliankrieger:
@lervag I agree that rewording it is a possible choice. Maybe we could write it in a more neutral way, like

> ```diff
> -Suggested solution: Don't use vim-polyglot.
> +Suggested solution: Either use |g:polyglot_disabled| = ['latex'] in your `.vimrc` file or avoid it altogether.

On second note, could you explain why you dislike polyglot?
> ```

lervag:
Yes, you're suggestion is good.

> On second note, could you explain why you dislike polyglot?

The short answer: People should aim to know which plugins are installed in their Vim configuration.

Of course. The reason is simple: vim-polyglot installs a whole bunch of filetype plugins, such as LaTeX-Box. This implies that a user is provided with a lot of plugins of varying quality. It seems tempting, because one gets a lot of features "for free". The main problem is that the user is not aware of which plugins are actually used. I've personally spent quite a lot of time debugging strange issues due to people trying to use vimtex with polyglot enabled. Of course, today I recognize this quite fast (I've even added a check for it).

Instead, I firmly believe that one should do the extra work of adding custom filetype plugins manually. This ensures at least a minimum of awareness. I also claim that the extra work is not really that much. Of all the extra plugins provided by vim-polyglot, I believe that most people really only use a few.

juliankrieger:
@lervag You're absolutely correct, I just checked how many plugins are installed by vim-polyglot, and it seems like an awful lot. I've gone ahead and uninstalled it, replacing it by the exact same syntax plugins it uses, but only for the filetypes I actually encounter.

Thanks!

lervag:
@juliankrieger I'm happy to hear I actually manage to convince people on this point. I understand why a lot of people feel like installing polyglot, and I can understand why my point sometimes feels too "strict" or "conservative".

In any case, I still think that your suggested update is relevant, and so I've updated the docs accordingly. I added some more text in an attempt to better justify my recommendation.

PR #1309, opened 2019-01-17
Add document to README and help about `tectonic`

ta2gch:
This request is a derived from #1280. I added some document to that request as you said.
I can add more document if you say, then tell me anything about your request, please. Thank you.

PR #1381, opened 2019-05-13
Combine all changes for tectonic

fgborges:
This PR is a revised version of #1309 .
I combine all changes at #1309 into one commit.

@doronbehar I grand you to edit my repository, so please check it.
The branch name is **tectonic**, not `tectonic_support` and `origin/tectonic_support`.


doronbehar:
According to [@lervag 's comment in #1309](https://github.com/lervag/vimtex/pull/1309#issuecomment-491630460), this is our TODO in my view:

- [X] Make errors / warnings from `tectonic` reach the quickfix list (should be very similar to whatever that reads XeLaTex chatter).
- [ ] Make `g:vimtex_view_automatic` actually make this backend open the viewer `on_exit` callback.
- [ ] Implement `s:compiler.clean` by deleting synctex data and logs.
- [ ] Make sure documentation corresponds to actual behaviour:
  * [ ] Update default compiler definition.
  * [ ] Mention that if `--keep-logs` or `-k` is not used, errors and warnings won't be loaded to quick fix.
  * [ ] Update notes on clean action for tectonic - saying it will delete only logs and synctex.

lervag:
Let me know when you want me to look at things.

doronbehar:
@lervag , my latest commits handle most of what I've been put in the TODO list. I haven't completed yet the documentation since that may need further changes if you'll decide after a review that the behaviour implemented should be changed.

Besides that, I'm not sure I understand what you wrote in [your last comment in #1309](https://github.com/lervag/vimtex/pull/1309#issuecomment-491630460), quoting:

>> Make the viewer open on_exit
>
> I think it makes sense to add support for g:vimtex_view_automatic. The idea is to couple this with the callback after successful or failed compilation. This is done for arara e.g. with the s:callback(ch, msg) function for Vim jobs feature, or s:callback_nvim_exit(id, data, event) function for neovim. The callback is coupled to the viewers, and the viewer method must take action to open a viewer if appropriate.

Does it mean that the actual command that will be used to open the viewer should be handled in the domain of the compiler backend or is it being invoked by some other function in the `vimtex#` domain? From reading your answer, it seems the later is right. From reading the code it seems the former is right. Referring to what you've said about the arara backend, I'll quote the code:

https://github.com/lervag/vimtex/blob/6c00a1dd07df9866811a765a01bc9a29060e2f16/autoload/vimtex/compiler/arara.vim#L176-L178

lervag:
https://github.com/lervag/vimtex/blob/6c00a1dd07df9866811a765a01bc9a29060e2f16/autoload/vimtex/compiler/arara.vim#L176-L178

Here the `vimtex#qf#inquire(s:cb_target)` will check the status of the compilation through the LaTeX log. The `vimtex#compiler#callback` function will open the viewer if `g:vimtex_view_automatic = 1` and if the viewer supports this. It should work for Zathura, MuPDF, and Skim, I believe. The confusion probably lies in the complexity in that the viewers are only interfaced within compiler callback here:

https://github.com/lervag/vimtex/blob/6c00a1dd07df9866811a765a01bc9a29060e2f16/autoload/vimtex/compiler.vim#L83

Because the viewer upon initialization adds a callback, here:

https://github.com/lervag/vimtex/blob/6c00a1dd07df9866811a765a01bc9a29060e2f16/autoload/vimtex/view.vim#L46

So, to answer

> Does it mean that the actual command that will be used to open the viewer should be handled in the domain of the compiler backend

Yes. Let me know if you want me to provide a minimal example that shows when this feature is actually activated, e.g. with the arara backend. I think it should suffice to use `g:vimtex_view_automatic = 1` and a supported viewer for this to work.

autoload/vimtex/doc.vim
```diff
@@ -235,7 +235,7 @@ function! s:packages_open(context) abort " {{{1
         \         ? '!xdg-open'
         \         : (l:os ==# 'mac'
         \            ? '!open'
-        \            : '!start'))
+        \            : '!start /b'))
```
lervag: I guess this is here because the branch is not rebased, right? Please avoid merging master into the branch, as I want to keep things clean. Just saying. :)

doc/vimtex.txt
```diff
@@ -3526,6 +3585,32 @@ The latter is supported by vimtex, though, through the callback.  See
...
 The compiler may be configured through the |g:vimtex_compiler_latexrun| option.
 
+------------------------------------------------------------------------------
+TECTONIC                                                      *vimtex-tectonic*
+
+                                      https://tectonic-typesetting.github.io/
+> Tectonic is a modernized, complete, self-contained TeX/LaTeX engine, powered
+> by XeTeX and TeXLive.
+>
+Tectonic is complete, self-contained TeX/LaTeX engine, powered by XeTeX and
```
lervag: Feels strange to repeat the quote (_almost_) directly in the first sentence. And minor typo: "is _a_ complete".

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,263 @@
...
+endfunction
+
+" }}}1
+
+function! s:compiler.build_cmd() abort dict " {{{1
+  let l:cmd = 'tectonic'
+
+  for l:opt in self.options
+    if l:opt =~ '^\-\-outdir[ =]' || (l:opt =~ '^-o')
+      call vimtex#log#warning("Don't use --outdir or -o in compiler options, use build_dir instead")
```
lervag: I prefer to keep all lines within 80 columns. I know it is inconvenient for long strings like this. Should be simple to break like this, e.g.:


```vim
      call vimtex#log#warning("Don't use --outdir or -o in compiler"
            \ . " options, use build_dir instead")
```

lervag: I notice that this "rule" is broken more places. I'll not comment on each line, instead, let this comment count for all long lines.

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,263 @@
...
+
+  return l:list
+endfunction
+
+" }}}1
+
+function! s:compiler.clean(...) abort dict " {{{1
+  if !get(self, 'saving_synctex') && !get(self, 'saving_logs')
+    " TODO: add relevant section in doc
+    call vimtex#log#warning('Nothing to clean since vimtex is configured to run tectonic without keeping intermidiets. See :help tectonic-intermidiets')
```
lervag: "Intermediate", not "Intermidiet"

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,263 @@
...
+    return
+  endif
+  let l:target_basename = self.build_dir . "/" . fnamemodify(self.target_path, ':t:r')
+  " TODO: should we remove the log and the pdf output as well?
```
lervag: The "banged" version, i.e. `:VimtexClean!`, should remove the PDF as well.

doronbehar: O.K, While reading the source code in order to see how should I check for the use of a bang, I've encountered this function:

https://github.com/lervag/vimtex/blob/21c9f5cc0e0dafad60d3cb4db457484586cca781/autoload/vimtex/compiler.vim#L238-L256

Why exactly does it do? I noticed it calls `b:vimtex.compiler.build_dir` and if the `build_dir` of the compiler is empty, nothing more is done. However, in our case, `build_dir` is never empty so what exactly more does it do? it seems it actually deletes all readable directoris there using `glob()` but I didn't see that actually happen according to my tests.

I've made a check in `s:compiler.clean` for `a:0` and I've made it delete the PDF and `log` files as well in case it is true.

doronbehar:
> So, to answer
> 
> > Does it mean that the actual command that will be used to open the viewer should be handled in the domain of the compiler backend
>
> Yes. Let me know if you want me to provide a minimal example that shows when this feature is actually activated, e.g. with the arara backend. I think it should suffice to use `g:vimtex_view_automatic = 1` and a supported viewer for this to work.

I wouldn't have been so perplexed unless I had noticed that arara **and** latexrun don't open the viewer automatically, at least with my . I've even set `g:vimtex_view_automatic = 1` explicitly in my `init.vim` and still they don't open it.

However, looking at the source code at the rest of vimtex, I think that my viewer is never opened automatically because I always have some warnings with my documents. I guess tectonic is very strict..

Anyway, I leave it the way it is now. The only thing left now as far as I can see is tests and documentation.

@ta2gch, I'm pushing with `--force` in order to rebase from master.

doc/vimtex.txt
doronbehar: I hope you'll like what I've wrote, see 81e7dff371bcc1de34831d55145898b006066dd1.

autoload/vimtex/compiler/tectonic.vim
lervag: > However, in our case, build_dir is never empty so what exactly more does it do?

No, that's not right. The point here is that `empty('')` evaluates to true, and in many cases, the `build_dir` setting is set to an empty string. The part that follows will remove the build directory if it is empty.

lervag:
I'll look more closely at the updates later today.

autoload/vimtex/doc.vim
lervag: This change is not relevant to the PR, and I suspect it is simply wrong. Can you confirm?

doc/vimtex.txt
```diff
@@ -964,6 +976,63 @@ OPTIONS                                                        *vimtex-options*
...
+    modify these, a few notes should be given:
+    - Don't use `--outdir` or `-o` here. Use the `build_dir` option instead.
+    - Without `--keep-logs`, you won't see errors / warnings on your Quick Fix
+      when compilation will finish.
+    - By default, `tectonic` cleans all kinds of intermediate files such as
+      `.aux`, `.toc` etc. If you modify the options and omit all options that
+      save any of these files, |<plug>(vimtex-clean)| and |VimtexClean| Won't
+      do anything and a warning will be printed.
+
+*g:vimtex_compiler_latexrun_engines*
```
lervag: I don't think this option is used anywhere, and so it should be removed. Or did I miss it?

lervag: Oh, sorry, was this just adding docs to an option that was not documented?

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,275 @@
...
+        \ l:opt =~ '^--keep-intermediates$'
+      let self.saving_logs = 1
+    endif
+    if l:opt =~ '^--synctex$'
+      let self.saving_synctex = 1
+    endif
+    let l:cmd .= ' ' . l:opt
+  endfor
+  if !get(self, 'saving_synctex')
+    call vimtex#log#warning("--synctex wasn't used in compiler options so this"
```
lervag: Here and below you are adding warnings when the user has changed from default options. In a way, I agree it makes sense, but at the same time, perhaps the user actually did it on purpose and does not want the warnings?

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,275 @@
...
+  let l:cmd = 'tectonic'
+
+  for l:opt in self.options
+    if l:opt =~ '^\-\-outdir[ =]' || (l:opt =~ '^-o')
+      call vimtex#log#warning("Don't use --outdir or -o in compiler options,"
+          \ . " use build_dir instead, see :help g:vimtex_compiler_tectonic"
+          \ . " for more details")
+      continue
+    endif
+    if l:opt =~ '^--keep-logs$' || l:opt =~ '^-k$' ||
```
lervag: Since you are parsing the options in the `self.options` list, you could as well just add the keys directly? It would be slightly cleaner, I think. I.e., allow the `g:vimtex_compiler_tectonic` dictionary to have `keep_logs` and `synctex` as keys with sane defaults, and then instead check if any extra options are added that are redundant similar to the `-outdir` option?

lervag:
To me it seems things are quite good now. I had one more round of comments, but I think it may be ready for merge after that. Good work!

autoload/vimtex/doc.vim
doronbehar: Confirm. Must have got cherry-picked by @ta2gch. Removed.

autoload/vimtex/compiler/tectonic.vim
doronbehar: Right.. I guess that makes sense considering the default options are documented shall I change these to `#log#info` or remove them all together?

autoload/vimtex/compiler/tectonic.vim
doronbehar: >  allow the g:vimtex_compiler_tectonic dictionary to have keep_logs and synctex as keys with sane defaults [..]

I'm not sure I understand that, are you referring to `self.saving_logs` and `self.saving_synctex`? The idea is that no matter what the user added or didn't to the compiler's options, the options will be parsed anyway. Naturally, with the defaults or with similar options set, these keys will be set appropriately.

I'm not sure I'm even capable to see if `self.options` was modified or not by the user inside this function.

doronbehar:
Thanks again @lervag for your comments. Besides our discussion above, I'm still unsure as for the tests.

How exactly do you run them? I'd like to test all of the scenarios in my [TODO comment](https://github.com/lervag/vimtex/pull/1381#issuecomment-491711012) but I'm not sure exactly how to mimic them.

doc/vimtex.txt
lervag: Did you miss this question?

autoload/vimtex/compiler/tectonic.vim
lervag: I would remove them alltogether, I think. Let's assume the user knows what he/she is doing.

autoload/vimtex/compiler/tectonic.vim
lervag: No, sorry, I misunderstood. My bad. However, as related to my above comment, I would remove the warnings, which means you may also simplify a lot of the code by removing the `self.saving_logs` and `self.saving_synctex` stuff.

lervag:
Very minor comment: When looking over things now I also noticed that there are some "end-of-line spaces". It would make me happy if you remove these (I know, I'm pedantic and silly).

Regarding tests: My initial suggestion would be to create the files:

```
VIMTEX/test/issues/1381
├── minimal.tex
└── minivimrc
```

Use `VIMTEX/test/issues/template` as an example. In `minivimrc`, add some configuration examples for the relevant options.

I know this is not a very satisfying test, but it is a good step. I use Vader for tests that are run on the master branch, but these tests do not run any actuall LaTeX commands, and as such, do not really work for testing the compiler backends.

Another possibility is to update the `VIMTEX/test/features/compiler/minivimrc`. It creates a suite of tests for the various compiler backends that I can run manually and that tests if `VimtexCompile` generates a PDF, if it can be stopped, etc.

doc/vimtex.txt
doronbehar: Oh now I got you. Again, don't know how this got added to the PR. Would you like to include it anyway? Probably @ta2gch added this but I'm not sure why.

autoload/vimtex/compiler/tectonic.vim
doronbehar: Oh right, as you wish. I assume your philosophy for this kind of dilemmas is RTFD?

doc/vimtex.txt
```diff
@@ -964,6 +976,64 @@ OPTIONS                                                        *vimtex-options*
...
+    Same as |g:vimtex_compiler_latexmk| / `build_dir`. This will be created if
+    needed, since tectonic won't build anything unless it exists.
+
+  options~
+    This is a list of options that are passed to `tectonic`. The default
+    options should work well for most people. But for anyone who wishes to
+    modify these, a few notes should be given:
+    - Don't use `--outdir` or `-o` here. Use the `build_dir` option instead.
+    - Without `--keep-logs` or alternatively `--keep-intermediates` (or `-k`,
+      you won't see errors / warnings on your Quick Fix when compilation will
```
lervag: Missing end paranthesis.

doc/vimtex.txt
```diff
@@ -964,6 +976,64 @@ OPTIONS                                                        *vimtex-options*
...
+      finish.
+    - By default,`tectonic`cleans all kinds of intermediate files such as
```
lervag: Missing space between `tectonic` and "cleans".

lervag: And after the comma.

lervag: And in the continuation there are more missing spaces...

autoload/vimtex/compiler/tectonic.vim
```diff
@@ -0,0 +1,258 @@
...
+    call add(l:list, ['cmd', self.cmd])
+  endif
+
+  return l:list
+endfunction
+
+" }}}1
+
+function! s:compiler.clean(...) abort dict " {{{1
+  if !get(self, 'saving_synctex') && !get(self, 'saving_logs') && !a:1
```
lervag: I think this can be removed as well.

autoload/vimtex/compiler/tectonic.vim
lervag: Yes, sort of. And also that for users who know what they are doing, they should not get unnecessary warning messages.

lervag:
The test folder `test/issues/1381`: You should remove the `profile.vim` file, and the `minivimrc` should be updated so that it is relevant. I want to do: `nvim -u minivimrc minimal.tex` and then be able to test the tectonic feature with default options, as a minimum.

doronbehar:
O.K, now I got you. I get the error:

```
E492: Not an editor command: VimtexCompile
```

Do I need to set my `$VIMRUNTIME` or something like that? I have vimtex installed in `~/.config/nvim/pack/filetype/start/tex`.

lervag:
Ok, I've now taken the liberty of making some minor cleanup and merging this. I have not really tested it, as I have still not been able to make tectonic run on my machine, so please open issues (or PRs) if you happen to find bugs!

Issue #1308, opened 2019-01-15
Text objects as a 'plugin' for targets.vim if installed


kiryph:
This issue started in #1299. Since the issue 1299 is actual about new motions and the discussion has become a bit lengthy I repeat a few things:

https://github.com/wellle/targets.vim/blob/master/plugins.md says

> It's possible to extend targets.vim by adding new sources. Text objects implemented as targets.vim plugin will automatically gain all the functionality of the built in ones. Most notably you can
> - use `n` and `l` to select next and last one,
> - provide counts to pick a specific one,
> - use seeking (`vix` pick current, next or last),
> - growing (`vixix` is the same as `v2ix` etc.) and
> - skipping (`vinxinx` is the same as v2inx).

Furthermore, it is possible to define `I` and `A` variants to exclude inner surrounding whitespace or include trailing/leading whitespace.

I have created a small snippet reusing vimtex definition for the text object `ie` to get `ane/ale/ine/ile` and `Ie` and `Ae`, ...:

`~/.vim/ftplugin/tex.vim`:
```viml
autocmd User targets#sources call targets#sources#register('tex_env', function('vimtextargets#envnew'))
autocmd User targets#mappings#plugin call targets#mappings#extend({'e': {'tex_env': [{'c': 1}]}})
```
`~/.vim/autoload/vimtextargets.vim`:
```viml
function! vimtextargets#envnew(args)
    return {
                \ 'args': {
                \     'count': get(a:args, 'c', 1),
                \ },
                \ 'genFuncs': {
                \     'c': function('vimtextargets#envcurrent'),
                \     'n': function('vimtextargets#envnext'),
                \     'l': function('vimtextargets#envlast'),
                \ },
                \ 'modFuncs': {
                \     'i': function('targets#modify#keep'),
                \     'a': function('vimtextargets#modifyouterenv'),
                \     'I': function('targets#modify#shrink'),
                \     'A': [function('vimtextargets#modifyouterenv'), function('targets#modify#expand')],
                \ }}
endfunction

function! vimtextargets#envcurrent(args, opts, state)
    execute "normal \<Plug>(vimtex-ie)"
endfunction

function! vimtextargets#envnext(args, opts, state)
    execute "normal \<Plug>(vimtex-]m)v\<Plug>(vimtex-ie)"
endfunction

function! vimtextargets#envlast(args, opts, state)
    execute "normal \<Plug>(vimtex-[M)v\<Plug>(vimtex-ie)"
endfunction

[...]

PR #1384, opened 2019-05-16
[RFC] Add text object variant 'targets'

kiryph:
@lervag 
I tried to implement my feature request in issue https://github.com/lervag/vimtex/issues/1308 for commands and environments. Any feedback is appreciated.

~~Also note my change to the current behavior regarding counts which I describe below to wellle. Maybe this change is not necessary.~~ This is resolved.

@wellle
I have a question about counts for text objects regarding targets.vim and this pull request.

When I'd created this pull request, I had to modify the re-used text objects from vimtex as following:

https://github.com/lervag/vimtex/commit/9c221a031dee25a9be10f0ffedbb6c3d40034429 (separate commit for these modifications)

Prior to my modification, you could give a large count to the text object which could have been larger than there is a corresponding text object. vimtex would simply iterate until there is no surrounding text object anymore. This made it possible to press something like `v99ae` to select the most outer environment (assuming there are not more than 99 nested environments).

I had to disable it. Otherwise pressing `vaeae` for a non-nested LaTeX environment would result in an infinite loop in following part of targets.vim (https://github.com/wellle/targets.vim/blob/master/autoload/targets/multigen.vim#L40-L56:)

```viml
    while 1
        let [target, idx] = s:bestTarget(targets, self.context, 'multigen')
        if target.state().isInvalid() " best is invalid -> done
            let self.currentTarget = target
            return self.currentTarget
        endif

        " if two generators produce the same target, skip it
        " also used for growing in some cases
        if self.lastRawTarget.equal(target)
            let targets[idx] = target.gen.next(0)
            continue
        endif

        let self.currentTarget = target
        return self.currentTarget
    endwhile
```

Now pressing `v99ae` does nothing.

Are there technical reason that targets.vim cannot be gracious to too large counts or am I simply wrong and did not get it right?

I was considering the old behaviour of vimtex a feature.


wellle:
@kiryph: targets.vim could behave in the way you describe, but currently doesn't for compatibility reasons. Let's look at how Vim behaves, start Vim without config/plugins:
```
vim -u NONE --noplugin
```
Paste this text into the buffer
```
a ( b ) ( c ) ( ( x ) ) ( e ) ( f ) g
```
and put your cursor on the `x`. Now if you type `vib` it selects the inner block, if instead you type `v2ib` it selects in the outer block, but if instead you type `v3ib` or any bigger number, then nothing will be selected. Similarly try `dib`, `d2ib` and `d3ib`.

So targets.vim just follows Vim behaviour here.

That being said, feel free to [open an issue in the targets.vim repo](https://github.com/wellle/targets.vim/issues/new) to make it possible to allow this "gracious" behaviour you described.

kiryph:
@wellle 
Thank you for your quick reply and your interest in this.

> So targets.vim just follows Vim behaviour here.

yes, I already noticed this.

It sounds like you are willing to change this in targets.vim anyway. So I will open an issue in targets.vim.

Since targets.vim reimplements the builtin text objects, some may actually find this behavior correct (as did you). Personally, I do not see a reason for this but I suspect others like it this way.
Let's see if others get angry :-).

wellle:
Hey, thanks for opening the issue 👍 

> Let's see if others get angry :-).

I won't change the default behaviour though. I'll just consider adding a setting to change the behaviour either for all targets or on a per target basis.

autoload/vimtex/text_obj/cmdtargets.vim
```diff
@@ -0,0 +1,109 @@
+" vimtex - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+"
+
+function! vimtex#text_obj#cmdtargets#new(args) " {{{1
+    return {
+                \ 'args': {
+                \     'count': get(a:args, 'c', 1),
```
wellle: Seems like this was taken from the line-targets example. Since `args.count` is never used, you can remove this line.

autoload/vimtex.vim
```diff
@@ -536,6 +528,38 @@ function! s:init_default_mappings() abort " {{{1
...
+          \ get(g:, 'vimtex_text_obj_variant', 'auto') == 'auto' ||
+          \ get(g:, 'vimtex_text_obj_variant', 'auto') == 'targets'
+          \ )
+      augroup targets_tex
+        autocmd User targets#sources call targets#sources#register(
+              \ 'tex_env', function('vimtex#text_obj#envtargets#new'))
+        autocmd User targets#sources call targets#sources#register(
+              \ 'tex_cmd', function('vimtex#text_obj#cmdtargets#new'))
+        autocmd User targets#mappings#plugin call targets#mappings#extend(
+              \ {'e': {'tex_env': [{'c': 1}]}})
```
wellle: And then you also don't need to set the unused `c` here.
```suggestion
              \ {'e': {'tex_env': [{}]}})
```

autoload/vimtex.vim
```diff
@@ -536,6 +528,38 @@ function! s:init_default_mappings() abort " {{{1
...
+        autocmd User targets#mappings#plugin call targets#mappings#extend(
+              \ {'c': {'tex_cmd': [{'c': 1}]}})
```
wellle: And here.

autoload/vimtex/text_obj/envtargets.vim
```diff
@@ -0,0 +1,120 @@
+" vimtex - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+"
+
+function! vimtex#text_obj#envtargets#new(args) " {{{1
+    return {
+                \ 'args': {
+                \     'count': get(a:args, 'c', 1),
```
wellle: Same here.

autoload/vimtex/text_obj/cmdtargets.vim
kiryph: Thansk for the suggestions.
Yes, this stems from line-targets.vim The other parts of `envtargets.vim` and `cmdtargets.vim` are based on https://github.com/wellle/targets.vim/blob/master/autoload/targets/sources/pair.vim

I force-pushed the changes.

kiryph:
I have updated this PR as follwoing:

- textobjects variant 'vimtex' (default) are again always gracious regarding too large counts (dropped corresponding commit)
- textobjects variant 'targets' are either gracious or not dependent on the variable `g:targets_gracious` [recently added to targets.vim](https://github.com/wellle/targets.vim/commit/a79447f261e4b8b4327557aa03726f3849334b84)

I am not sure whether I will find the time to look into
- `i$/a$` (assuming more complicated due to symmetric delimiters),
- `id/ad` (assuming more complicated because this is actually a multi-object) and
- `iP/aP` (rarely used by myself)
to provide also variants based on targets.vim.

Furthermore, vader tests are missing for this PR. However, they would require that targets.vim is available for testing. Not sure what the best way to handle this would be.

So, I will let this PR standing as it is.

lervag:
I'm sorry for taking so long to review and comment. Life's busy with work and kids. For now I just want to ask: Does this PR make #1308 redundant so that I may close it and let the discussions continue here?

kiryph:
Yes, we can close the issue #1308  and continue discussions here.

lervag:
Ok, again, sorry for the major delay! I've finally taken time to review this properly.

First, thanks for creating this! I generally don't find anything wrong with your code, and I think I want to merge it quite soon. I have two minor comments/questions that I'll add inline.

Finally, regarding testing: I find it difficult to add Vader tests for this. My pragmatism makes me think that it would be easier to create a `test/features/targets.vim/...` folder with a minimal vimrc file, an example LaTeX file, and a Makefile. That would make it easy to do manual testing. What do you think?

autoload/vimtex.vim
```diff
@@ -545,9 +538,44 @@ function! s:init_default_mappings() abort " {{{1
...
+
+    if exists('g:loaded_targets') && (
+          \    (type(g:loaded_targets) == type(0)  && g:loaded_targets)
+          \ || (type(g:loaded_targets) == type('') && !empty(g:loaded_targets))
+          \ ) && (
+          \    g:vimtex_text_obj_variant == 'auto'
+          \ || g:vimtex_text_obj_variant == 'targets'
+          \ )
+      let g:vimtex_text_obj_variant = 'targets'
+      augroup targets_tex
```
lervag: 1. Should we not use a `autocmd!` at the top?
2. And/Or: Should we use `<buffer>` patterns here?

lervag:
Also, thanks to @wellle for pitching in!

autoload/vimtex.vim
kiryph: Yes, I agree that there should be an `autocmd!` at the top.

And you are probably right about using `<buffer>`-local autocommands (`:h autocmd-buffer-local`).
I was following too closely the example by wellle which was not a ft-specific object.

kiryph:
> My pragmatism makes me think that it would be easier to create a test/features/targets.vim/... folder with a minimal vimrc file, an example LaTeX file, and a Makefile. That would make it easy to do manual testing. What do you think?

Sounds good to me. Some way of testing is better than none.

autoload/vimtex.vim
lervag: Good. Although I found that it seems `User` autocmds are never `<buffer>` local. So we only need to ensure they are not defined more than once.

lervag:
Ok, I've fixed the autocmd stuff and added a test file. To me, it seems things are working, except: For the nested command examples, it seems the `ilc` stuff is not working as indicated. Can you test by pulling the latest version of this branch, then going to the test directory and open with `nvim -u minivimrc`, going to the nested command tests and try e.g. `vilc`?

kiryph:
> Can you test by pulling the latest version of this branch, then going to the test directory and open with nvim -u minivimrc, going to the nested command tests and try e.g. vilc?

I did it (except that I still use vim instead of nvim):

I can confirm that `vilc` does **not** work for nested commands but also for the non-nested ones (line 5 in minimal.tex).

Can you check this as well?

I had a look at the `function! vimtex#text_obj#cmdtargets#last(args, opts, state)` in `autoload/vimtex/text_obj/cmdtargets.vim`:
```viml
function! vimtex#text_obj#cmdtargets#last(args, opts, state) " {{{1
  if targets#util#search('\\\S*{', 'bW') > 0
    return targets#target#withError('no target')
  endif

  let oldpos = getpos('.')
  let target = s:select(1)
  call setpos('.', oldpos)
  return target
endfunction
```

 It appears to me that the backwards search places the cursor on the backslash of the current command and then runs `s:select(1)` which invokes `keepjumps normal v'.a:count."\<Plug>(vimtex-ac)v`. The cursor is after the backwards search still in the starting region (marked with `^^^^` in the test file and cheatsheet) and therefore the specifier `l` has no effect.

Could you test following modification?

```diff
❯ g diff
diff --git a/autoload/vimtex/text_obj/cmdtargets.vim b/autoload/vimtex/text_obj/cmdtargets.vim
index 15926f2..b63755f 100644
--- a/autoload/vimtex/text_obj/cmdtargets.vim
+++ b/autoload/vimtex/text_obj/cmdtargets.vim
@@ -43,7 +43,13 @@ endfunction
 " }}}1

 function! vimtex#text_obj#cmdtargets#last(args, opts, state) " {{{1
-  if targets#util#search('\\\S*{', 'bW') > 0
+  " Normalize cursor position of current cmd: place on leading '\'
+  if targets#util#search('\\\S\+{', 'bWc') > 0
+    return targets#target#withError('no target')
+  endif
+
+  " Move to the last non-surrounding cmd
+  if targets#util#search('\\\S\+{.\{-}}', 'bW') > 0
     return targets#target#withError('no target')
   endif
```

Further issues

- commands split over several lines. Consider following change in minimal.tex by splitting in line 5 the `c` cmd
   ```tex
   a \cmd{ bbbbbbbb }  \cmd{
   ccccccc }  \cmd{ dddddd }  \cmd{ eeeeeee } \cmd{ ffffffff }
   
   %           ^^^^^^^^^^^^^^
   ```
   Pressing `valc` when on the `d` cmd selects the `b` cmd instead of the splitted `c` cmd.

- Gitgutter users which do not know that `ic/ac` are defined by gitgutter might notice that the `ic` and `ac` text object of the variant 'targets' do not work. Ftplugins overwrite gitgutter (i.e. text object variant `vimtex`)  but not when the text objects are defined in the way for targets.vim.

lervag:
> I can confirm that vilc does not work for nested commands but also for the non-nested ones (line 5 in minimal.tex).

For me, it almost works for the non-nested example. If I repeat the `ilc`, it works, but the first invocation selects the current instead of the previous text object.

> Could you test following modification?

Tested, and it seems to fix the first tests.

> Further issues

I can reproduce your first issue. Did you by any chance have a solution for this? I could investigate if not.

Regarding gitgutter: Could we perhaps detect this case in the initialization code?If `ic/ac` are mapped to gitgutter, it could be safe to assume that we can unmap before running targets.vim initialization code?

lervag:
Regarding the new issue, what do you think of the following patch:

```diff
--- a/autoload/vimtex/text_obj/cmdtargets.vim
+++ b/autoload/vimtex/text_obj/cmdtargets.vim
@@ -49,7 +49,7 @@ function! vimtex#text_obj#cmdtargets#last(args, opts, state) " {{{1
   endif

   " Move to the last non-surrounding cmd
-  if targets#util#search('\\\S\+{.\{-}}', 'bW') > 0
+  if targets#util#search('\\\S\+{\_.\{-}}', 'bW') > 0
```

lervag:
Ok, I've fixed some of the remaining edge cases. It's not perfect, but I think it is close to good enough.

kiryph:
I noticed a regression with your recent commits:

Pressing `vIc` should be the same as `vic` for `\cmd{ffffffff}`. However, I get following selection

Screenshot 2019-08-13 at 09 10 50

Can you confirm this?

wellle:
> but not when the text objects are defined in the way for targets.vim

Do you think there’s anything targets.vim could/should do differently?

kiryph:
@wellle 

> Do you think there’s anything targets.vim could/should do differently?

The first thing which comes to my mind is for neovim users: provide a health report (nvim: `:h health-dev`) called via `:checkhealth` which could compare `:nmap i/a` to the registered targets and report about those with a warning which are defined twice. Of course, in the case for vimtex, the user has to run `:checkhealth` from a tex file.

> If ic/ac are mapped to gitgutter, it could be safe to assume that we can unmap before running targets.vim initialization code? (lervag)

I would not be sure. If the user has not set the variable `g:vimtex_text_obj_variant` explicitly to `targets', he/she might still want to use ic/ac from gitgutter and not vimtex. Even though I think this is the minority of users of vimtex. ic/ac is probably one of the more often used text objects.

But for the matter of consistency I would actually unmap them because this happens for the variant 'vimtex'.

Maybe vimtex's health report could report on this unmapping or clash in general as well.
I learned that a clean installation of plugins should not throw any messages.

I still would add some documentation how to configure vimtex that it does not overwrite vim-gitgutter. I see two options

1. disable them in vimtex (not all text objects)
2. pick alternatives in vimtex

lervag:
> Can you confirm this?

Yes, I can. It seems the `targets#modify#shrink` function does not work as expected with input `aaaaaa`, as it then gives output `aaaa`.

I'm not sure if we should fix by making our own variant, or if it should be considered a bug with `targets.vim`. I'll let @wellle decide on that.

> I still would add some documentation how to configure vimtex that it does not overwrite vim-gitgutter. I see two options
>
>  1.  disable them in vimtex (not all text objects)
>  2.  pick alternatives in vimtex

Personally, I'm all for the "simple" and consistent options. So I'll go for option 1 and disable them (just the conflicting ones), so as to be consistent with the previous ones.

----

By the way, I've noticed that the new inner command text objects with `targets.vim` are not the same as the original vimtex ones. Notice that vimtex will select the macro name, not the macro argument. I am not sure what is the most expected behaviour. Selecting the inner part of the first argument may be a consistent and expected behaviour, but LaTeX macros may sort of randomly have zero or more arguments, including optional arguments. I think the only two options are to consider the macro name (as in the original version) or to consider the first argument.

lervag:
Ok, I've tried to unmap them, but I was not able to do it properly. Perhaps it would be possible to add an option to targets.vim to force the map somehow?

lervag:
Ok, I've found a way. Quite of a hack, but it seems to work. I'd be happy to get some feedback on this.

Btw, I noticed another bug: Doing `vic` at the first `a` on line 4 selects the `document` of line 2. This seems very wrong.

wellle:
Thanks, I'll have a look at these things 👍

wellle:
All right, here's what I found:

1. I tried running it as `vim -u minivimrc`. Calling Startup() from the autocommand didn't really work for me, not sure why. I changed it to `map s :call Startup()<cr>` and called it manually. Then I had syntax highlighting and the gitgutter mappings. From the autocommand I had neither.

2. Conflicting mappings with gitgutter: The problem is how targets.vim gets invoked on mappings like `omap i`. If there are other mappings like `omap ic` and you type `vic` quickly (without waiting for a second between `i` and `c`) then Vim waits after the `i` to see if the `c` is coming. If it is, then Vim invokes the `ic` mapping, which is gitgutter in this case. Only if it runs into `'timeout'` or if there is no additional mapping for `ic` then the `i` mapping from targets.vim gets invoked which then checks what internal mappings have been set up. So I don't currently see a clean way of how targets.vim could change this behavior, as this is core Vim functionality. Messing without `'timeout'` often has ugly side effects. So unmapping seems like the best way. I like the idea of a health report to check for such cases. The hack to unmap does seem really hacky. It's prone to race conditions if the unmapping would happen before the mapping. And especially the way how gitgutter sets its mappings it seems like it might re-add them later after being unmapped in some cases? So yeah, not sure how to address this.

3. Shrink not working: The thing with `targets#modify#shrink` is that it drops both DELIMITERS and whitespace. So it's supposed to be called when a target with delimiters is selected. The best way to fix it seems to be doing it in `cmdtargets.vim` like you do it in `envtargets.vim`: Have `vimtex#text_obj#cmdtargets#inner` not drop the delimiters. Then use it together with `targets#modify#drop` for `i` and with `shrink` for `I`. (That's also how it's done in the builtin tag targets).

4. Wrong cmd target being selected from beginning of line 5: To me it seems like the implementation of the cmd target is somewhat broken. There's a suspicious use of a script local variable `s:current`. That sort of interaction between multiple generations should never be needed. If one generator needs to maintain some state, then the `a:state` argument should be used. The only place I used it so far is for quote targets, and there it's only to cache some somewhat expensive information.

    What's important to know: When seeking gets involved like here, targets.vim creates three generators: One for `current`, one for `next` and one for `last`. Then each one of those gets called exactly once. `current` should always return (on the first invocation) the smallest target which intersects the cursor (cursor being at start/end or within the target). `next`/`last` should return (on the first invocation) the closest target after/before the cursor (which does not intersect the cursor at all). Then targets.vim picks the best one of those depending on the range scores.

    What's happening with the current implementation is that no current target is found (as expected). No `next` target is returned (I assume this is because of the script local variable which was set in the `current` generator just before. The a `last` target is pickd on line 2 (shouldn't it be on line 4?). So the only contender is the `last` target on line 2.

    By the way, I saw this by uncommenting the block starting on line 143 in multigen.vim. You might try that for development/testing too.

5. I'm curious. What is the `++once`/`++nested` thing you're using in the autocommand function calls?

lervag:
>    1. I tried running it as `vim -u minivimrc`. Calling Startup() from the autocommand didn't really work for me, not sure why. I changed it to `map s :call Startup()<cr>` and called it manually. Then I had syntax highlighting and the gitgutter mappings. From the autocommand I had neither.

I don't understand why the autocommands did not work for you. Perhaps it is a version issue? I'm running pretty bleading edge: neovim 0.4.0 and Vim 8.1-1776.

>    5. I'm curious. What is the `++once`/`++nested` thing you're using in the autocommand function calls?

See `:help :autocmd`:

```
:au[tocmd] [group] {event} {pat} [++once] [++nested] {cmd}
                        ...
                                                        *autocmd-once*
                        If [++once] is supplied the command is executed once,
                        then removed ("one shot").
...
...
                                                        *autocmd-nested* *E218*
By default, autocommands do not nest.  If you use ":e" or ":w" in an
autocommand, Vim does not execute the BufRead and BufWrite autocommands for
those commands.  If you do want this, use the "++nested" flag for those
commands in which you want nesting. ...
```

>    2. Conflicting mappings with gitgutter: The problem is how targets.vim gets invoked on mappings like `omap i`. If there are other mappings like `omap ic` and you type `vic` quickly (without waiting for a second between `i` and `c`) then Vim waits after the `i` to see if the `c` is coming. If it is, then Vim invokes the `ic` mapping, which is gitgutter in this case. Only if it runs into `'timeout'` or if there is no additional mapping for `ic` then the `i` mapping from targets.vim gets invoked which then checks what internal mappings have been set up. So I don't currently see a clean way of how targets.vim could change this behavior, as this is core Vim functionality. Messing without `'timeout'` often has ugly side effects. So unmapping seems like the best way. I like the idea of a health report to check for such cases. The hack to unmap does seem really hacky. It's prone to race conditions if the unmapping would happen before the mapping. And especially the way how gitgutter sets its mappings it seems like it might re-add them later after being unmapped in some cases? So yeah, not sure how to address this.

Thanks for looking into this and for explaining how `targets.vim` works. Yes,
I also agree that the hack is hacky. But it worked in my tests, so I'm thinking
perhaps to keep it. I'll wait for a comment by @kiryph to decide.

>    3. Shrink not working: The thing with `targets#modify#shrink` is that it drops both DELIMITERS and whitespace. So it's supposed to be called when a target with delimiters is selected. The best way to fix it seems to be doing it in `cmdtargets.vim` like you do it in `envtargets.vim`: Have `vimtex#text_obj#cmdtargets#inner` not drop the delimiters. Then use it together with `targets#modify#drop` for `i` and with `shrink` for `I`. (That's also how it's done in the builtin tag targets).

Thanks, that works!

>    4. Wrong cmd target being selected from beginning of line 5: To me it seems like the implementation of the cmd target is somewhat broken. There's a suspicious use of a script local variable `s:current`. That sort of interaction between multiple generations should never be needed. If one generator needs to maintain some state, then the `a:state` argument should be used. The only place I used it so far is for quote targets, and there it's only to cache some somewhat expensive information.

First, thanks for the good explanation of how `targets.vim` works! Given the following LaTeX code, where `|` indicates the cursor position:

```tex
a \cmd{ bbbbbbbb }  \c|md{
ccccccc }  \cmd{ dddddd}  \cmd{eeeeeee } \cmd{ffffffff}
```

Without my `s:current` variable, `vic` matches the `b`s, whereas I would expect it to match the `c`s. When I debug as you suggest, I notice that the `b`s are given score 19, while the `c`s only get score 18. Not sure how to fix that, or even how to determine if it is really broken...

lervag:
Note, when the cursor is within the `c`s (score 17), `vic` will select the `d`s (score 20), which is even more surprising.

lervag:
Ok, so, it seems `lb` has lower score than `ll` and `ar` has lower score than `rr`. It does not feel intuitive to me, because this means that you choose the previous/next text object instead of the current.

wellle:
Thanks!

As for the surprising selections your seeing above is basically what https://github.com/wellle/targets.vim/issues/234 is about. I just commented on there with some background on why it behaves like this and how I'd like to improve it.

In short: By default targets.vim does prefer targets within the cursor line (`bbb` and `ddd` in your example) over multiline targets (`ccc` in your example). This can be changed with the seekRanges setting.

lervag:
Great, I'm happy to see an open issue on this and that you are thinking about possible improvements. And the seekRanges option seems to work well for now to change the default.

@kiryph Do you have any further comments? Are there remaining concerns to be addressed before we merge?

wellle:
I did some more quick tests and here's one thing I noticed:
```tex
a \cmd{ bbbbbbbb }  \cmd{ ccccccc }  \cmd{ dddddd}  \cmd{eeeeeee } \cmd{ffffffff}
%                                  ^^                                            
```

With the cursor between two commands `c` and `d`, `vilc` should select `c`, but it actually selects `b`.

kiryph:
### Automated Tests

I have adapted the automated tests from targets.vim in a first version for the vimtex-targets.

Right now the ok version is just copied over from the first run and has to be adjusted for the encountered issues.

### Conflicting mappings

Could vimtex define its mappings as following to overwrite vim-gitgutter as ftplugin:

```viml
omap <expr> <Plug>(targets-i) targets#e('o', 'i', 'i')
xmap <expr> <Plug>(targets-i) targets#e('x', 'i', 'i')
omap <expr> <Plug>(targets-a) targets#e('o', 'a', 'a')
xmap <expr> <Plug>(targets-a) targets#e('x', 'a', 'a')

omap <expr> ic <Plug>(targets-i)c
xmap <expr> ic <Plug>(targets-i)c
omap <expr> ac <Plug>(targets-a)c
xmap <expr> ac <Plug>(targets-a)c
```

### Multi-line issues

I did not follow all details in the discussion here about script-local variables and the scoring for the seek behavior of targets.vim.

I just noticed another multi-line issue:

`vic` selects in following situation when on the `cccccc` the `d` cmd not the current splitted `c` cmd:
```tex
a \cmd{ bbbbbbbb }  \cmd{
ccccccc }  \cmd{ dddddd}  \cmd{eeeeeee } \cmd{ffffffff}

%^^^^^^^
```
Can someone else confirm this with the most recent commit?

I think the next step could be to extend the automated tests for the multi-line issues with commands.

lervag:
> I have adapted the automated tests from targets.vim in a first version for the vimtex-targets.

Great, I like this idea very much. A side note: I might someday convert all my tests from the current Vader backend to a more plain backend with Makefiles.

> Right now the ok version is just copied over from the first run and has to be adjusted for the encountered issues.

Ok, we'll make sure the ok version is really OK before we merge.

> Could vimtex define its mappings as following to overwrite vim-gitgutter as ftplugin:

I think this looks like a good idea. It is still a little bit hacky, but it feels less hacky than the one I've added so far. What do you think, @wellle?

> I just noticed another multi-line issue:
> ...
> Can someone else confirm this with the most recent commit?

If I'm not mistaken, you've noticed the exact same issue that I noticed before. If so, it is not really a bug, but a "feature". It may be "resolved" if you change your `g:seekRanges` option as suggested by @wellle in an earlier comment.

> I think the next step could be to extend the automated tests for the multi-line issues with commands.

Yes. I'd say something like this:

- [ ] Add tests for multiline commands
- [ ] Add tests for environments
- [ ] Possibly: Update the travis settings to run these tests in addition to the vader tests?

lervag:
I've implemented the suggested vim-gitgutter mappings, and they seem to work well. I've also improved the Makefile a little bit. To me, it seems it now only remains to add the final tests.

wellle:
Yes, sounds good to me 👍

FYI: I just checked again, the issue I mentioned yesterday in https://github.com/lervag/vimtex/pull/1384#issuecomment-522754198 seems to still be present.

This seems to fix it AFAICT:
```diff
diff --git a/autoload/vimtex/text_obj/cmdtargets.vim b/autoload/vimtex/text_obj/cmdtargets.vim
index 6dec0ec..8623b0c 100644
--- a/autoload/vimtex/text_obj/cmdtargets.vim
+++ b/autoload/vimtex/text_obj/cmdtargets.vim
@@ -42,13 +42,8 @@ endfunction
 
 " }}}1
 function! vimtex#text_obj#cmdtargets#last(args, opts, state) " {{{1
-  " Normalize cursor position of current cmd: place on leading '\'
-  if targets#util#search('\\\S\+{', 'bWc') > 0
-    return targets#target#withError('no target')
-  endif
-
-  " Move to the last non-surrounding cmd
-  if targets#util#search('\\\S\+{\_.\{-}}', 'bW') > 0
+  " Move to the end of the last non-surrounding cmd
+  if targets#util#search('\\\S\+{\_.\{-}}', 'bWe') > 0
     return targets#target#withError('no target')
   endif
```

lervag:
Thanks again for your input, @wellle! I've tested your suggested fix, and it seems to fix the issue. I've also added the case to the tests.

wellle:
By the way, I just added a basic health check which checks for conflicting mappings as was suggested above: https://github.com/wellle/targets.vim/pull/244

Any news on this PR? Cheers!

lervag:
Sorry, I've had much to do lately and have only prioritied the "simple" issues in the meantime. I'll try to finish this PR soon.

I'll start by rebasing on top of current master. If @kiryph should have time to help with adding the tests, I would appreciate it, else I'll do it as soon as possible.

wellle:
No worries, thanks for the update 👍

lervag:
I did the rebasing now and made some minor updates. I've recently done what I mentioned; converting my tests into Makefile based tests.

One thing I noticed now was that the tests output a lot of noise seemingly from targets.vim:

```
:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo:call targets#undo(1):echo"test1.out" "test1.out" [New] 146L, 14999C written:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo:call targets#undo(2):echo"test1.out" "test1.out" 146L, 14999C written%
```

Perhaps there's a lingering debug echo in targets.vim? Or did I just miss it in
vimtex?

wellle:
That's coming from here: https://github.com/wellle/targets.vim/blob/be309773998ca729213206950109a758be15b556/autoload/targets.vim#L355

It's pretty old code (from 2014). As far as I remember this was the only way to trigger an undo in some cases in order to not mess up the undo history if we need to abort some execution (if no match was found). And I believe back then I only got it to work via `feedkeys()`. And this was visible in the command line. The `echo` part is to hide that from users.

But yes, it still can show up in places like this. Btw, how did you capture that output? In my tests I use `redir >> file` but I don't see any `targets#undo` in there, even though this case does get triggered in tests. 🤔 

I might revisit that some day to check if maybe today there are other ways which don't involve `feedkeys()`.

lervag:
Ok. I believe this PR is ready for merge. There might be some more tests that should be added, but I don't want to let perfect be the enemy of good. So if there are no objections, I'll merge by the end of the week. Obvoiusly, if there should be issues after merge then I'll be happy to work on solving them.

Note: Since August I've implemented a much better system for testing (IMHO), and so in the future I will be adding tests in a different manner then I've done the last _n_ years. This was partly due to inspiration from @kiryph's work, so thanks!

wellle:
Thanks! Sounds good to me 👍

lervag:
Great. This is merged now.

@kiryph Feel free to open a new issue/PR if you find there are still weak points that should be addressed!

Issue #1389, opened 2019-05-22
indentation of long command arguments


nilsbecker:
i am quite happy with vimtex's formatting functionality. a nitpick i still have is this:

typing this figure float:
```
\begin{figure}[htpb]
	\includegraphics[width=0.8\linewidth]{name.ext}%
	\centering%
	\caption{alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s.alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s. alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s.alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s. alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s.}
\end{figure}
```
and then `gqip` gives this:
```
\begin{figure}[htpb]
	\includegraphics[width=0.8\linewidth]{name.ext}%
	\centering%
	\caption{alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s.alksdjf alksdj
		alkjsjd alksj dklfjalkjs dklfj alkj s. alksdjf alksdj alkjsjd alksj
		dklfjalkjs dklfj alkj s.alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj
	s. alksdjf alksdj alkjsjd alksj dklfjalkjs dklfj alkj s.}
\end{figure}
```
the same result can be obtained by changing indentation and then doing `=ip`. the indentation of the last line of the `\caption` to align with the `\caption` command seems odd to me. i could agree if the last line had only the closing brace, but with text it is distracting. 

i would propose to align the last line of the `caption` argument with the previous lines of the `caption` argument.

kiryph:
If this is bothering you right now and lervag does not change it soon, you could invest time into using https://github.com/cmhughes/latexindent.pl as `formatprg`/`equalprg`.

lervag:
> lervag does not change it soon

Sorry guys, life's been very busy lately, and so I'm not able to stay quite on top of the issues. I'll try my best to fix and address things (including this issue) in the following couple of weeks.

lervag:
Thanks for raising this issue. It's been a few years since I last considered this. The last time, I settled on the current behaviour, because I was able to make it work reliably. The issue with the feature you are asking for is that it is harder to keep track of the delimiter counts.

[...]

PR #1405, opened 2019-06-06
Add new delimiter indent behaviour (cf. #1389)

lervag:
This PR aims to implement a solution for #1389.

A very good contribution to this PR would be to get a good and complicated example for testing that the indent algorithm works as intended.

TODO:
- [x] Ensure that `\[ ... \]` is handled properly
- [ ] Get feedback on the choice of option name
- [ ] Get feedback on whether the feature seems correctly implemented

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1628,14 +1628,13 @@ OPTIONS                                                        *vimtex-options*
     close~
       List of regexes for closing delimiters that should reduce indents.
 
+    close_indented~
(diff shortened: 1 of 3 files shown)


lervag:
@nilsbecker @kiryph 

Ok, I've pushed this to master now after some minor modifications. The old indentation behaviour is still default, so to test it you need to set your `g:vimtex_indent_delims` option as specified in the docs.

nilsbecker:
great, i will i try it out.

PR #1415, opened 2019-07-01
Allow arbitrary hooks into latexmk output.

rsekman:
Sometimes when a document takes a long time to compile I'd like to know how far along the build is. This could be easily accomplished by matching latexmk output looking for "Run number [0-9]+ of rule" and doing something with that, but then I need to read latexmk output in a callback. Currently vimtex only allows callbacks when compilation finishes. 

To add more callback support we can mirror how `g:vimtex_compiler_callback_hooks` are handled already.

The option `g:vimtex_compiler_callback_hooks_cont` (`_cont` for continuous) is a list of funcrefs that will be called whenever a line is read as output from latexmk running in continuous mode. That line is passed as an argument to the hook. Funcrefs are used because `execute` with string concatenation will not work; the message can contain arbitrary strings (c.f. SQL injection).

Example usage:

> ``` viml
> function! LatexmkCallback(msg)
> 	let l:m =  matchlist(a:msg, '\vRun number ([0-9]+) of rule ''(.*)''')
> 	if !empty(l:m)
> 		echomsg l:m[2] . ' (' . l:m[1] . ')'
> 	endif
> endfunction
> let g:vimtex_compiler_callback_hooks_cont = [function('LatexmkCallback')]

Whenever a new command is started by latexmk, it is `echomsg`:d with the run number in parentheses.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -29,7 +29,6 @@ function! vimtex#init_options() abort " {{{1
   call s:init_option('vimtex_compiler_progname',
         \ get(v:, 'progpath', get(v:, 'progname')))
   call s:init_option('vimtex_compiler_callback_hooks', [])
-  call s:init_option('vimtex_compiler_callback_hooks_cont', [])
   call s:init_option('vimtex_compiler_latexmk_engines', {})
   call s:init_option('vimtex_compiler_latexrun_engines', {})
 
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -29,6 +29,7 @@ function! vimtex#init_options() abort " {{{1
   call s:init_option('vimtex_compiler_progname',
         \ get(v:, 'progpath', get(v:, 'progname')))
   call s:init_option('vimtex_compiler_callback_hooks', [])
+  call s:init_option('vimtex_compiler_callback_hooks_cont', [])
   call s:init_option('vimtex_compiler_latexmk_engines', {})
   call s:init_option('vimtex_compiler_latexrun_engines', {})
 
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -524,6 +524,11 @@ function! s:callback_continuous_output(channel, msg) abort " {{{1
   elseif a:msg ==# 'vimtex_compiler_callback_failure'
     call vimtex#compiler#callback(0)
   endif
+
+  for l:Hook in g:vimtex_compiler_callback_hooks_cont
+      call l:Hook(a:msg)
+  endfor
+
 endfunction
 
 " }}}1
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -41,12 +41,12 @@ let s:compiler = {
       \   '-synctex=1',
       \   '-interaction=nonstopmode',
       \ ],
+      \ 'hooks' : [],
       \ 'shell' : fnamemodify(&shell, ':t'),
       \}
 
 function! s:compiler.init(options) abort dict " {{{1
   call extend(self, a:options)
-
   call self.init_check_requirements()
   call self.init_build_dir_option()
   call self.init_pdf_mode_option()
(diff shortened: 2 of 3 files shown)


lervag:
I can see your reasoning for this, but the implementation is currently lacking. I'll comment on the code directly.

autoload/vimtex.vim
```diff
@@ -29,6 +29,7 @@ function! vimtex#init_options() abort " {{{1
   call s:init_option('vimtex_compiler_progname',
         \ get(v:, 'progpath', get(v:, 'progname')))
   call s:init_option('vimtex_compiler_callback_hooks', [])
+  call s:init_option('vimtex_compiler_callback_hooks_cont', [])
```
lervag: The option is specific to latexmk, so I would rather use `g:vimtex_compiler_latexmk.hooks`. Let me know if you need further clarification on this.

autoload/vimtex/compiler/latexmk.vim
```diff
@@ -524,6 +524,11 @@ function! s:callback_continuous_output(channel, msg) abort " {{{1
   elseif a:msg ==# 'vimtex_compiler_callback_failure'
     call vimtex#compiler#callback(0)
   endif
+
+  for l:Hook in g:vimtex_compiler_callback_hooks_cont
```
lervag: Please use `l:hook` instead of `l:Hook` to be style consistent.

autoload/vimtex/compiler/latexmk.vim
```diff
@@ -524,6 +524,11 @@ function! s:callback_continuous_output(channel, msg) abort " {{{1
...
+      call l:Hook(a:msg)
+  endfor
+
```
lervag: Remove unnecessary empty line.

autoload/vimtex/compiler/latexmk.vim
```diff
@@ -524,6 +524,11 @@ function! s:callback_continuous_output(channel, msg) abort " {{{1
   elseif a:msg ==# 'vimtex_compiler_callback_failure'
     call vimtex#compiler#callback(0)
   endif
+
+  for l:Hook in g:vimtex_compiler_callback_hooks_cont
+      call l:Hook(a:msg)
```
lervag: Use two spaces for indentation.

autoload/vimtex.vim
rsekman: This is a much better name than mine,

autoload/vimtex/compiler/latexmk.vim
rsekman: The capital letter is to avoid E704 being raised.

lervag: You don't need that when you prepend with `l:`.

rsekman: I tried it with lowercase `h` first and changed it to a capital `H' after encountering E704. Maybe it's a vim/neovim difference? On my system this script raises E704 and then E884.

> ``` viml
> function! Test()
>   echomsg "Hello"
> endfunction
> function! Caller()
>   let l:f = function('Test')
>   call l:f()
> endfunction
> call Caller()
> ```

I'm on

> \>  vim --version
> VIM - Vi IMproved 8.1 (2018 May 18, compiled May  6 2019 22:03:03)
> MS-Windows 64-bit console version
>Included patches: 1-1286

lervag: Ok, perhaps you're right, sorry.

autoload/vimtex/compiler/latexmk.vim
```diff
@@ -525,10 +525,9 @@ function! s:callback_continuous_output(channel, msg) abort " {{{1
     call vimtex#compiler#callback(0)
   endif
 
-  for l:Hook in g:vimtex_compiler_callback_hooks_cont
-      call l:Hook(a:msg)
+  for l:Hook in b:vimtex.compiler.hooks
```
rsekman: This was the only way to access the compiler object from this scope that I could figure out. I don't know if it's "correct".

lervag:
Great. Good work.

lervag:
I made some minor updates. Feel free to test that this still works as expected.

kiryph:
BTW, I would have added the example to the doc:


> For example, whenever a new command is started by latexmk, following hook prints a message with the command and the run number:
> 
> ```viml
>  function! LatexmkCallback(msg)
> 	let l:m =  matchlist(a:msg, '\vRun number ([0-9]+) of rule ''(.*)''')
> 	if !empty(l:m)
> 		echomsg l:m[2] . ' (' . l:m[1] . ')'
> 	endif
> endfunction
> let g:vimtex_compiler_latexmk = { 'hooks': [function('LatexmkCallback')] }
> ```

lervag:
Yes, good point. I agree.

PR #1419, opened 2019-07-05
Add imap `H for \hbar

rsekman:
I use `\hbar` a lot and it would be nice to have an imap for it. `` `h `` is taken for `\eta`, but `` `H `` is available.

I thought it logical to put it near `\langle` and `\rangle` because of bra-ket notation.


lervag:
Seems OK. Thanks for suggesting a PR!

Issue #1425, opened 2019-07-20
completion (closing) of quotes


927589452:
**Is your feature request related to a problem? Please describe it.**
It would be nice to use ]]-completion for \grqq 

**Describe the solution you'd like**
\glq with \grq
\glqq with \grqq
\flq with \frq
\flqq with \flqq
included in the  ]]-completion dictionary

**Describe alternatives you've considered**
Usage of cs quotes

**Additional context**
taken from [here](https://de.wikibooks.org/wiki/LaTeX-W%C3%B6rterbuch:_Anf%C3%BChrungszeichen#Deutsche_Anf%C3%BChrungszeichen)

kiryph:
Did you already try one of the "autoclosing" plugins:

- **Autopairs** https://github.com/jiangmiao/auto-pairs#multibyte-pairs
- **Lexima** https://github.com/cohama/lexima.vim#customization
  (https://github.com/deathlyfrantic/lexima-template-rules)

Also [**UltiSnips**](https://github.com/SirVer/ultisnips) can help (if you like with auto-expansion otherwise remove the `A` at the end of the snippets below):

```
snippet glq "\glq \grq" wA
\glq$1\grq$0
endsnippet

snippet glQ "\glqq \grqq" wA
\glqq$1\grqq$0
endsnippet

snippet flq "\flq \frq" wA
\flq$1\frq$0
endsnippet

snippet flQ "\flqq \frqq" wA
\flqq$1\frqq$0
endsnippet
```

927589452:
I would prefer for it to be the same way as for other delims,
so I would add it to 
function! s:init_delim_lists() 
as an subdictionary if @lervag doesn't object

I try to get away from this quotation as BibLaTeX prefers usage of cs quotes too,
but I would implement it at this position, if there are no problems with this

thanks @kiryph for your suggestions, but the first one directly "conflicts" with ( ) closing of vimtex for example

kiryph:
> thanks @kiryph for your suggestions, but the first one directly "conflicts" with ( ) closing of vimtex for example

This should be [configurable](https://github.com/jiangmiao/auto-pairs/blob/3c290d9308c1b651c588865c3741a12654ae640e/doc/AutoPairs.txt#L206-L217): `au Filetype tex let b:AutoPairs = {"\glq": "\grq"}`

> so I would add it to `function! s:init_delim_lists()`

Did you already try 

```viml
g:vimtex_delim_list = {
        \ 'delim_tex' : {
[...]

PR #1431, opened 2019-07-23
Update vimtex.txt

927589452:
Added an example from #1425 to the doc


lervag:
Thanks for the suggestion! I took the liberty of adding the example in the docs for the option and instead adding a reference to the option in the section you changed.

Issue #1426, opened 2019-07-21
[docs-update request] Provide a more direct link to `latexrefman` in the documentation


krishnakumarg1984:
In the documentation page for `vimtex`, I notice that the link to `latexreman` is given as 
`https://puszcza.gnu.org.ua/projects/latexrefman/` 

This is in-fact the GNU project-tracker page, which includes a lot more "meta-info" than the actual Latex-help contents of `latexrefman`.

It would be more helpful to readers if the link was updated to directly point to the actual documentation, i.e. [`https://latexref.xyz/`](https://latexref.xyz/) which is what most users are looking for.

lervag:
Thanks! This is resolved by #1432 and my recent commit.

krishnakumarg1984:
You are awesome! Thanks @lervag

lervag:
Thanks! :)

PR #1432, opened 2019-07-23
Update vimtex.tex | added reference to latexref.xyz

927589452:
added a reference to https://latexref.xyz/ for the online and the offline section
as wished for in #1426


lervag:
Thanks!

krishnakumarg1984:
Thanks to you both!

PR #1453, opened 2019-08-21
move examples into comments

927589452:


lervag:
Thanks!

By the way: For the issue template, I'm thinking perhaps that it will look better if the `minivimrc` and similar are typed in bold instead of italized. What do you think?

927589452:
On 19-08-22 00:14:27, Karl Yngve Lervåg wrote:
> By the way: For the issue template, I'm thinking perhaps that it will look better if the `minivimrc` and similar are typed in bold instead of italized. What do you think?
I think so too, will provide a PR, but this means dropping the **steps to reproduce**, as we loose the structure, if steps to reproduce is the same "level"

PR #1464, opened 2019-08-27
Build dir doubling

927589452:


lervag:
Ok, thanks for showing me what I need to reproduce the issue. I'll not accept the PR, because it is a minimal change and I want to do it a little bit differently. But it is still useful in showing me exactly what was wrong with the example.

PR #1468, opened 2019-09-01
Add newcommand-like completion for DeclareDocumentCommand from xparse

philipdexter:


lervag:
Thanks. This seems useful and properly implemented. I like that you also added a minimal example!

Issue #1463, opened 2019-08-27
Add a `:VimtexBug` and `:VimtexFeatureRequest` command


927589452:
**Is your feature request related to a problem? Please describe it.**
It would be very nice, to type a bug report or feature request in `vim` and only copy paste it here

**Describe the solution you'd like**
A command using the template (and maybe some magic comments to run the commands like `:VimtexInfo` etc automatically), probably warning if the `minimal.vim` is more than X lines and a comment at the buttom for new users to use `"*yg` and paste it into a github buffer.

**Describe alternatives you've considered**
Just opening the file in `.github/*` manually

**Additional context**
The template could hint at this option it self, to generate better reports, eg guide people to supply the needed files.

lervag:
I partly agree it would be nice, but I fear the feature is quite complex to implement. I would rather implement a better `VimtexInfo` or similar that could include some basic checks that could indicate solutions to basic features.

Another idea is to create an issue template (which sort of already exists), and to ask people to reproduce their bug by altering the template. E.g. instructions like:

1. Copy template with `cp -r VIMTEX/test/issues/template my-test-dir` and `cd my-test-dir`
2. Edit `minivimrc` and `minimal.tex` so that `vim --servername VIM -u minivimrc` properly reproduces the issue.

927589452:
This sounds like a good idea too.

A lighter (for now) version could be to just open a buffer (that is not closed by `ESC`, as there are many people still using arrows to navigate) and load the template from the `.github` directory (all plugin managers do a full `git clone`, right? ) and possibly strip the internal github comments from it.

I will do some research on the syntax of the github template, eg if there is a comment marker, which is not displayed on github, but is displayed if opened in vim

my main reason to suggest this feature is: My randomly crashing firefox taking all my already typed notes down with it.

lervag:
> my main reason to suggest this feature is: My randomly crashing firefox taking all my already typed notes down with it.

I use Vim to edit text almost everywhere, including github. I use a simple script like this:

```sh
#!/usr/bin/env bash

file="/tmp/edit-with-vim_$(date +%F_%H:%M:%S)"
xclip -o > "$file"

urxvt -e nvim "$file" -c "set filetype=wiki"
truncate -s -1 "$file"

xclip -selection c "$file"

xdotool key ctrl+v
```

Save it, make it executable and available to your desktop environment, then map a key for it (I personally use `<alt>e`). This will:

1. Copy the contents of your clipboard to a file.
[...]

PR #1472, opened 2019-09-07
add commands to embed files from `:VimtexIssue`

927589452:
this is a work in progress

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_report.md
+++ b/.github/ISSUE_TEMPLATE/issue_report.md
@@ -56,6 +56,8 @@ labels: bug
 <!-- Describe both the expected and the observed behaviour-->
 
 <!-- *Note*: if relevant, include the content of your `.latexmkrc` file -->
+<!-- If you opened this from vimtex, you can just type `<ESC>:read ~\.latexmkrc`
+     or `<ESC>:read .latexmkrc` for a local `.latexmkrc` -->
 
 **Output from VimtexInfo**
 <!-- Run `:VimtexInfo` and paste the content here -->


927589452:
this is for #1463

927589452:
I am reading the `autoload/vimtex/info.vim`, but I don't know how to get the return values of the internal commands.
maybe a public function wrapping the inner lines of the print statement, which can be used inside the template to get the lines and inside the print statement to append them

lervag:
> I am reading the autoload/vimtex/info.vim, but I don't know how to get the return values of the internal commands.

Not quite sure what you mean.

> maybe a public function wrapping the inner lines of the print statement, which can be used inside the template to get the lines and inside the print statement to append them

An example could perhaps help make it clearer what you mean?

927589452:
> > I am reading the autoload/vimtex/info.vim, but I don't know how to get the return values of the internal commands.
> 
> Not quite sure what you mean.
> 
> > maybe a public function wrapping the inner lines of the print statement, which can be used inside the template to get the lines and inside the print statement to append them
> 
> An example could perhaps help make it clearer what you mean?

If I understood correctly, the output of `VimtexInfo` is assembled directly in the buffer, but I wanted to do a `<ESC>:read VimtexInfoOutput`, so it could be easily embedded in the document.
But while thinkig about it, I just realized a `VimtexInfoInsert` inserting the output into the current buffer could work too

lervag:
I really have to admit I don't quite understand how this should work. I think I might have mentioned it before, but I propose that before you try to implement anything, you should write the documentation. This would make it easier for me to understand what you want to achieve, and it would be easier to help.

lervag:
I'm closing due to lack of activity, but please feel free to reopen if you want to continue the discussion/idea!

Issue #1479, opened 2019-09-10
Include label type in completion


clason:
At the moment, vimtex shows, e.g., `(3.1)` in the `menu` field of the completion item for all labels that receive the number 3.1 -- equations, Theorems, Figures, etc. It would be nice to distinguish the type of reference here, i.e., printing `Theorem 3.1` or `Figure 3.1` in the menu field. This would also allow easily filtering on reference type (if no consistent label form is used).

I believe this is doable since this information is contained in the `.aux` file, e.g.,
```
\newlabel{lem:state_existence}{{2.1}{3}{}{theorem.2.1}{}}
\newlabel{fig:variation:lsc1}{{\relax 2.1a}{15}{<caption text>}{figure.caption.68}{}}
```
(the fact that the first shows the counter -- which here coincides with that for theorems even though it is a lemma -- would be acceptable to me; relying on `cref` to get the "correct" label name is probably not worth the added effort). I think it would be sufficient to extend `tex2tree` to include the additional field and use a more detailed case distinction in `parse_number` (also returning appropriately formatted numbers instead of deferring that to `completer_ref.complete`)?

lervag:
Interesting proposal. I agree it seems doable. Perhaps, as a minor improvement, one could have an option to map e.g. "theorem" to "Theorem" and "figure caption" to "Figure"?

clason:
Oh, yes, proper capitalization was assumed :) (As was splitting at the first `.`, so `caption` would be removed anyway.)

lervag:
Hmm. I've pushed a simple example for testing this. If you go to the test dir and do `make`, it should output the completion candidates. However, if you open the resulting `minimal.aux`, you will see it does _not_ contain the added context.

I thus believe this is probably less doable than first thought. That is, I think the content of the `\newlabel` depends on packages. Probably, `cleveref`?

clason:
Don't you need to actually set a `label` for that (instead of two `cref`)? And `cleveref` should not be needed for that unless you want to distinguish different theorem environments sharing the same counter (which I would not do at first).

lervag:
> Don't you need to actually set a label for that (instead of two cref)? And some environment (or at least section) to attach the label to?

Not sure what you mean. But please, if you can help me to fix the minimal example so that it properly represents your proposal, then that would be helpful!

> And cleveref should not be needed for that unless you want to distinguish different theorem environments sharing the same counter (which I would not do at first).

Ok, and agreed.

clason:
[...]

PR #1488, opened 2019-09-15
expand the regex to find commands

927589452:


927589452:
this is missing the 
```
\NewDocumentEnvironment
\RenewDocumentEnvironment
\ProvideDocumentEnvironment
\DeclareDocumentEnvironment
```
yet

lervag:
Could you please add a description of this as well? And preferably also a test document to show the effect.

927589452:
Whats the best way to test the autocompletion?
Will open a feature request so the testfiles have a home

lervag:
> Will open a feature request so the testfiles have a home

Actually, you could just as well add files under `issues/1488`. That is, no reason to open a new issue. Instead, you could just have added the text you put in #1490 to the original post here.

No harm done, though. Just my two cents on making this more concise.

> Whats the best way to test the autocompletion?

Good question. I typically do something like [the test files for #1479](https://github.com/lervag/vimtex/tree/master/test/issues/1479). It includes a Makefile, so testing is simply a `make` away.

(In the future, I plan to remove the use of Vader for auto testing, and instead rely on something similar to this type of Makefile based testing.)

lervag:
I've updated the test files, please pull changes to your local version. I've added a Makefile and updated the minimal.vim file so that you can easily test. Note, right now, the test fails, because the command completion lacks a couple of the candidates like `VimtexStandardProvide`.

autoload/vimtex/complete.vim
```diff
@@ -4,7 +4,7 @@
 " Email:      karl.yngve@gmail.com
 "
 
-function! vimtex#complete#init_buffer() abort " {{{1
+function! vimtex#complete#init_buffer() abort " {{{2
```
lervag: I assume this was a mistake?

autoload/vimtex/complete.vim
```diff
@@ -854,9 +854,9 @@ function! s:completer_env.gather_candidates_from_newenvironments() dict abort "
 
   let l:candidates = vimtex#parser#tex(b:vimtex.tex, {'detailed' : 0})
 
-  call filter(l:candidates, 'v:val =~# ''\v\\((re)?newenvironment|(New|Renew|Provide|Declare)DocumentEnvironment)''')
+  call filter(l:candidates, 'v:val =~# ''\v\\((renew|new)environment|(New|Renew|Provide|Declare)DocumentEnvironment)''')
   call map(l:candidates, '{
-        \ ''word'' : matchstr(v:val, ''\v\\((re)?newenvironment|(New|Renew|Provide|Declare)DocumentEnvironment)\*?\{\\?\zs[^}]*''),
+        \ ''word'' : matchstr(v:val, ''\v\\((renew|new)environment|(New|Renew|Provide|Declare)DocumentEnvironment)\*?\{\\?\zs[^}]*''),
```
lervag: This change seems unnecessary.

lervag:
See my two comments to the code. Also, just because I'm curious: From which package or class do this family of commands come from, e.g. `\NewDocumentEnvironment`, `\RenewDocumentEnvironment`, `\ProvideDocumentEnvironment`, and `\DeclareDocumentEnvironment`?

927589452:
> See my two comments to the code. Also, just because I'm curious: From which package or class do this family of commands come from, e.g. `\NewDocumentEnvironment`, `\RenewDocumentEnvironment`, `\ProvideDocumentEnvironment`, and `\DeclareDocumentEnvironment`?

that is all xparse

autoload/vimtex/complete.vim
927589452: it technically is, but this keeps the notion similiar for environment and command
this gives IMO a better readability and is needed to differentiate between `provide` and `renew`

I have no idea on performance, but I thought about going back to the beginning and splitting the `xparse` parsing into a dedicated function

and add some more for other packages

autoload/vimtex/complete.vim
927589452: probably, I will look into it, but monday is print time

lervag:
Ok, thanks. Is this PR ready to merge, in your opinion?

927589452:
> Ok, thanks. Is this PR ready to merge, in your opinion?

depends on wether later regexes should
- extend these line
- add a function per package providing commands

lervag:
I would keep it as it is for now, then consider to update later if necessary. For now, there is only a single package that adds to the standard commands (`xparse`).

lervag:
Ok, I've merged this now. Thanks!

PR #1492, opened 2019-09-17
Detect *.cls files as tex filetypes.

ruifm:
By default. neither vim nor neovim mark `*.cls` files as of the `tex` filetype.

This is a simple addition I had in my dotfiles that I thought would be a good fit to vimtex. I might be wrong. In case many people interpret  `*.cls` files as something else than a LaTeX class file, then disregard this pull request.

### DIFF
--- a/ftdetect/tex.vim
+++ b/ftdetect/tex.vim
@@ -1,0 +1,1 @@
+autocmd BufNewFile,BufFilePre,BufRead *.cls setfiletype tex


lervag:
No, this is not true. See `$VIMRUNTIME/filetype.vim` line 1526 (or close to it):

```vim
" Smalltalk (and TeX)
au BufNewFile,BufRead *.st			setf st
au BufNewFile,BufRead *.cls
	\ if getline(1) =~ '^%' |
	\  setf tex |
	\ elseif getline(1)[0] == '#' && getline(1) =~ 'rexx' |
	\  setf rexx |
	\ else |
	\  setf st |
	\ endif
```

The reason it is not cleanly `cls` to `tex` is because there are other file formats that use the `.cls` extension. This is handled with a minor heuristic as shown above. If I accepted this pull request, it would disrupt other peoples workflow that use `cls` files for e.g. Smalltalk.

As such, I won't accept the pull request. If you want to force filetype `tex` for `cls` files beyond this default setting, then I think it should be done on a personal customisation level.

ruifm:
> The reason it is not cleanly `cls` to `tex` is because there are other file formats that use the `.cls` extension. This is handled with a minor heuristic as shown above.

I guess that heuristic does not work for me since it assumes that there is a comment on the first line and that is not always the case.

> If you want to force filetype `tex` for `cls` files beyond this default setting, then I think it should be done on a personal customisation level.

You are absolutely right. I thought of adding an if guard to the `autocmd` that would check a global variable such as `g:vimtex_assume_cls_as_tex` that would default to `0` but that would be kind of pointless anyway since it would involve the same number of user configuration lines.

> As such, I won't accept the pull request.

No problem! It would only make sense if it would be a universal good fit. Thanks for taking the time to look into it anyway!

lervag:
> Thanks for taking the time to look into it anyway!

No problem. I'm just happy that people share their ideas and suggestions. Even if I don't agree with everything, I think it is very helpful in building a good community!

PR #1519, opened 2019-10-16
document root file name as unique with latexmain

tiagoboldt:
I've spent some time fighting with `.latexmain` and vimtex, with the plugin not picking up the right file. I ended up understanding that if the root file has the same name as files inside other folders, it conflicts. Changing the root name file solved it.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -547,6 +547,12 @@ File .latexmain specifier~
 <
   Here `path/file.tex.latexmain` indicates for `file1.tex` and `file2.tex`
   that `path/file.tex` is the main LaTeX file.
+  
+  Note that your root file name must be unique. You cannot have a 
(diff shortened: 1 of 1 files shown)


lervag:
Huh; interesting. I would have guessed that this should have worked. I'll look into it, perhaps it is a bug.

lervag:
Given the following files:

```
.
├── section1
│   └── main.tex
├── included.tex
├── main.tex
└── main.tex.latexmain
```

Now, my understanding is that you expect the main file `./main.tex` to be detected for all files here, including `section1/main.tex`, right? This is the observed behaviour for me.

If this does not work for you, can you please provide an example where it fails?

tiagoboldt:
This is actually working now. I must have been on the wrong root path. Sorry for the confusion, thanks for the fast response.

lervag:
No problem, and thanks for going straight for a PR when you had the issue. I like that attitude/initiative!

Issue #1533, opened 2019-11-08
Base file not being correctly set for \input files


FreddieOliveira:
**Issue**
When the `\documentclass{}` is `\input`ed into the main.tex instead of writing it directly in it, the base file for all subsequents  `\input`ed files is not set to the main.tex automatically, thus preventing forward searches to work (failing with `vimtex: Viewer cannot read PDF file!` error). One needs to manually put `%! TEX root = ../main.tex` in the sub.tex first line or create the `main.tex.latexmain` file.

To reproduce the problem consider the following directory structure:
 .
├── main.tex
├── preambule.tex
└── **dir**
                        └── sub.tex

and the following files contents:

**preambule.tex**

  ```tex
  \documentclass{article}
  ```

**main.tex**

  ```tex
  \input{preambule.tex}
  \begin{document}
  \input{dir/sub.tex}
  \end{document}
  ```

**sub.tex**

  ```tex
  \chapter{Hello World}
  Some text here...
  ```
**Observed Behaviour**
When running `:VimtexInfo` inside sub.tex window the oberserved base file is **sub.tex**

**Expected Behaviour**
When running `:VimtexInfo` inside sub.tex window the base file should be **main.tex** in the same way it works when not `input`ing the preambule and writing the `\documentclass{article}` directly in the main.tex

lervag:
Yes, this is a known limitation. See `:help vimtex-multi-file` for the description of how vimtex handles multi file projects. There are several ways of detecing the main file. The "default" method is an automatic, recursive search. From the docs:

>   If no other method provides an appropriate candidate, then the recursive
  search detects the main LaTeX file by searching for a file in the current
  and parent directories that includes the present file and has the
  `\documentclass` line.

When I implemented this, I found it difficult to handle the situation that you are describing here. I could not find a robust method of properly detecting this case unless I defined `preambule.tex` as the main file. But that would be wrong, because the entire point of this proposed LaTeX structure is to let different "main" files use the same preamble header.

I still think this sounds too difficult to properly handle, and since there are alternative ways to handle it that does work well, I propose that you use them:

1. Use the `TeX root` directive, or
2. Use the `.latexmain` specifier.

FreddieOliveira:
Thank you for the plugin and your response. It's totally understandable the not implementation for this limitation. Currently I'm using `.latexmain` specifier and it's working great.

lervag:
[...]

PR #1538, opened 2019-11-18
Fix file path comparation to correctly match subfiled's  parent

FreddieOliveira:
Ensure the comparation between files path are correct. Further explained [here](https://github.com/lervag/vimtex/issues/1533#issuecomment-554896387). This fixes the recursive search for who subfiled a file

### DIFF
--- a/autoload/vimtex/state.vim
+++ b/autoload/vimtex/state.vim
@@ -419,15 +419,21 @@ function! s:file_reaches_current(file) abort " {{{1
     let l:file = matchstr(l:line, g:vimtex#re#tex_input . '\zs\f+')
     if empty(l:file) | continue | endif
 
-    if !vimtex#paths#is_abs(l:file)
-      let l:file = fnamemodify(a:file, ':h') . '/' . l:file
-    endif
-
     if l:file !~# '\.tex$'
       let l:file .= '.tex'
     endif
 
-    if expand('%:p') ==# l:file
+    " l:current_file is the current file's absolute path by default
+    let l:current_file = expand('%:p')
+
+    " adjust it to relative path if needed
+    if !vimtex#paths#is_abs(l:file)
+    	let l:current_file = expand('%')
+    endif
+
+    " now its guaranteed we're comparing absolute path
+    " with absolute, or relative with relative
+    if l:current_file ==# l:file
           \ || s:file_reaches_current(l:file)
       return 1
     endif


lervag:
I'm happy for the PR and the investigation on your side. I think, though, that instead of patching `s:file_reaches_current`, I instead want to patch `s:get_main_from_subfile` and ensure that the file argument is passed in is absolute. I'll also want to add a comment about this to avoid similar errors in the future. I'm 99% sure that this would work, and that it is a slightly better way to solve the issue. Finally, I would love to have a test for this specific issue.

Let me know if you want to give this a shot yourself. Else I'll handle it. In particular, I would be very happy if you could provide a simple example where the current version fails. Either give it in the PR thread, or, if you want, add it to the PR within the `test/tests/test-get-main/test-subfiles` folder.

FreddieOliveira:
The problem with patching `s:get_main_from_subfile` is that it has nothing to do with this little bug. The argument it passes to `s:file_reaches_current` when calling it in line 290 is the file name as found in `\documentclass[<filename>]{subfiles}` inside the subfiled file.

Let's consider the following minimal example:

Directory structure:

**/home/user/Documents**
├── main.tex
└── **dir**
         └── sub.tex

Files contents:

**main.tex**

  ```tex
\usepackage{subfiles}
\documentclass{article}
\begin{document}
\subfile{dir/sub.tex}
\end{document}
  ```

**sub.tex**

  ```tex
\documentclass[main.tex]{subfiles}
\begin{document}
\chapter{Hello World}
Some text here...
\end{document}
  ```

When opening sub.tex from the root folder (/home/user/Documents/) `s:get_main_from_subfile` will set the variable `l:candidate` to `main.tex` in line 288 and then call `s:file_reaches_current` passing `l:candidate` to it. `s:file_reaches_current` will then parse the main.tex file contents, find the `\subfile{dir/sub.tex}` line on it and extract its argument to `l:file`, which is now holding the value `dir/sub.tex`. 

Now it's when the problem starts (not in the `s:get_main_from_subfile`): It will then compare the `l:file` with the absolute path of the current opened file, in our case it's sub.tex. But its absolute path will expand to /home/user/Documents/dir/sub.tex, so the comparation `if expand('%:p') ==# l:file` will translate to `if '/home/user/Documents/dir/sub.tex' ==# 'dir/sub.tex'`, which is what's causing `s:file_reaches_current` to fail

As far as I can see, the only way to patch this bug is by modifying the if comparation in `s:file_reaches_current`.

Let me know if I misinterpreted things

lervag:
> Let me know if I misinterpreted things

You did. The simple proof is given in the latest update that fixes things as I suggested. I think the main thing you missed is the point of this:

```vim
if !vimtex#paths#is_abs(l:file)
  let l:file = fnamemodify(a:file, ':h') . '/' . l:file
endif
```

This code is meant to ensure that the parsed filenames become absolute paths based on the location of the file that has the `\subfile` macro.

> the if comparation

FYI: "comparison" is probably the word you want here. :)

FreddieOliveira:
Cool. Now it's working! 

>You did. The simple proof is given in the latest update that fixes things as I suggested. I think the main thing you missed is the point of this:

Ooh, of course. I completely misunderstood this piece of code. Now I understand your solution.

> FYI: "comparison" is probably the word you want here. :)

Oops, hehe

>The only upgrade we can make in s:get_main_from_subfile is to append .tex to l:filename in case it doesn't have already, right after line 275, just like you did here. This would allow the user to write \documentclass[main.tex]{subfiles} as well as \documentclass[main]{subfiles} in the sub.tex.

Please, take a look at this whenever you have time, I wont even attempt to make a pull request 😅 I did some local changes though by adding

```vim
if l:filename !~# '\.tex$'
    let l:filename .= '.tex'
endif
```

right after [line 275](https://github.com/lervag/vimtex/blob/e555f48c131088fd66a9f7822c24e6edba8bfd03/autoload/vimtex/state.vim#L275) and it seems to be working fine.

lervag:
Ah, sorry, I didn't notice that comment. So, what you are saying is that `\documentclass[main]{subfiles}` is valid TeX subfiles syntax. I was not aware of it, and I agree that you suggested solution looks good. I just want to test it first, then I'll push it.

FreddieOliveira:
Thank you!

Issue #1536, opened 2019-11-15
vimtex#fold: commented code with three closing }}} ends environment fold unexpectedly


kiryph:
**Issue**
Consider the folds of this minimal.tex
  ```tex
\documentclass{article}
\usepackage{tikz}

\begin{document}

\begin{tikzpicture}[x=1.75cm,y=1.75cm]

    \path foreach \a in {0,...,3} {foreach \b in {0,...,3} {(-120:\a) +(\b,0) pic {p6mm}}};

    \path (-120:2) +(2,0)    pic {p2={$(-120:1)-(1,0)$}{$(-120:1)+(1,0)$}{8pt}};
    \node at ($(-120:2.5)+(0.1,0)$) {$\stackunder{p2}{\mathclap{\stackunder{\vec{a}-\vec{b},}{\vec{a}+\vec{b}}}}$};
    \path (-120:2.5) +(2,0)  pic {p2={$(-120:1)-(1,0)$}{$(-120:1)+(1,0)$}{8pt}};

\end{tikzpicture}

\end{document}
  ```
as shown in this screenshot
![Screenshot 2019-11-15 at 09 11 04](https://user-images.githubusercontent.com/371756/68927514-39822a80-0788-11ea-83d4-0692eff74542.png)

When I comment the line `\node...`, the fold for the environment `tikzpicture` changes what I do not expect
![Screenshot 2019-11-15 at 09 10 33](https://user-images.githubusercontent.com/371756/68927574-5b7bad00-0788-11ea-9454-aba84c249f1f.png)

**minimal.vim**

  ```vim
  set nocompatible
  let &runtimepath  = '~/.vim/bundle/vimtex,' . &runtimepath
  let &runtimepath .= ',~/.vim/bundle/vimtex/after'
  filetype plugin indent on
  syntax enable
let g:vimtex_fold_enabled = 1

  ```

lervag:
Hmm. I'm not sure how to properly solve this. But I had a related issue not too long ago, see #1515 and #1502. In particular, I think the best solution is to change the foldmarker or disable it:

```vim
" Either:
let g:vimtex_fold_types = {'markers': {'enabled': 0}}

" Or alternatively, something like this:
let g:vimtex_fold_types = {
      \ 'markers': {
      \  'open': '<<:',
      \  'close': ':>>',
      \ },
      \}
```

kiryph:
Sorry for not checking the issue tracker properly.

[...]

PR #1540, opened 2019-11-20
Enhance regex for fold end marker

927589452:
Instead of matching every line containing the fold end marker, this RE
matches only the last comment on a line, if the comment does not include
a command (Something started with a literal \).

Helps with #1536

Does not fix it for multiline commands, eg:
```LaTeX
% \emph{
%   \underline{
%     \thirdcommand{
%       content
% }}}
```

### DIFF
--- a/autoload/vimtex/fold/markers.vim
+++ b/autoload/vimtex/fold/markers.vim
@@ -20,7 +20,7 @@ let s:folder = {
       \}
 function! s:folder.init() abort dict " {{{1
   let self.re.start = '%.*' . self.open
-  let self.re.end = '%.*' . self.close
+  let self.re.end = '%[^\\\%]*' . self.close
   let self.re.text = [
         \ [self.re.start . '\d\?\s*\zs.*', '% ' . self.open . ' '],
         \ ['%\s*\zs.*\ze' . self.open, '% ' . self.open . ' '],


kiryph:
Another situation where this fails is:

```tex
\mycmd{} % for \mycmd see preamble }}}
```

I also think that this PR is not relevant for people changing the fold markers to `<<<` and `>>>` which is supported and lervag suggests that people should do.

IMHO allowing the user to pick a regex is better because finding one with which everyone can live is difficult.

927589452:
> Another situation where this fails is:
> 
> ```
> \mycmd{} % for \mycmd see preamble }}}
> ```
> 
I did not think of this scenario.
> I also think that this PR is not relevant for people changing the fold markers to `<<<` and `>>>` which is supported and lervag suggests that people should do.
> 
> IMHO allowing the user to pick a regex is better because finding one with which everyone can live is difficult.
I fear when doing this, you will get people defining their RegEx like yours, changing the fold marker, because someone suggested it to them, but it seems to be broken (because the RegEx is validated and it does not contain the foldmarker reference) to them.

lervag:
I'm responding to several comments now. Hope that's OK.

> [@kiryph] I have one idea: Could we allow to modify the regex for %.* }}} in ...
>
> I want to adjust the regex to exclude very likely candidates of commented commands as following:
>
> I am considering following two situations
>
> ...

> [@927589452] I would suggest going a slightly different route:
>
>     let self.re.end = '%[^\\\%]*' . self.close
>
> This just suppresses commented commands (and only the last comment is matched).
Did some quick tests with this here.

So, @kiryph, you propose to allow to modify the regexes individually, e.g. with something like this:

```vim
let g:vimtex_fold_types = {
      \ 'marker': {
      \  're': {
      \    'start': regex,
      \    'end': regex,
      \    'text': regex,
      \ }
      \}
```

This would definately be possible. But I don't think it is a good idea. The reason is that it will introduce a lot of entropy. In particular, the folding mechanism also relies on a single large regex in order to filter out the relevant lines. This is done to make the folding (much) more efficient. For the marker type folding, it means we need to define a separate regex that matches both the start and end marker.

However, it is already possible to modify the regexes. It does require more advanced customization, though:

```vim
let s:vimtex_fold_type_markers = {
      \ 'open': '{{{',
      \ 'close': '}}}',
      \ 're': {
      \   'start': regex,
      \   'end': regex,
      \   'text': regex,
      \   'fold_re': regex,
      \ },
      \}

function! s:vimtex_fold_type_markers.init() abort dict
  return self
endfunction

let g:vimtex_fold_types = {
      \ 'marker': s:vimtex_fold_type_markers,
      \}
```

Here you need to specify the four regexes properly. The `text` regex is used to create the fold text (see the main source code for more info). The `fold_re` regex is used for the large combined regex. I know that this is not really convenient, I'm just saying it is possible.

Back to the discussion at hand. @927589452 is proposing an update to the default regex that should work even if one changes the markers. To me, it seems a sensible approach. I don't really aim for "perfect", and I don't want to add too many options. So for me, it seems better to improve where one can and to settle for something that works well enough.

However, @kiryph raised an example where the proposed change actually fails. This is a good catch, and so now I'm not quite sure if I want to merge this PR as it stands. But if an update is proposed that improves the current regex while not introducing failing cases, then I will be glad to accept it!

kiryph:
Thanks for your elaborate response.

I agree that adding another option is always undesirable. I was struggling to come up with a good regex where almost no one stumbles upon.

My suggested regex `'%\({[^{]*}\)*[^{]*{\=[^{]*{\=[^{]*}}}'` is not good enough because it does not handle nested braces.

Since regex is not the tool for matching braces, we can only find a regex which works in most cases. I do not feel like my LaTeX knowledge is good enough to imagine all possible situations.

Right now, I have disabled fold markers in vimtex. If I want to use foldmarkers, I use them exclusively and set in the modeline the fold method to marker and do not use vimtex at all for folding.

lervag:
No problem. As said, any proposals for better regexes will definately be considered. But in the mean time, I've also disabled fold markers, as I don't really think it makes much sense to make "manual" folds like that.

I'll close this PR, but feel free to reopen or continue the discussion!

PR #1542, opened 2019-11-25
Add option to insert space after each imap

CooperWallace:
This pull request adds the ability to insert a space to the end of the RHS for each imap definition.

I wanted this option personally as it allows for one to use an imap macro, and then continue without pressing the spacebar. For example: ``` `qx``` shall be expanded to `\theta x`, rather than `\thetax` under the current implementation. 

By default, this option is disabled as to not affect current users. When it is toggled on, it affects the definition of all imaps.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -126,6 +126,7 @@ function! vimtex#init_options() abort " {{{1
 
   call s:init_option('vimtex_imaps_enabled', 1)
   call s:init_option('vimtex_imaps_disabled', [])
+  call s:init_option('vimtex_imaps_space', 0)
   call s:init_option('vimtex_imaps_list', [
         \ { 'lhs' : '0',  'rhs' : '\emptyset' },
         \ { 'lhs' : '6',  'rhs' : '\partial' },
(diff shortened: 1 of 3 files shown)


CooperWallace:
I've realized after implementing this that it solves the initial issue that I was having between imap macros and other types of variables. However, there is now an issue that one needs to insert backspace to add a subscript or superscript. I think that this tradeoff may not be exactly worth it with that in mind.

lervag:
> However, there is now an issue that one needs to insert backspace to add a subscript or superscript.

Right. I agree this tradeoff is not worth it.

Another solution to your issue is to define the maps manually. If you see in [`autoload/vimtex.vim`](https://github.com/lervag/vimtex/blob/0647bb8991a490f92de173ca67e9b49a3f1fddcd/autoload/vimtex.vim#L129-L194), you can find the predefined list of imaps. You could simply overwrite these with your own definitions where you add a space to the ones where you want them.

lervag:
My impression is that we don't add a new option in this case and that we close this PR. Do you agree?

CooperWallace:
I definitely do agree! Thank you for the time and consideration.

lervag:
No problem, happy to help!

PR #1578, opened 2020-01-26
add positioning for pdf viewer

tobiasweede:
Hi Karl,

Thank you for the great Plugin.

I (am not a i3 user and therefore) needed a way to position my zathura viewer after opening it.

Well, I ended up spending a while to figure it out. Would be great if it could go upstream (maybe in a modified version).

Especially default values should be checked and probably an on/off-switch added. Some additional documentation could also not be a bad idea. Let me know if you want me to help you with that despite my obvious lack of vimscipt knowledge :)

Best,
Tobias

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -295,6 +295,8 @@ function! vimtex#init_options() abort " {{{1
   call s:init_option('vimtex_view_skim_activate', 0)
   call s:init_option('vimtex_view_skim_reading_bar', 1)
   call s:init_option('vimtex_view_zathura_options', '')
+  call s:init_option('vimtex_view_zathura_move', '0 0')
+  call s:init_option('vimtex_view_zathura_size', '1366 742')
 endfunction
 
 " }}}1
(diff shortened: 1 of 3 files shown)


lervag:
Thanks! I appreciate your effort and the kind words!

> I (am not a i3 user and therefore) needed a way to position my zathura viewer after opening it.

I use Fluxbox myself, and I use my WM to control where new windows appear. I think this is generally a good approach.

> Well, I ended up spending a while to figure it out. Would be great if it could go upstream (maybe in a modified version).

I'm glad you did it. However, I don't want to add this as new options. Instead, we'll rely on something older and not so well documented. Here's what I want:

1. Move the `.move` and `.size` functions into `common.vim` and into the `xwin_template`, as these functions could equally well work for e.g. `MuPDF`.

2. Use `resize` instead of `size`.

3. Let the "targets" be in the arguments, not in options.

4. Make sure that the xwin IDs exist before doing anything.

Finally, when this is done, the new features may be used with the following setup:

```vim
let g:vimtex_view_zathura_hook_view = 'ViewHook'

function! ViewHook() abort dict
  call self.move('0 0')
  call self.size('1366 742')
endfunction
```

This hook feature is actually not really documented, so I'll make to do that as part of this PR.

tobiasweede:
I open a new, clean pull request.

lervag:
Sounds good, thanks!

PR #1582, opened 2020-01-28
Add support for acronym autocompletion

vyasr:
The glossaries package supports a variety of additional commands for filling in acronyms (see table 13.1 [here](http://mirrors.rit.edu/CTAN/macros/latex/contrib/glossaries/glossaries-user.html)). I was wondering if you would be interested in supporting these for autocompletion. I've done the absolute minimal hack right now to make it work, but could expand on this PR if you thought it would be useful.


lervag:
Thanks for this. I think the main thing that is lacking is a test. A good starting point is `test/tests/test-completion-glossary`. If you go there in a terminal and type `make`, the current tests should run. The test setup should be quite simple to follow, at least if you know some Makefile syntax, bash scripting and some vim script. I think you could add a test by editing the file `glossaries.vim` and add lines similar to this:

```vim
let s:candidates = vimtex#test#completion('\ac{', '')
call vimtex#test#assert_equal(len(s:candidates), <number>)
```

It would also be nice to see a simple and concrete example of what this feature is supposed to fix. Preferably with a reference to the glossaries documentation. I.e., list the commands that should trigger glossary comletion that are currently missing.

vyasr:
Got it, thanks for the quick response. I'm pretty familiar with make and bash, but I'm a novice with vim scripting. I use vim pretty much exclusively, but I've never developed any plugins myself, so I'll come back with questions if I run into issues.

For starters: what is `chronic`? I can't run tests without that executable, but I'm not sure where to get it. Also, do the tests depend on any neovim specific features? My executable is just vim, but I'm not sure if a simple sed replacement of `nvim --headless` in the tests would cause any problems (without committing those changes back, of course).

I'll open an issue to track this feature request.

lervag:
`chronic` comes from moreutils and is used to ignore all output except when there is an error.

And yes, sorry: With Vim you should use `MYVIM="vim -T dumb --not-a-term --noplugin -n" make`. Sorry for the inconvenience!

lervag:
> I'll open an issue to track this feature request.

No need to raise another issue, I think. Let's keep the discussions in this thread.

vyasr:
Whoops. I apologize, I just created the issue as you messaged. I can close it if you'd prefer.

vyasr:
I will say that I do think there's some benefit to getting other opinions on what commands to support autocompletion for. However, nothing prevents us from doing that on the PR directly if you prefer.

lervag:
No, sorry, disregard my message.

vyasr:
Thanks for the info. I went ahead and installed these. There were a couple other minor problems, some of which I was able to get around (see #1584), but the tests in test-compiler currently fail for me for most of the options. In general, turning off the background or callback switches seems to cause compilation to fail for me; only the first set of options in test-compiler/test.vim:92 run successfully. Any ideas why that might be? My `vim --version` shows `+job` and `+clientserver`.

lervag:
> Thanks for the info. I went ahead and installed these. There were a couple other minor problems, some of which I was able to get around (see #1584)

Thanks for noticing and fixing this. That PR is merged now.

> ..., but the tests in test-compiler currently fail for me for most of the options. In general, turning off the background or callback switches seems to cause compilation to fail for me; only the first set of options in test-compiler/test.vim:92 run successfully. Any ideas why that might be? My vim --version shows +job and +clientserver.

Correct me if I'm wrong, but this is not really relevant to the current thread. I have one idea that might explain it, though: The compiler test also runs `latexrun` and `arara`. If these are not installed, then the tests will fail. Feel free to open a new issue regarding this, as it might indicate that we need a new update to the CONTRIBUTION.md file.

vyasr:
You're right, it's not relevant except insofar as I need to know if running those tests is necessary to verifying that my changes work; otherwise, the only way I have to test is to push and rely on CI. I'll see if I can resolve or at least identify the issue more clearly before I open a new issue.

vyasr:
It looks like we have a reasonable path forward from #1583. Could you help me understand how some of the different files fit together so that I know how this works? The lines I changed in complete.vim add `ac` as another string to be recognized at the beginning of the regex, so I'm assuming that the `s:complete_gls` regex is getting matched to all the other gls* features. How does the `complete/glossaries` file relate to this? Do I need to do anything in that file, or is it just a matter of updating the regex in `complete.vim` to also include the `[cpdr]?`?

lervag:
> You're right, it's not relevant except insofar as I need to know if running those tests is necessary to verifying that my changes work; otherwise, the only way I have to test is to push and rely on CI. I'll see if I can resolve or at least identify the issue more clearly before I open a new issue.

Good point; although I would not be upset if you only cared about "your own" tests. That is, as a first step, add tests in `test/tests/test-completion-glossary` as I mentioned and ensure all these tests pass.

> It looks like we have a reasonable path forward from #1583. Could you help me understand how some of the different files fit together so that I know how this works?

Yes, of course, I'll be happy to help.

> The lines I changed in `complete.vim` add `ac` as another string to be recognized at the beginning of the regex, so I'm assuming that the `s:complete_gls` regex is getting matched to all the other gls* features.

Yes, I think this is close to what we want. According to #1583, it seems you should use `|acr|Acr|ACR`, as all the acronym commands seem to be prefixed with either `acr`, `Acr`, or `ACR`. (Similar updates apply to the `vimtex#re#...` part as well.)

> How does the `complete/glossaries` file relate to this?

It is not quite related: This file should include the macros provided by the glossaries package for command completion.

> Do I need to do anything in that file, or is it just a matter of updating the regex in `complete.vim` to also include the `[cpdr]?`?

I think it should suffice to update the regexes. It seems something like this should suffice:

```
'\v\\([cpdr]?%(gls|Gls|GLS)|acr|Acr|ACR)\w*\s*\{[^}]*$'
```

vyasr:
I see, so the `complete/glossaries` file controls what happens if I type `\gl<Tab>`, whereas the regex in `complete.vim` controls what happens if I type `\gls{`?

I'll update the regexes once I have a chance to write appropriate tests. I should be back to writing early next week and will get this working then. Thanks for your help!

lervag:
> I see, so the complete/glossaries file controls what happens if I type \gl<Tab>, whereas the regex in complete.vim controls what happens if I type \gls{?

Yes, precisely; sort of. Command completion is also specified in `complete.vim`, but it uses the list of macros specified in the files under `complete/...` as a source.

vyasr:
@lervag sorry for the delay, but I finished up the changes. Two notes:

1. Instead of using the `acr` variants, I used `ac` to also capture the shortcuts (glossaries supports using `\ac` instead of `\gls`, for example).
2. I noticed one other issue while completing this. For some reason, the regexes to capture the additional parts of the `gls*` commands were not properly matching. For example, `\glsname{` was not popping up a menu. I was able to fix this by replacing the `\a*` match with an explicit `[a-zA-z]*` match. I'm not sure if this is a portability issue because I'm running vim 8.1 on a Mac, but I figured that the explicit character class (while more verbose) is guaranteed to work everywhere so this change shouldn't cause problems. I matched this change for neocomplete, deoplete, and ncm2, which had small variants on the same thing (deoplete was using `\w*`, which oddly enough works for ncm2 but is not what we want since it would also match `\gls1{`, for example). Let me know if you think there are problems with these changes.

One small question. I noticed (via trial and error) that in re.vim the ncm2 section was the one that was being used in my case. However, since I'm running on vim, not neovim, I don't have ncm2 installed. Is that set of labels used as the default when no other completion engine is available? I'm using YouCompleteMe, so perhaps it's registering the same thing under the hood?

autoload/vimtex/complete.vim
```diff
@@ -617,7 +617,7 @@ endfunction
 " {{{1 Glossary
 
 let s:completer_gls = {
-      \ 'patterns' : ['\v\\(gls|Gls|GLS)\w*\s*\{[^}]*$'],
+      \ 'patterns' : ['\v\\([cpdr]?(gls|Gls|GLS)|ac|Ac|AC)[a-zA-z]*\s*\{[^}]*$'],
```
lervag: I think this is close, but I am afraid that this will match way too much. For instance, `\acute{` would now complete glossaries. I think an easy solution is something like this:

```vim
       \ 'patterns' : [
       \   '\v\\([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*\s*\{[^}]*$',
       \   '\v\\(ac|Ac|AC)\s*\{[^}]*$',
       \ ],
```

autoload/vimtex/re.vim
```diff
@@ -39,7 +39,7 @@ let g:vimtex#re#neocomplete =
       \ . '|hyperref\s*\[[^]]*'
       \ . '|includegraphics\*?%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
       \ . '|%(include%(only)?|input|subfile)\s*\{[^}]*'
-      \ . '|\a*(gls|Gls|GLS)(pl)?\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
+      \ . '|[a-zA-z]*([cpdr]?(gls|Gls|GLS)|ac|Ac|AC)[a-zA-z]*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
```
lervag: For this and other regexes in `re.vim`:
- I don't think we need to match the initial `\a*` here, right?
- Should also update similarly as above to avoid matching `\acute` and other non-related commands.

lervag:
Also, very good work. I'm glad you are doing this!

autoload/vimtex/complete.vim
vyasr: Good call, I'll fix that.

autoload/vimtex/re.vim
vyasr: I agree, I wasn't sure why the initial match was there but didn't want to make that change unilaterally.

vyasr:
No problem, always happy to contribute to projects I use! I addressed the issue with overmatching `\ac`.  Regarding YCM, you're right about the setting I'm using. I just noticed these lines in the `re.vim` file:
```
let g:vimtex#re#ncm = copy(g:vimtex#re#ncm2)
let g:vimtex#re#youcompleteme = map(copy(g:vimtex#re#ncm), "'re!' . v:val")
```
so that answers my questions about how that is working for me.

Regarding `[a-zA-z]` vs `\a`, I should have been more specific. I have used `\a` in regexes before and it works fine, and it still does. It's specifically in the context of this regular expression that it doesn't matching. Reusing my original example, swapping 

```'\\([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)[a-zA-A]*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',```

with 

```'\\([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*',```

in re.vim causes `\glsname{`to not autocomplete. If I simply type `/\a` in vim, though, I can match any letters. However, I added the test as you requested, and the test passes with both versions. That suggests that maybe it's something conflicting with my own vim configuration, so I'll try and isolate the problem.

**Comment by vyasr** [comment]
I've reduced my configuration as much as possible. Here's my ~/.vimrc:

```
call plug#begin('~/.vim/plugged')
Plug 'https://github.com/lervag/vimtex'
Plug 'https://github.com/Valloric/YouCompleteMe.git'
call plug#end()

if !exists('g:ycm_semantic_triggers')
let g:ycm_semantic_triggers = {}
endif
au VimEnter * let g:ycm_semantic_triggers.tex=g:vimtex#re#youcompleteme
```

I've temporarily moved almost everything out of my `.vim` folder. All that's left is `.vim/autoload/plug.vim`, and the actual plugins in `.vim/plugged/`, but I've remove those `Plug *` lines from my vimrc, so I'm assuming they won't have any effect. Any thoughts on what else could be causing this problem? I don't know of any other files that could be affecting the behavior of this regex, especially since it only seems to be a problem in the context of the vimtex match but not normal regex searches in vim.

**Comment by lervag** [comment]
> No problem, always happy to contribute to projects I use! I addressed the issue with overmatching `\ac`.  Regarding YCM, you're right about the setting I'm using. I just noticed these lines in the `re.vim` file:
> ```
> let g:vimtex#re#ncm = copy(g:vimtex#re#ncm2)
> let g:vimtex#re#youcompleteme = map(copy(g:vimtex#re#ncm), "'re!' . v:val")
> ```
> so that answers my questions about how that is working for me.

Yes, exactly :)

> Regarding `[a-zA-z]` vs `\a`, I should have been more specific.
> ...
> If I simply type `/\a` in vim, though, I can match any letters. However, I added the test as you requested, and the test passes with both versions. That suggests that maybe it's something conflicting with my own vim configuration, so I'll try and isolate the problem.

This is very strange. Does `\v\\([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*` match `\glsname{` if you use it as a search term directly (type it letter for letter)?

> I've reduced my configuration as much as possible. Here's my ~/.vimrc:
> ...
> Any thoughts on what else could be causing this problem?

No, not really. Is it possible for you to test with neovim as well, or any other Vim version?

File: autoload/vimtex/re.vim [hunk 3]
```diff
@@ -57,7 +58,8 @@ let g:vimtex#re#deoplete = '\\(?:'
       \ . '|hyperref\s*\[[^]]*'
       \ . '|includegraphics\*?(?:\s*\[[^]]*\]){0,2}\s*\{[^}]*'
       \ . '|(?:include(?:only)?|input|subfile)\s*\{[^}]*'
-      \ . '|\w*(gls|Gls|GLS)(pl)?\w*(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
+      \ . '|([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
```
lervag: I don't think `\a` is valid in Python regexes. You should use `\w` or `[a-zA-Z]` instead.

vyasr:
Yes, using it as a directly as a search term seems to work... so I'm not sure why it specifically doesn't work in this context. I have tried this on two separate machines running OSX with the same results. I unfortunately don't have an easy way to test this using a separate Vim installation on these machines due to some limitations with the Python environments I'm using, but I have access to a Linux machine so I'm going to try it there.

vyasr:
OK, I can confirm that the problem persists on a Linux machine. My local vim is 8.1, the Linux machine uses vim 8.2. I can also confirm that the switch to `[a-zA-Z]` also works on this machine.

lervag:
Ok, this is strange. Can I suggest we merge this PR as it is now and that you open an issue regarding this problem? Then, if we find a solution or decide to avoid `\a`, I could either fix things directly myself or you could open a new PR just for this. Note, I'll not mind to change from `\a` to `[A-Za-z]` if we don't find any solution for this.

lervag:
In the new issue, please include a minimal example for this particular issue. Then I could try to reproduce it myself and see if I can figure it out.

vyasr:
That seems fine to me. I've addressed your last change request, so once tests pass feel free to consider this complete. I'll open the other issue now.

autoload/vimtex/re.vim
```diff
@@ -39,7 +39,8 @@ let g:vimtex#re#neocomplete =
       \ . '|hyperref\s*\[[^]]*'
       \ . '|includegraphics\*?%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
       \ . '|%(include%(only)?|input|subfile)\s*\{[^}]*'
-      \ . '|\a*(gls|Gls|GLS)(pl)?\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
+      \ . '|([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'
```
lervag: I just realized that this regex is not the same as the one in `completion.vim`. The following shows the difference:

```
'([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*\s*\{[^}]*$',
'([cpdr]?(gls|Gls|GLS)|acr|Acr|ACR)\a*%(\s*\[[^]]*\]){0,2}\s*\{[^}]*'

only in re.vim (this matches 0-2 sets of options):
=> %(\s*\[[^]]*\]){0,2}'
```

This is the same also for the following regexes.

lervag:
Sorry about the inconvenience, but I had one more comment. :\

autoload/vimtex/re.vim
vyasr: Yes I saw that but didn't change it since I wasn't sure if there was a reason they differed. I also don't actually know what kinds of optional arguments are possible, so it wasn't clear which version was "correct". Which version do you want?

lervag:
>
>  Which version do you want?
>
The one we use in completion.vim.

>

vyasr:
OK, one last question that I meant to ask earlier but forgot :) I noticed that the neocomplete regexes have a lot of percent signs floating around. Do those have a specific meaning in that context, or are they extraneous?

lervag:
> OK, one last question that I meant to ask earlier but forgot :) I noticed that the neocomplete regexes have a lot of percent signs floating around. Do those have a specific meaning in that context, or are they extraneous?

Yes, good question. The percentages are (slightly) meaningful. The difference between `%(` and `(` is that the first creates a group that you can't refer to. This is slightly more efficient. See `:h \%(`. Feel free to add them where appropriate (the `g:vimtex#re#{completor}` regexes do not need to remember any groups).

vyasr:
Got it, thanks for the info. I ended up not adding them because I had some trouble getting the non-capturing groups to work (similar issues with the completion not showing up), so I decided to just try and wrap this up for now so that the key features are enabled. If we can get them working as part of #1594 that would probably make sense.

lervag:
Great, thanks! This is merged now. Next we'll look at #1594.

PR #1584, opened 2020-01-28
Fix usage of vim and add some notes on running tests

vyasr:
The glossaries test was hardcoded to use `nvim --headless` instead of `MYVIM`, so it failed even when the variable was overridden.

I also made some notes in the CONTRIBUTING document on how to run tests. I'm happy to edit as  you see fit.


lervag:
Thanks for fixing the error and suggesting the addition to the CONTRIBUTION document. I made some minor change to the text.

PR #1589, opened 2020-02-01
doc: Note that b:vimtex_main has to be set before the buffer is loaded

latricewilgus:
Document that b:vimtex_main has to be set prior to loading the buffer. Otherwise it will have no effect.

Also mention the workaround with :VimtexReloadState to make vimtex aware of a change in b:vimtex_main.


lervag:
Thanks for this!

Issue #1561, opened 2019-12-29
Missing completion of new commands and environments in custom classes & packages


GergelyKalmar:
**Issue**
I often use custom classes in my tex files, however, it seems that packages loaded in these class files are not picked up by vimtex. 

Let's have in ``custom.cls`` for example:

```
\NeedsTeXFormat{LaTeX2e}
\ProvidesClass{custom}[2020/01/01 v1.0 Custom class]
...
\RequirePackage{amssymb, amsmath, amsthm}
...
```

Then while using it in ``test.tex`` (being in the same folder):

```
\documentclass{custom}
\begin{
```

... ``align`` does not show up in the YCM autocompletion list. If I explicitly load amsmath with ``\usepackage{amsmath}`` in ``text.tex`` it works as expected. I see similar problems with everything defined in class files (e.g. new commands, new environments and so on) being ignored.

Is this a known limitation or am I doing something wrong?

I'm using vim 8.1.1401 and vimtex from around October.

lervag:
Does it work if you open `test.tex` when it has been compiled? I think it does. Vimtex does not itself parse every documentclass and package to create the list of packages. It only searches the preamble for all packages, and then uses the `.fls` file if it is available. This file is created during compilation and gives a list of all packages that are loaded for the given project. Thus, if you do this on your project, I think it should work:

1. Open `test.tex` with Vimtex
2. Compile it
3. Close Vim and reopen `test.tex`
4. Now the completion (and other things) will work as expected

Finally, note that you can simplify step 3: You can either do `:edit!` to completely reload the file, which should 1) stop any running compilers, and 2) reload vimtex. You can also use `:VimtexReload`, which does mostly the same (except it only reloades vimtex specific stuff).

lervag:
> I'm using vim 8.1.1401 and vimtex from around October.

You should update Vimtex before you post an issue. I don't think it would be very relevant in this case, but in general, it is good practise. Else you might complain about a bug that is already fixed.

GergelyKalmar:
Thanks for the reply! No, it does not work after compilation - it might be because I'm using latexmk with ``output-directory=.aux``, which causes the ``.fls`` files to end up in that directory. Is there a potential workaround for such cases?

[...]

PR #1591, opened 2020-02-04
Complete commands from custom packages and classes

lervag:
This addresses #1561 and is currently a work in progress.

**Remaining actions**
- [x] Improve completion by parsing custom classes and packages, that is, any `.cls` and `.sty` files that are not already available as precompiled lists.
  - [x] Collect macros defined by `\newcommand`s for command completion.
  - [x] Collect environments defined by `\newenvironment`s for env completion.

Any input or comment is appreciated!


lervag:
@GergelyKarmar I would be happy if you had the time to test this. It should currently parse your `custom.cls` for any `\newcommands`. Definitions from `\let`s and `\def`s are not yet parsed (not sure if we should?). Environments are also not parsed yet.

GergelyKalmar:
hi @lervag, thank you very much for your work! I'll check it over the weekend and will let you know how it worked.

GergelyKalmar:
Once I re-generated the aux files it seems to work great! However, when I open a tex file and start typing a command (e.g. \begin) while using one of my more complex custom classes it seems to take a few seconds for the completions to pop up for the first time. During those few seconds the cursor seems to disappear as well. Can I do something about this or is this expected?

lervag:
> Once I re-generated the aux files it seems to work great! However, when I open a tex file and start typing a command (e.g. `\begin`) while using one of my more complex custom classes it seems to take a few seconds for the completions to pop up for the first time. During those few seconds the cursor seems to disappear as well. Can I do something about this or is this expected?

Partly expected. I noticed that this had an impact on the completion speed, but it was difficult to assess how bad it was. It is cached, so the delay should only occur the first time. Still, I don't really like it.

If you could provide an example where the delay is substantial, then I could look into ways to optimize this. E.g., perhaps it is possible generate the cache in the background when Vim starts.

GergelyKalmar:
Sure! Here would be a more complete class file: [custom.zip](https://github.com/lervag/vimtex/files/4185543/custom.zip). If I compile ``test.tex`` and open it in vim it takes a few seconds for the first suggestion to pop up. Responsiveness of vim is low and CPU usage is quite high for those few seconds. I suppose the user experience can be improved by populating the cache in the background when a tex file is opened.

lervag:
I can reproduce this. For me, it takes about 3 seconds the first time. After that it is quick due to caching. The main reason is that `kpsewhich` is slow. I use this to find the path of packages/classes. In the test it takes about 2.6 seconds to run it 29 times, thus about 86% of the time.

Possible ways to fix this:
1. Add persistant caching of kpsewhich calls. This means that at least once, things will be slow, but after that vimtex may read the cache and update it once in a while. This should not be too difficult to implement and other parts may benefit from it.
2. Simplify and restrict this feature to only locally available classes (i.e. in the same folder as the main latex file).

Do you have thoughts on this?

lervag:
Ok, I implemented a first version of caching. This update exposes an option `g:vimtex_cache_root` to specify where you want your cache files, by default at `~/.cache/vimtex/`. The `vimtex#kpsewhich#find` calls are now cached and should be much faster. For me, I am now closer to 0.4 seconds on the test you gave me. I could look into optimizing even more if you think it is worth the effort.

GergelyKalmar:
@lervag Ah, looks very nice, thank you! I will test it in the evening or tomorrow.

I would have also chosen option 1 given that I usually install the custom classes in my texmf folder.

lervag:
> @lervag Ah, looks very nice, thank you! I will test it in the evening or tomorrow.

Great, thanks! I look forward to getting some feedback. I think the current version should work quite well, but before I merge this I also want to implement environment completion. Perhaps also completion of macros defined by `\let` and `\def`.

> I would have also chosen option 1 given that I usually install the custom classes in my texmf folder.

Yes, I thought so. But I was curious how easy it would be to implement the caching, so I did it mostly for fun. It seems to work well and to be useful, even if we should end up doing something like option 1.

GergelyKalmar:
Looks great to me! The first startup is indeed still laggy but after that the delay is not really noticeable anymore. I think I could use it pretty much as it is - helps a lot with my custom classes. Lovely!

lervag:
> Looks great to me! The first startup is indeed still laggy but after that the delay is not really noticeable anymore. I think I could use it pretty much as it is - helps a lot with my custom classes. Lovely!

Great, happy to hear it. I'll do some more adjustments now then merge this after.

GergelyKalmar:
@lervag sorry for the late response, I've checked it now in master and works like charm! Thank you very much for this feature and your commitment. If you're ever in Zurich and feel like having a beer/wine just ping me! :)

lervag:
Great, I'm happy to hear it! If I do visit Zurich (probably not in near future), I might take you up on the offer! :)

As always, feel free to raise further feature requests/issue reports.

PR #1605, opened 2020-02-21
add pdf viewer positioning functions

tobiasweede:
Follow up on #1578

--

Hej Karl,

finally I got my very little change working. Learned a few things about vimscript I would say.

In order to have less commits here is the new pull request.

Maybe you find a better place to document it. I added a note in the hook section.

Let me know if something is still wrong / you want me to change something.

Best,
Tobias


lervag:
Thanks! I think this was very close to sufficient, so I just made some minor modifications myself.

I appreciate your effort and I'll be happy to help with future PR's as well!

PR #1611, opened 2020-02-28
Add an imap for Feynman slash notation

rsekman:
```
`/p -> \slashed{p}
```

One can also imagine similar mappings like `` `_v -> \mathbf{v} ``.

### DIFF
--- a/autoload/vimtex/imaps.vim
+++ b/autoload/vimtex/imaps.vim
@@ -149,6 +149,11 @@ function! vimtex#imaps#wrap_environment(lhs, rhs) abort " {{{1
   return l:return
 endfunction
 
+function! vimtex#imaps#feynmanslash()
+  let l:c = getchar()
+  execute "normal a\\slashed{" . nr2char(l:c) . "}\<Esc>"
+endfunction
+
 " }}}1
 
 "
--- a/autoload/vimtex/imaps.vim
+++ b/autoload/vimtex/imaps.vim
@@ -99,9 +99,12 @@ function! s:create_map(map) abort " {{{1
     endif
     let b:vimtex_context[l:key] = a:map.context
   endif
+  if ! get(a:map, 'expr', 0)
+    let a:map.rhs = string(a:map.rhs)
+  endif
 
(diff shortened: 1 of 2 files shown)


autoload/vimtex/imaps.vim
```diff
@@ -99,9 +99,12 @@ function! s:create_map(map) abort " {{{1
     endif
     let b:vimtex_context[l:key] = a:map.context
   endif
+  if ! get(a:map, 'expr', 0)
```
lervag: I prefer:

```vim
   if !get(a:map, 'expr') 
```

The 0 is default and the space after `!` is unnecessary, IMHO.

lervag:
I like this idea. It is simple and provide some extra utility. However, I'm curious, is it possible to make it slightly more "general"? If not, then I'd prefer to keep the Feynman specific stuff out and instead only add the part that allows the functionality. This allows you to add the Feynman map in your own custom configuration without too much inconvenience.

Note: The tests do not pass. So there is something that is not quite right yet.

lervag:
The tests fail due to a minor error on my part. I'll push a fix for it now, please rebase your work on top of new master (i.e. with `git rebase master` then `git push --force`).

rsekman:
> I like this idea. It is simple and provide some extra utility. However, I'm curious, is it possible to make it slightly more "general"? If not, then I'd prefer to keep the Feynman specific stuff out and instead only add the part that allows the functionality. This allows you to add the Feynman map in your own custom configuration without too much inconvenience.

I agree, there are several similar applications, like, `\mathbf, \mathfrak, \overline, \mathcal`. The reason I coded it like this is I like how vimtex's imaps activate only in math mode.

The general pattern is `c -> \$command{c}`, which my function is easily extended to handle. So perhaps we could consider a new "class" of imaps `{'lhs': '/', 'rhs': 'slashed', }` -- maybe with a prefix different from `\``, since most of those are already taken. I don't think `#` is used much in math mode, so that could be an idea.

rsekman:
Closing in favor of #1612 which is more general.

lervag:
Great, thanks.

PR #1612, opened 2020-02-29
Add imaps of the type <leader>xc -> \$command{c}.

rsekman:
Useful for commands such as mathbf, mathfrak, mathcal.  For now, specify these like
`    \ { 'lhs' : 'b',  'rhs' : 'vimtex#imaps#style("mathbf")', 'expr' : 1, 'leader' : '#'},`
where `'expr'` makes vimtex evaluate the rhs as an expression, and `vimtex#imaps#style` is a helper function.

Refactoring and generalization of #1611.

### DIFF
--- a/autoload/vimtex/imaps.vim
+++ b/autoload/vimtex/imaps.vim
@@ -155,6 +155,9 @@ function! vimtex#imaps#wrap_environment(lhs, rhs) abort " {{{1
 endfunction
 
 function! vimtex#imaps#style(command)
+  if !s:is_math()
+    return ''
+  endif
   let l:c = getchar()
   return '\' . a:command . '{' . nr2char(l:c) . '}'
 endfunction
--- a/autoload/vimtex/imaps.vim
+++ b/autoload/vimtex/imaps.vim
@@ -101,9 +101,12 @@ function! s:create_map(map) abort " {{{1
     endif
     let b:vimtex_context[l:key] = a:map.context
   endif
+  if !get(a:map, 'expr')
+    let a:map.rhs = string(a:map.rhs)
+  endif
 
   silent execute 'inoremap <expr><silent><nowait><buffer>' l:lhs
-        \ l:wrapper . '("' . escape(l:lhs, '\') . '", ' . string(a:map.rhs) . ')'
+        \ l:wrapper . '("' . escape(l:lhs, '\') . '", ' . a:map.rhs . ')'
 
   let s:created_maps += [a:map]
 endfunction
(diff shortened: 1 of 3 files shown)


lervag:
This looks well done. I see one minor issue: Writing `#b` and similar outside math mode now "glitches" because the mapping is still activated. The expansion is not done, since it is stilled wrapped in math mode, but the `getchar` function is executed and ignored.

If you want to look into it, then I'll wait before I merge. Else I think it is probably not something people will notice and care about. That is, I'll gladly merge this as is, but I will first let you consider if it is possible to avoid the minor inconvenience.

autoload/vimtex/imaps.vim
```diff
@@ -155,6 +155,9 @@ function! vimtex#imaps#wrap_environment(lhs, rhs) abort " {{{1
 endfunction
 
 function! vimtex#imaps#style(command)
+  if !s:is_math()
+    return ''
+  endif
```
lervag: Yes, but: What if we wanted to use this for non math stuff? E.g. `\emph{...}`?

rsekman: We could use another helper function, perhaps rename this one to `vimtex#imaps#style_math`.

I guess ideally the wrapper should prevent the rhs from being called, but I think this requires passing in function handles, testing for this in every wrapper, and so on.

lervag: I think the simplest solution is to simply rename the function. I'll merge this now and just do that. Pragmatism is imporant to avoid over engineering.

lervag:
Thanks, I've merged this now. I made very minor modifications and generally appreciate that you've adapted so well to my style when adding this. Also, very good work, this seems like it could be useful.

Btw: Perhaps we should also update the docs?

autoload/vimtex/imaps.vim
rsekman: Good! However, the name of the function also needs to be changed in the pre-configured imaps. I'll push a commit with this change.

postylem:
I was trying to make a new imap macro for easily typing characters in `\mathrm` when in math mode, and stumbled around a lot before finding I needed to have `'expr' = 1` in the map.  I'm not sure if I've looked in the right places, but is this mentioned in the docs anywhere?  

It might be useful to have an example in the docs for neophytes like me that don't understand what they're doing: If you want to make it so typing `#rX` will result in the string `\mathrm{X}`, put the following in tex.vim:

```vim
call vimtex#imaps#add_map({
			\ 'lhs' : 'r',
			\ 'rhs' : 'vimtex#imaps#style_math("mathrm")',
			\ 'expr' : 1,
			\ 'leader' : '#',
			\ 'wrapper' : 'vimtex#imaps#wrap_math'
			\})
```

lervag:
Thanks, I've pushed an update to the docs.

PR #1615, opened 2020-03-03
Avoid SyntaxWarning on Py38.

anntzer:
`n[k] is not 0` results in
`SyntaxWarning: "is not" with a literal. Did you mean "!="?`
on Py3.8.


lervag:
Thanks!

PR #1621, opened 2020-03-09
add prose linters

Konfekt:
In a similar vein to the syntax linters, these TeX prose linters might be of use for little effort.

### DIFF
--- a/compiler/style-check.vim
+++ b/compiler/style-check.vim
@@ -1,7 +1,8 @@
 if exists("current_compiler") | finish | endif
 let current_compiler = "style-check"
 
-if exists(":CompilerSet") != 2		" older Vim always used :setlocal
+" older Vim always used :setlocal
+if exists(":CompilerSet") != 2		
   command -nargs=* CompilerSet setlocal <args>
 endif
 
--- a/compiler/style-check.vim
+++ b/compiler/style-check.vim
@@ -1,0 +1,19 @@
+if exists("current_compiler") | finish | endif
+let current_compiler = "style-check"
+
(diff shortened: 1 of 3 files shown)


compiler/style-check.vim
```diff
@@ -0,0 +1,19 @@
+if exists("current_compiler") | finish | endif
+let current_compiler = "style-check"
+
+if exists(":CompilerSet") != 2		" older Vim always used :setlocal
```
lervag: I prefer to put comments above instead of inline.

compiler/textidote.vim
```diff
@@ -0,0 +1,28 @@
+if exists("current_compiler") | finish | endif
+let current_compiler = "textidote"
+
+if exists(":CompilerSet") != 2		" older Vim always used :setlocal
```
lervag: I prefer to put comments above instead of inline.

compiler/textidote.vim
```diff
@@ -0,0 +1,28 @@
...
+  command -nargs=* CompilerSet setlocal <args>
+endif
+
+let s:cpo_save = &cpo
+set cpo&vim
+
+if exists('g:textidote_jar') && filereadable(fnamemodify(g:textidote_jar, ':p'))
```
lervag: Is there a reason you do not use the `g:vimtex_...` naming scheme here?

Konfekt: No, just an oversight on transforming this from a general compiler into on that becomes part of the vimtex plug-in.

lervag:
Thanks for this!

Issue #1390, opened 2019-05-22
feature idea: toggle fraction


nilsbecker:
a feature that would ease editing a lot for me is the following. an automatic conversion between inline fractions `$a/b$` and display fractions `$\frac{a}{b}$`. ideally, this would be a command similar to `tsd` now, maybe `tf`? 

the direction from display to inline is relatively easy:
- invocation should work anywhere on the fraction command.
- first argument
- if it is not surrounded by parentheses and consists of more than one word, surround by parentheses
- "/"
- second argument
- same parentheses treatment as for first argument.
here the question is what a 'word' is. i think a simple definition like `W` in vim would be already useful.

the direction from inline to display involves deciding from where to invoke. 
- if there is a marked region, split that on '/' and make the first part the first argument and the second part the second argument. 
- if there is no marked region, the command can only work either when the cursor is on the numerator or on the denominator. a choice has to be made. let's say the denominator. then: search backwards to the '/'. search forward to 'E', the end of the current 'W' word. this span is the second argument. from the '/' search further backwards to the 'B' of the previous 'W'. 
- remove surrounding parentheses in the arguments if present
- output the `\fraction` command

does that sound within scope?

kiryph:
This could also be something for the plugin https://github.com/AndrewRadev/switch.vim

Following is untested but should be close:
```viml
autocmd FileType tex let b:switch_custom_definitions =
      \ [
      \   {
      \     '$\(\S*\)/\(\S*\)$':        '$\frac{\1}{\2}$',
      \     ' \(\S*\)/\(\S*\) ':        ' \frac{\1}{\2} ',
      \   }
      \ ]
```

nilsbecker:
interesting plugin! thanks for pointing it out. i'm unsure where this functionality should go, vimtex or another plugin

nilsbecker:
i tried it out, now i have this:
```
autocmd FileType tex let b:switch_custom_definitions =
      \ [
      \   {
      \     '\(\S*\)/\(\S*\)' : '\\frac{\1}{\2}',
      \     '\\frac{\(\S*\)}{\(\S*\)}': '\1/\2' 
      \   }
      \ ]
```
which does not have the parentheses logic yet; also space on either side of the `/` is not recognized. so far so good.

nilsbecker:
... but it does not respect math environment delimiters if they are adjacent without space. i guess the definition of what a 'word' is would have to be refined considerably.

lervag:
[...]

PR #1625, opened 2020-03-11
Add tsf for toggling inline vs displayed fractions

lervag:
I've finally made a prototype version of this feature (cf. #1390). It is not going to be perfect (I won't fix all edge cases), but it is probably still not quite ready. Please feel free to test and provide feedback!

It would be good to get some example of inline sentences that does not work as expected. Examples should be given as:

1. A single line of text in which fraction toggle should work.
2. The position in the line where the toggle is activated.
3. The expected result.
4. The observed result.

PS! Sorry for taking so very long to work on this issue! However, I think this should be not so far from finished now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ disabled if desired.
     `csc`/`cse`/`cs$`/`csd`
   - Toggle starred command or environment with `tsc`/`tse`
   - Toggle between e.g. `()` and `\left(\right)` with `tsd`
+  - Toggle (inline) fractions with `tsf`
   - Close the current environment/delimiter in insert mode with `]]`
   - Insert new command with `<F7>`
   - Convenient insert mode mappings for faster typing of e.g. maths
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -491,6 +491,7 @@ function! s:init_default_mappings() abort " {{{1
   call s:map('n', 'dsc',  '<plug>(vimtex-cmd-delete)')
   call s:map('n', 'csc',  '<plug>(vimtex-cmd-change)')
   call s:map('n', 'tsc',  '<plug>(vimtex-cmd-toggle-star)')
+  call s:map('n', 'tsf',  '<plug>(vimtex-cmd-toggle-frac)')
   call s:map('i', '<F7>', '<plug>(vimtex-cmd-create)')
   call s:map('n', '<F7>', '<plug>(vimtex-cmd-create)')
   call s:map('x', '<F7>', '<plug>(vimtex-cmd-create)')
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -22,6 +22,9 @@ function! vimtex#cmd#init_buffer() abort " {{{1
 
   nnoremap <silent><buffer> <plug>(vimtex-cmd-toggle-star)
         \ :<c-u>call <sid>operator_setup('toggle_star')<bar>normal! g@l<cr>
+
+  nnoremap <silent><buffer> <plug>(vimtex-cmd-toggle-frac)
+        \ :<c-u>call <sid>operator_setup('toggle_frac')<bar>normal! g@l<cr>
 endfunction
 
 " }}}1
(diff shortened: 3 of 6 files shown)


lervag:
I'll take the liberty of tagging some possibly interested users:
@nilsbecker @clason @kiryph

nilsbecker:
how do i test this? i clone the repo, switch to the toggle-fractions branch and then? how do i install it in vim?

lervag:
Sorry, should have said! There are multiple ways to do it.

1. If you use vim-plug, then you can use this:

    ```vim
    Plug 'lervag/vimtex', {'branch': 'toggle-fractions'}
    ```

    After the change, you should do `PlugUpdate vimtex` and restart.

2. Go to your vimtex folder, e.g. `cd ~/.vim/bundle/vimtex`, then:

    ```sh
    git pull
    git checkout toggle-fractions

    # To revert, do
    git checkout master
    ```

nilsbecker:
thank -- i use Vundle right now which afaict has no git branch support -- i'll manually switch then

nilsbecker:
preliminary testing: 
`\frac{\alpha}{\mu} ` becomes `(\alpha)/(\mu)` which has superfluous parentheses. 
i'm not sure if this is problematic to fix but a fix would be good.

`\frac{r}{f}` becomes `r/f` which is fine; on the way back that becomes `\frac r f` which i think is valid latex but is not ideal -- if someone wants to extend the numerator or denominator it's probably good to have the curly braces? might come down to taste.

but the most important problem is that sometimes vim freezes for 5s with 100% cpu when i invoke tsd?!

nilsbecker:
`(r+t)/((\mu))` becomes `\frac{r+t}{(\mu})`. notice the swapped parentheses.

lervag:
Thanks for the new cases, I'll look into fixing them. I think it should be possible to cover all of them.

> \frac{r}{f} becomes r/f which is fine; on the way back that becomes \frac r f which i think is valid latex but is not ideal -- if someone wants to extend the numerator or denominator it's probably good to have the curly braces? might come down to taste.

Yes, this is probably a metter of taste. I'll change it back to curly braces,
since that might be more standard. And I agree, it makes it easier to extend.

lervag:
I've pushed an update that should fix the edge cases you've noticed.

nilsbecker:
i've noticed this:
`\frac{\delta_{02}}{\delta_{02} + \delta_{01}}`
becomes 
`(\delta_{02})/(\delta_{02} + \delta_{01})`
which is fine. but the way back works only when the cursor is over the denominator. i was surprised at first but maybe that is actually a feature? it might avoid some complexity in deciding how far the fraction extends?

nilsbecker:
what is the current heuristic to decide how far a denominator extends? if there are parentheses, everything inside them i guess? only `()` or also `[]` and `\{\}` ? if there are no parentheses, the current `W`ord or something fancier?

nilsbecker:
another one:
`\(a/p_\text{b}\)`. this gives
`\(\frac{a}{p_}\text{b}\)` when invoked over `a` or over `p` and does nothing over the subscript

lervag:
Thanks! I think the first issue should be possible to solve, I'll look into it. The latter issue is more difficult.

> what is the current heuristic to decide how far a denominator extends? if there are parentheses, everything inside them i guess? only () or also [] and \{\} ? if there are no parentheses, the current Word or something fancier?

If there are parantheses, I try to match the opening and closing parantheses. This should work well. The problem is in the case where there is no parantheses, such as the last example you provide. Here we need more complex parsing. In particular, we need a smart way to detect that `p_\text{b}` is a single "atom". Currently, I use a simple regex for word characters: `^\s*\\\?\w*`. The problem is that it does not match the super/sub script.

nilsbecker:
hi, i just stumbled upon `f(x+y)/g(z)` which will try to make a fraction out of the arguments of the functions. i think there cannot be a general rule that always works according to the intentions of the writer -- in this case it could be reasonable to treat 1. parentheses 2. stuff outside parentheses that is not separated by space as part of numerator/denominator. but in fact i think a visual mode version of the tsf command could be quite a bit easier: split all of the visual selection on `/` and just take these two parts as num and denom.

nilsbecker:
the heuristic above would mean `f(x)g(y)/h(z)` becomes `\frac{f(x)g(y)}{h(z)}` but that's fine imo.

lervag:
> in this case it could be reasonable to treat 1. parentheses 2. stuff outside parentheses that is not separated by space as part of numerator/denominator

Yes, I agree. This is a reasonable approach. I think it will be very difficult to cover every edge case, and so falling back to a relatively simple heuristic should suffice. We'll make sure to document it well.

> but in fact i think a visual mode version of the tsf command could be quite a bit easier: split all of the visual selection on / and just take these two parts as num and denom.

I agree, but I want that to come after the normal mode version is OK and the current PR is merged.

lervag:
Ok, I've updated again so all tests now pass. I think we are ready for merge, but let me know if you find other edge cases that should be updated. I'll also add some docs before merging.

lervag:
I'm merging this now, since I think it is already useful and it should not break anything. Feel free to open new issues if you find things that should work but doesnt.

Also, feel free to suggest a visual mode version of this; if you do, let me know how you think it should work. E.g. from `$a/b$`, is it enough to assume "split on `/`"?

nilsbecker:
i have used it a few times with success. i think it's already useful.

about visual mode, i think a version that is easy to understand and memorize is indeed 'split the entire selection on `/`'. then user is responsible for selecting something reasonable. 
on the way back, from `\frac{x}{y}`, i'm not sure a visual mode makes much sense?

lervag:
Thanks, I've added a new issue for it (see #1651). Let me know in that thread if you agree with the specification.

PR #1629, opened 2020-03-17
Recognize \subfileinclude{file} as file include

latricewilgus:
Add 'subfileinclude' to the recognized include indicators (option
g:vimtex_include_indicators). \subfileinclude is a variant of \subfile
and is provided by the subfiles [0] package (since version 1.5).

[0] https://www.ctan.org/pkg/subfiles


lervag:
Thanks!

Issue #1634, opened 2020-03-23
Startup speed improvements by disabling optional features?


j2kun:
Love vimtex! Thanks for building/maintaining it!

I have noticed that vimtex is quite slow on startup ~2 seconds. When profiling, I see that most of it comes from calls to `vimtex#parser#general#parse()` (perhaps due to calls to `gather_sources`?) which for my project results in some 23k calls to a seemingly expensive read function (section of the profiler output pasted below, entire log here: 
[vim.log](https://github.com/lervag/vimtex/files/4366669/vim.log))

I'm wondering if there is a way for me to turn off some optional features to reduce this load on startup time. Is it searching for and parsing a whole bunch of files I don't need for most work? (such as aux, out, and log files?)

```
FUNCTION  433()
    Defined: ~/.vim/bundle/vimtex/autoload/vimtex/parser/general.vim line 41
Called 33 times
Total time:   1.990302
 Self time:   1.977864

count  total (s)   self (s)
   33              0.000645   if !filereadable(a:file) || index(self.prev_parsed, a:file) >= 0
    1              0.000001     return []
   32              0.000025   endif
   32              0.000107   call add(self.prev_parsed, a:file)
                            
   32              0.000070   let l:parsed = []
   32              0.000050   let l:lnum = 0
23165              0.045789   for l:line in readfile(a:file)
23134              0.039972     let l:lnum += 1
                            
23134              0.031205     if self.finished
                                  break
23134              0.012942     endif
                            
23134              0.068357     if has_key(self, 're_stop') && l:line =~# self.re_stop
    1              0.000003       let self.finished = 1
    1              0.000001       break
23133              0.012778     endif
                            
23133              0.025432     if self.detailed
21946              0.083893       call add(l:parsed, [a:file, l:lnum, l:line])
 1187              0.000898     else
 1187              0.003237       call add(l:parsed, l:line)
23133              0.017361     endif
                            
23133              0.292964     if l:line =~# self.input_re
   27   0.006763   0.000405       let l:file = self.input_parser(l:line, a:file, self.input_re)
   27              0.000191       call extend(l:parsed, self.parse(l:file))
   27              0.000040       continue
23106              0.013738     endif
23138              0.021502   endfor
                            
   32              0.000055   return l:parsed
```

lervag:
> Love vimtex! Thanks for building/maintaining it!

Thanks! <3

[...]

PR #1643, opened 2020-03-28
Gather source files from fls

lervag:
Attempt to parse the fls file to gather list of main source files. See #1634 for details.

@clason @j2kun I would be happy if you could test this. I've tried to apply some heuristics to parse the fls file for a list of the main source files.

Please note that this may or may not work as expected! If you find it does not work, then I would be very happy to see explicit examples of that.


clason:
At first glance, seems to work OK. Timings for `time nvim BIG_PROJECT_SUBFILE.TEX +q` without `.fls`
```
Executed in  855.26 millis    fish           external
   usr time  628.30 millis  127.00 micros  628.17 millis
   sys time  139.07 millis  411.00 micros  138.66 millis
```
with `.fls`
```
Executed in  544.99 millis    fish           external
   usr time  342.12 millis  142.00 micros  341.98 millis
   sys time  135.14 millis  465.00 micros  134.68 millis
```
Anything in particular I should watch out for, or functionality I should test? (Completion seems to work project-wide.)

lervag:
> Anything in particular I should watch out for, or functionality I should test? (Completion seems to work project-wide.)

Yes: Check that the list of source files is correct (you find it with `:VimtexInfo` / `\li`).

clason:
It only shows the base file (correctly), not the included files -- is that correct?

lervag:
Nope, that seems wrong. If you remove the `.fls` file and open one of the LaTeX files, then the vimtex info screen should show a list of `source files:`. This list should include all of the `included` or `input`ed files. If you compile and have a `.fls` file, it should show the same list. If it doesn't, then something is wrong.

As I mentioned in #1634, I'm not sure if it is easy to parse the `.fls` file correctly. I don't know if it is fully structured. I tested and observed a couple of things that I thought could be a pattern (i.e., all INPUT tex files between the OUTPUT aux and OUTPUT pdf are included source files. This looked correct on a couple of examples I made.

A different option could be to simply collect all INPUT tex files, but ignore files that match some common patterns (e.g. `/usr/share/texmf` or perhaps any kind of `texmf` path).

clason:
Oh, I see. the `source files` part of the Info is missing completely. Most actual functionality (I've tested) still works, though?

> (i.e., all INPUT tex files between the OUTPUT aux and OUTPUT pdf are included source files.

No, that's definitely not correct (in my example). Maybe it depends on the number of (pdf)latex runs?

My naive thought would have been to filter all `INPUT` lines, then filter all lines _not_ starting with an absolute path, and then filter on a whitelist (tex/tikz/bib etc.)?

j2kun:
Having completely missed what appears to have been an extremely productive conversation, I ran the new code against my existing repository (~100 tex files, 22k lines, with `.fls`) and startup time is down from 2s to 750-850 ms, a 60% improvement!

😍😍😍😍

For reference, the FLS file has 608 `INPUT` lines, 396 unique.

Without the fls, it's down to 1.2s, a 40% improvement.

Well. Effing. Done. Got a donations box?

lervag:
@j2kun Happy to hear the improvements are appreciated! Feel free to donate (there's a button on the README/github front page), but also know that I do this stuff because I enjoy it and I have a well payed job. I would be equally happy if you would donate some money to a charity/people in need.

For now, it seems the `.fls` parsing is not quite good  yet. I would be happy to get some good examples of projects where the parsing is off  so that I could investigate and improve. It would be good help to get examples where things are wrong.

clason:
@lervag You could try https://arxiv.org/e-print/2001.00216 (download the source, add `.tar.gz` extension, unpack). (Ignore the biblatex warning, it will still generate a fls file that shows the same behavior as mentioned above.)

clason:
That seems to work for my project 👍 (although it also picks up the asy-generated tex files, which the other parser doesn't?)

lervag:
Thanks. The example is good because it also highlights that the `fls` file may be incomplete.

I've pushed an update where I instead filter out all `.tex` files in `texmf` like paths. I don't think this is OK, because you could be working on files within a texmf tree and things would stop working.

I'm thinking that perhaps this route is not so good, because 1) incomplete `fls` files could lead to unexpected behaviour, and 2) it is difficult to reliable filter out source files that are not part of the current project.

lervag:
> That seems to work for my project +1 (although it also picks up the asy-generated tex files, which the other parser doesn't?)

Wow, you're fast! :)

Could you check again what the speed benefit is? If it is substantial, then I might reconsider. But if so, it is important to find a good way to reliable exclude (most) files that are not really relevant.

clason:
The timings are pretty much the same as before (slightly higher individually, but that is consistent between with and without fls, so not due to the parsing).

clason:
Overall I would say that after your previous changes, the benefit of parsing the fls instead is too small to merit sinking work into this. (Especially if the fls can't be trusted.)

clason:
I should add that I'm starting with  `nvim $subfile.tex`, not the project main file (under the assumption that this would be the harder challenge).

lervag:
> The timings are pretty much the same as before (slightly higher individually, but that is consistent between with and without fls, so not due to the parsing).

Ok!

> I should add that I'm starting with nvim $subfile.tex, not the project main file (under the assumption that this would be the harder challenge).

Should not really be much more of a challenge, at least not wrt. parsing the `.fls` file. Perhaps a slightly worse startup time.

> Overall I would say that after your previous changes, the benefit of parsing the fls instead is too small to merit sinking work into this. (Especially if the fls can't be trusted.)

The `.fls` file can only be trusted if compilation is succesfully completed. If there are errors, then it seems the `.fls` file can be incomplete. In the example from arxiv there was an error with `biblatex` that interrupted compilation, and so the `.fls` file did not contain INPUT lines for all source files.

So, I think I'll close this PR. In case I'll pick this up again in the future, here's the current patch:

```diff
diff --git a/autoload/vimtex/parser/fls.vim b/autoload/vimtex/parser/fls.vim
index 83a39c3..fbe9ea3 100644
--- a/autoload/vimtex/parser/fls.vim
+++ b/autoload/vimtex/parser/fls.vim
@@ -13,3 +13,29 @@ function! vimtex#parser#fls#parse(file) abort " {{{1
 endfunction
 
 " }}}1
+function! vimtex#parser#fls#parse_files(vimtex) abort " {{{1
+  " Argument: vimtex is a vimtex blob because we need several properties
+
+  if !filereadable(a:vimtex.fls())
+    return []
+  endif
+
+  " This seems like a correct heuristic to limit the scope for parsing the
+  " included main source files
+  let l:files = readfile(a:vimtex.fls())
+
+  " Only include the INPUT lines with .tex extension
+  call filter(l:files, 'stridx(v:val, ''INPUT'') == 0')
+  call filter(l:files, 'v:val =~# ''\.tex$''')
+
+  " Ignore package files under texmf trees
+  call filter(l:files, 'v:val !~# ''\/texmf[^\/]*\/''')
+
+  " Remove INPUT prefix
+  call map(l:files, 'strpart(v:val, 6)')
+
+  " Add the main file to the list before returning
+  return vimtex#util#uniq_unsorted(l:files)
+endfunction
+
+" }}}1
diff --git a/autoload/vimtex/state.vim b/autoload/vimtex/state.vim
index 9160250..bee726d 100644
--- a/autoload/vimtex/state.vim
+++ b/autoload/vimtex/state.vim
@@ -642,8 +642,12 @@ endfunction
 
 " }}}1
 function! s:vimtex.gather_sources() abort dict " {{{1
-  let self.sources = vimtex#parser#tex#parse_files(
-        \ self.tex, {'root' : self.root})
+  let self.sources = vimtex#parser#fls#parse_files(self)
+
+  if empty(self.sources)
+    let self.sources = vimtex#parser#tex#parse_files(
+          \ self.tex, {'root' : self.root})
+  endif
 
   call map(self.sources, 'vimtex#paths#relative(v:val, self.root)')
 endfunction
diff --git a/test/tests/test-metadata/test-sources-fls.vim b/test/tests/test-metadata/test-sources-fls.vim
new file mode 100644
index 0000000..9ead884
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls.vim
@@ -0,0 +1,17 @@
+set nocompatible
+let &rtp = '../../..,' . &rtp
+filetype plugin on
+
+let g:tex_flavor = 'latex'
+
+silent edit test-sources-fls/main.tex
+
+call vimtex#test#assert_equal([
+  \ 'main.tex',
+  \ 'include1.tex',
+  \ 'sub1/include2.tex',
+  \ 'sub2/include3.tex',
+  \ 'subfile.tex',
+  \], b:vimtex.sources)
+
+quit!
diff --git a/test/tests/test-metadata/test-sources-fls/include1.tex b/test/tests/test-metadata/test-sources-fls/include1.tex
new file mode 100644
index 0000000..376611f
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/include1.tex
@@ -0,0 +1,3 @@
+Hello world from include1.
+
+\subimport*{sub1/}{include2}
diff --git a/test/tests/test-metadata/test-sources-fls/main.fls b/test/tests/test-metadata/test-sources-fls/main.fls
new file mode 100644
index 0000000..3452afb
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/main.fls
@@ -0,0 +1,26 @@
+PWD /home/lervag/.vim/bundle/vimtex/test/tests/test-metadata/test2
+INPUT /usr/share/texmf-dist/web2c/texmf.cnf
+INPUT /var/lib/texmf/web2c/pdftex/pdflatex.fmt
+INPUT main.tex
+OUTPUT main.log
+INPUT /usr/share/texmf-dist/tex/latex/base/minimal.cls
+INPUT /usr/share/texmf-dist/tex/latex/base/minimal.cls
+INPUT /usr/share/texmf-dist/tex/latex/import/import.sty
+INPUT /usr/share/texmf-dist/tex/latex/import/import.sty
+INPUT /usr/share/texmf-dist/tex/latex/subfiles/subfiles.sty
+INPUT /usr/share/texmf-dist/tex/latex/subfiles/subfiles.sty
+INPUT main.aux
+INPUT main.aux
+OUTPUT main.aux
+INPUT include1.tex
+INPUT include1.tex
+INPUT ./sub1/include2.tex
+INPUT ./sub1/include2.tex
+INPUT sub2/include3.tex
+INPUT sub2/include3.tex
+INPUT subfile.tex
+INPUT subfile.tex
+OUTPUT main.pdf
+INPUT /var/lib/texmf/fonts/map/pdftex/updmap/pdftex.map
+INPUT main.aux
+INPUT /usr/share/texmf-dist/fonts/type1/public/amsfonts/cm/cmr10.pfb
diff --git a/test/tests/test-metadata/test-sources-fls/main.tex b/test/tests/test-metadata/test-sources-fls/main.tex
new file mode 100644
index 0000000..a91afc6
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/main.tex
@@ -0,0 +1,11 @@
+\documentclass{minimal}
+\usepackage{import}
+\usepackage{subfiles}
+
+\begin{document}
+
+\import{./}{include1}
+\inputfrom{sub2/}{include3}
+\subfile{subfile.tex}
+
+\end{document}
diff --git a/test/tests/test-metadata/test-sources-fls/sub1/include2.tex b/test/tests/test-metadata/test-sources-fls/sub1/include2.tex
new file mode 100644
index 0000000..360be04
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/sub1/include2.tex
@@ -0,0 +1 @@
+Hello world from include2.
diff --git a/test/tests/test-metadata/test-sources-fls/sub2/include3.tex b/test/tests/test-metadata/test-sources-fls/sub2/include3.tex
new file mode 100644
index 0000000..36ffa3a
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/sub2/include3.tex
@@ -0,0 +1,3 @@
+Hello world from include3.
+
+Note: This file tests that we also search from the main root.
diff --git a/test/tests/test-metadata/test-sources-fls/subfile.tex b/test/tests/test-metadata/test-sources-fls/subfile.tex
new file mode 100644
index 0000000..8856625
--- /dev/null
+++ b/test/tests/test-metadata/test-sources-fls/subfile.tex
@@ -0,0 +1,3 @@
+\documentclass[main.tex]{subfiles}
+\begin{document}
+\end{document}
```

PR #1655, opened 2020-04-03
Use non-greedy match for lstinline/lstinputlisting

AymenQ:
Use a non-greedy match (`.\{-}`) instead of `.*` for matching the contents of square brackets after `\lstinline` or `\lstinputlisting`.

This PR fixes an issue where:

`\lstinline[language=Bash]{cd ..} some text and another \lstinline[language=Bash]{ls}`

would be all matched and highlighted as an entire texZone/texInputFile instead of two separate regions.


lervag:
Thanks!

PR #1664, opened 2020-04-15
Allow customizing FZF parameters

JoosepAlviste:
Hey!

I had defined a custom default FZF layout for myself using `let g:fzf_layout = { ... }`, but this didn't customize the Vimtex FZF window. So, I figured that it would be nice to allow the FZF window to be customized.

I suppose that another approach would be to use the `g:fzf_layout` variable automatically if it has been defined. But that might have some unexpected consequences for some people so I implemented a more explicit configuration method.

I didn't add/change any tests since I couldn't find anything related to the FZF integration.


lervag:
Thanks, this looks good. I've merged it now.

PR #1666, opened 2020-04-17
Fix bug where stopping and starting compile causes multiple hooks registered

camspiers:
`g:vimtex_compiler_callback_hooks` gets multiple entries of `b:vimtex.viewer.compiler_callback` in it. This stops that and ensures just one gets added.

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -43,8 +43,11 @@ function! vimtex#view#init_state(state) abort " {{{1
   " Add compiler callback to callback hooks (if it exists)
   "
   if exists('*l:v.compiler_callback')
-    call add(g:vimtex_compiler_callback_hooks,
-          \ 'b:vimtex.viewer.compiler_callback')
+    if index(g:vimtex_compiler_callback_hooks,
+          \ 'b:vimtex.viewer.compiler_callback') == -1
+      call add(g:vimtex_compiler_callback_hooks,
+            \ 'b:vimtex.viewer.compiler_callback')
+    endif
   endif
 
   "


camspiers:
I don't know if this is the right way to address it, but it at least fixes the underlying bug that I am running into. For clarity, when I run VimtexCompile, and then VimtexCompileStop, and then VimtexCompile again, `g:vimtex_compiler_callback_hooks` ends up with multiple entries of `'b:vimtex.viewer.compiler_callback'` in it, and as such when you save a document, the callback gets run more times than it is meant to.

camspiers:
If you want to see a repro:

```
let g:vimtex_view_general_callback = 'TermPDF'
let g:vimtex_view_automatic = 0

function! TermPDF(status) abort
  echom "TermPDF rendered:" . b:vimtex.root . '/' . b:vimtex.name . '.pdf'
endfunction
```

Then run VimtexCompile and VimtexCompileStop a couple of times. Run VimtexCompile a final time, then make a change to the text document and save it.

If you then run `:messages` you will see there are many `TermPDF rendered` calls after each save.

lervag:
Thanks! My first thought is that you are not fixing the root cause of the issue, and that it should be resolved in some other way, but I'll look into it.

lervag:
I can't reproduce your issue directly. That is, for me, there is only a single hook independent of how many times I run `VimtexCompile`. However, if I do `:VimtexReload`, then the hook is added several times, and in this case it seems your proposed change may be appropriate.

lervag:
I've pushed your commit, slightly altered; I combined the conditionals into a single one. Thanks!

I'm curious, though: Could it be that your real issue was related to `:VimtexReload` (mapped to `\lx`) instead of `:VimtexCompile` (`\ll`)?

Issue #1702, opened 2020-05-22
Table of contents do not include sections from recursively included subfiles


thor:
_(A short background introduction: the support for `subfiles` and synctex is the combinational killer feature for me personally with vimtex. A big thank-you to you and all the other contributors.)_

Looking at #155 and [maxfl's comment](https://github.com/lervag/vimtex/issues/155#issuecomment-98073693) I've used subfiles with an absolute path for each subfile pointing to `main.tex` based on the working directory.

Now I'm trying to add another level of subfiles, specifically a structure like this:

```
─ $(pwd)
  ├── main.tex
  └── parts/
      ├── chapter.tex
      └── sections/
          ├── first.tex
          └── second.tex
```

While compilation proceeds without any issues whether I compile `main.tex` or any one of `chapter.tex`, `first.tex` or `second.tex` by switching the current compilation target with `<localleader>ls`, I can see that the vimtex table of contents does not include the sections from `parts/sections/*.tex`.I presume this is because it tries to read the contents of `$(pwd)/sections/first.tex`, as opposed to `$(pwd)/parts/sections/first.tex`.

### Mucking about

*If* I change the inclusion in `chapter.tex` to `parts/sections/*.tex` it will include the files in the list of detected source files. However, compilation will fail.

The closest I've come is by changing the parameter to the `subfiles` document class for `chapter.tex`, but this seems like inviting trouble as it suddenly inconsistent with absolute or relative paths.

```latex
\documentclass[../main.tex]{subfiles}
% ...
```
Upside:
* If I open `parts/chapter.tex` from `$(pwd)` the table of contents include the sections.

Downside:
* If I open `main.tex` or any of the `sections/*.tex` the table of contents _do not_ include the sections.

_Another_ way to kind of muck about that sort of works around it, if you _only_ compile via vimtex: change `$(pwd)` to `parts/`, then open `chapter.tex`. vimtex correctly finds the `main.tex` above, despite `$(pwd)` being incorrect, and compilation works from within nvim. I note that the path for the `parts/sections/*.tex` files end up with a full path in the table of contents.

**Expected Behaviour**

The table of contents includes all the sections included by subfiles that are included by subfiles (that are included by ...).

Perhaps while using absolute paths for `\documentclass[abs/path/file.tex]{subfiles}` the files included by the `subfiles` package should be relative to the current file the `subfiles` are used in?

[...]

Issue #155, opened 2015-04-16
Multi-file project with subfile


maxfl:
Hi, I'm used to do multi-file projects, based on subfile package. In other words each included file has its own begin/end document environment and can be compiled independently. The compilation and the compiled files happen in the root directory, even if the file itself is located in subfolder. There fore I have several questions.

1) Is there a simple way to tell vimtex where the main file is? It would be good to have a possibility to set it from local vimrc. Or may vimtex detect \documentclass[main.tex]{subfiles} line to determine the main file?
2) Is it possible to change the compilation root for the current file to the project root and compile the current file even when main file is set?
3) Is it possible to view the current file's pdf even when the main file set?

Thank you in advance. Vimtex is amazing and fast.

lervag:
Hi! First, I've never heard about the subfile package. It seems like a nice idea, although it does introduce some complexities that I am not quite sure if I want to introduce to `vimtex`.

Regarding your first question: Please see `:h vimtex-multi-file`, which explains how you can use a comment at the top of the file similar to
```tex
%! TEX root = my-main.tex
```
in order to set the main file. This should always work.

For your second question, the answer is currently no. The question is interesting though, as it could be a nice functionality to compile the current file outside of context of the main file. However, I don't see a simple way to do , and I am not quite sure that I want to complicate the code in order to support this. Much because there does exist a "standard" way to solve the problem through the `\includes` in a main file.

The same applies to your third question: `vimtex` works under the assumption of a single project, and does not support the idea to let a file (or set of files) be considered separate from the main file. Thus only the main file's pdf is known to `vimtex`.

I would be willing to discuss ideas for new features and enhancements, but I want to stress that I one of the main benefits of `vimtex` IMHO is that it is a clean and relatively uncomplicated code. This particular issue does seem a little bit complicated. However, if we can find small and elegant solutions that help improve support for your needs, then I will definately want to implement them.

maxfl:
I should have written that I know about '%! TEX root'. The reason I'm asking about the other way is that I have around 40 files in a project and adding this line to every file seems inelegant. Before when using LatexBox I was using autocommand to assign the necessary variable. Moreover, every file has already a '\documentclass[my-main.tex]{subfiles}' which can be reused for this purpose. Since vimtex needs a master file for the correct completions and already parses the file for the string '%!..' I think it would be quite essential to determine the main file by looking for documentclass as well. 

[...]

PR #1703, opened 2020-05-24
Improve subfiles handling (addresses #1702)

lervag:
Currently a work in progress, requires ~~more robust code~~ and persistent tests. But it seems to fix the issue raised in #1702.

### DIFF
--- a/autoload/vimtex/parser/tex.vim
+++ b/autoload/vimtex/parser/tex.vim
@@ -153,7 +153,7 @@ function! s:input_parser(line, current_file, root) abort " {{{1
   " Handle \space commands
   let l:file = substitute(a:line, '\\space\s*', ' ', 'g')
 
-  " Handle import package commands
+  " Handle import and subfile package commands
   if l:file =~# g:vimtex#re#tex_input_import
     let l:root = l:file =~# '\\sub'
           \ ? fnamemodify(a:current_file, ':p:h')
--- a/autoload/vimtex/parser/toc.vim
+++ b/autoload/vimtex/parser/toc.vim
@@ -171,10 +171,14 @@ let s:matcher_include = {
 function! s:matcher_include.get_entry(context) abort dict " {{{1
   let l:file = matchstr(a:context.line, self.re)
   if !vimtex#paths#is_abs(l:file[0])
-    let l:file = b:vimtex.root . '/' . l:file
+    " Handle import and subfile package commands
+    let l:root = a:context.line =~# '\\sub'
+          \ ? fnamemodify(a:context.file, ':p:h')
+          \ : b:vimtex.root
+    let l:file = l:root . '/' . l:file
   endif
   let l:file = fnamemodify(l:file, ':~:.')
-  if !filereadable(l:file)
+  if !filereadable(l:file) && filereadable(l:file . '.tex')
     let l:file .= '.tex'
   endif
   return {
--- a/autoload/vimtex/re.vim
+++ b/autoload/vimtex/re.vim
@@ -10,11 +10,13 @@ let g:vimtex#re#not_comment = '\v%(' . g:vimtex#re#not_bslash . '\%.*)@<!'
 let g:vimtex#re#tex_input_root =
       \ '\v^\s*\%\s*!?\s*[tT][eE][xX]\s+[rR][oO][oO][tT]\s*\=\s*\zs.*\ze\s*$'
 let g:vimtex#re#tex_input_latex = '\v\\%('
-      \ . join(get(g:, 'vimtex_include_indicators',
-      \            ['input', 'include', 'subfile', 'subfileinclude']),
-      \        '|') . ')\s*\{'
-let g:vimtex#re#tex_input_import =
-      \ '\v\\%(sub)?%(import|%(input|include)from)\*?\{[^\}]*\}\{'
+      \ . join(get(g:, 'vimtex_include_indicators', ['input', 'include']), '|')
+      \ . ')\s*\{'
+let g:vimtex#re#tex_input_import = '\v\\%('
+      \ . 'subfile%(include)?'
+      \ . '|'
+      \ . '%(sub)?%(import|%(input|include)from)\*?\{[^\}]*\}'
+      \ . ')\s*\{'
 let g:vimtex#re#tex_input_package =
       \ '\v\\%(usepackage|RequirePackage)%(\s*\[[^]]*\])?\s*\{\zs[^}]*\ze\}'
 
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -558,17 +558,22 @@ Subfiles package~
   the `subfiles` package, the |VimtexToggleMain| command is particularly
   useful.
 
-  With `subfiles`, included files will look like this: >
+  With `subfiles`, included files will typically look like this: >
 
-    \documentclass[main.tex]{subfiles}
+    \documentclass[<main-path>]{subfiles}
     \begin{document}
     ...
     \end{document}
 <
-  Absolute paths for to the main file (e.g. `main.tex`) will work as expected,
-  but when using a relative path, one should note that vimtex assumes we start
-  compilation from the directory of the main file. Thus one should not write
-  e.g. `../main.tex`.
+  Here `<main-path>` is the path to the main file. It must be specified as
+  relative to the particular subfile. So, given the structure: >
+
+    main.tex
+    sub/sub.tex
+<
+  The header in `sub.tex` should be `\documentclass[../main.tex]{subfiles}`.
+  Absolute paths like `/home/user/main.tex` are also allowed and should work
+  as expected.
 
   [0]: https://www.ctan.org/pkg/import
   [1]: https://www.ctan.org/pkg/subfiles
@@ -1663,12 +1668,13 @@ OPTIONS                                                        *vimtex-options*
 *g:vimtex_include_indicators*
   Vimtex will recognize included files for a lot of different purposes. Most
   of these come from e.g. `\input{file}` or `\include{file}`. This option
-  allows to add more commands that are used to include files.
+  allows to add more commands that are used to include files, e.g. custom
+  macros.
 
   Note: This option is read during initialization of vimtex, and so it must be
         set early. I.e., it can not be set in `after/ftplugin/tex.vim`.
 
-  Default value: ['input', 'include', 'subfile', 'subfileinclude']
+  Default value: ['input', 'include']
 
 *g:vimtex_include_search_enabled*
   Vimtex sets 'includeexpr' to recognize included files. If a file isn't found
--- a/test/tests/test-subfiles/Makefile
+++ b/test/tests/test-subfiles/Makefile
@@ -1,0 +1,14 @@
+MYVIM ?= nvim --headless
+
+INMAKE := 1
+export INMAKE
+
+TESTS := $(wildcard test*.vim)
+TESTS := $(TESTS:.vim=)
+
+.PHONY: default $(TESTS)
+
+default: $(TESTS)
+
+$(TESTS):
+	@$(MYVIM) -u $@.vim
--- a/test/tests/test-subfiles/test_nested.vim
+++ b/test/tests/test-subfiles/test_nested.vim
@@ -1,0 +1,19 @@
+set nocompatible
+let &rtp = '../../..,' . &rtp
+filetype plugin on
+
+set nomore
+
+nnoremap q :qall!<cr>
+
+silent edit test_nested/parts/chapter.tex
+call vimtex#test#assert_equal(expand('%:p:h:h'), b:vimtex.root)
+silent VimtexToggleMain
+call vimtex#test#assert_equal(expand('%:p:h'), b:vimtex.root)
+
+silent edit test_nested/parts/sections/first.tex
+call vimtex#test#assert_equal(expand('%:p:h:h:h'), b:vimtex.root)
+silent VimtexToggleMain
+call vimtex#test#assert_equal(expand('%:p:h'), b:vimtex.root)
+
+quitall!
--- a/test/tests/test-subfiles/test_nested/main.tex
+++ b/test/tests/test-subfiles/test_nested/main.tex
@@ -1,0 +1,9 @@
+\documentclass[a4paper]{report}
+\usepackage{subfiles}
+
+\begin{document}
+\tableofcontents
+
+\subfile{parts/chapter.tex}
+
+\end{document}
--- a/test/tests/test-subfiles/test_nested/parts/chapter.tex
+++ b/test/tests/test-subfiles/test_nested/parts/chapter.tex
@@ -1,0 +1,7 @@
+\documentclass[../main.tex]{subfiles}
+\begin{document}
+\chapter{Chapter with subsections}
+Here we introduce the bug as it currently is experienced by a user.
+\subfile{sections/first.tex}
+\subfile{sections/second.tex}
+\end{document}
--- a/test/tests/test-subfiles/test_nested/parts/sections/first.tex
+++ b/test/tests/test-subfiles/test_nested/parts/sections/first.tex
@@ -1,0 +1,10 @@
+\documentclass[../../main.tex]{subfiles}
+\begin{document}
+
+\section{The first section}
+Titling notwithstanding, could this issue be a case of PEBCAC?
+
+\subsection{A subsection}
+Hello world.
+
+\end{document}
--- a/test/tests/test-subfiles/test_nested/parts/sections/second.tex
+++ b/test/tests/test-subfiles/test_nested/parts/sections/second.tex
@@ -1,0 +1,5 @@
+\documentclass[../../main.tex]{subfiles}
+\begin{document}
+\section{The second section}
+You don't always think it be like this, but it do.
+\end{document}
--- a/test/tests/test-toc/chapters/sections/first.tex
+++ b/test/tests/test-toc/chapters/sections/first.tex
@@ -1,0 +1,10 @@
+\documentclass[../../main.tex]{subfiles}
+\begin{document}
+
+\section{The first section}
+Titling notwithstanding, could this issue be a case of PEBCAC?
+
+\subsection{A subsection}
+Hello world.
+
+\end{document}
--- a/test/tests/test-toc/chapters/sections/second.tex
+++ b/test/tests/test-toc/chapters/sections/second.tex
@@ -1,0 +1,5 @@
+\documentclass[../../main.tex]{subfiles}
+\begin{document}
+\section{The second section}
+You don't always think it be like this, but it do.
+\end{document}
--- a/test/tests/test-toc/chapters/subfile.tex
+++ b/test/tests/test-toc/chapters/subfile.tex
@@ -1,7 +1,10 @@
-\documentclass[main.tex]{subfiles}
+\documentclass[../main.tex]{subfiles}
 
 \begin{document}
 \chapter{Subfile chapter}
 Lorem ipsum...
+
+\subfile{sections/first.tex}
+\subfile{sections/second.tex}
 
 \end{document}
--- a/test/tests/test-toc/test.vim
+++ b/test/tests/test-toc/test.vim
@@ -2,6 +2,8 @@ set nocompatible
 let &rtp = '../../..,' . &rtp
 filetype plugin on
 
+set nomore
+
 nnoremap q :qall!<cr>
 
 let g:vimtex_toc_custom_matchers = [
@@ -14,6 +16,15 @@ silent edit main.tex
 if empty($INMAKE) | finish | endif
 
 let s:toc = vimtex#toc#get_entries()
-call vimtex#test#assert_equal(len(s:toc), 8)
+call vimtex#test#assert_equal(len(s:toc), 13)
+
+" let s:i = 0
+" for s:x in s:toc
+"   echo s:i '--' s:x.title "\n"
+"   let s:i += 1
+" endfor
+
+call vimtex#test#assert_equal('chapters/sections/first.tex', s:toc[5].file)
+call vimtex#test#assert_equal('chapters/sections/second.tex', s:toc[8].file)
 
 quit!


thor:
I've tested the changes with regards to the table of contents alone, and it seems to work very well! :+1: If I understand correctly, you've added/altered the `\subfile` variants to be treated in the same way as the `\import`, and then get their location and use that to parse the files relative to the file they're included from(?).

I'm not sure what the other issue at hand is... I think. 
I thought I was able to compile the `parts/chapter.tex` in standalone mode, but I can't seem to, not via vimtex or with `latexmk` directly from within the root folder. I'm a bit confused where my thoughts were or what I did in addition, so I'll have to fiddle around with that again. I thought I could, but I can't, thus I must've misunderstood something essential.

lervag:
> I've tested the changes with regards to the table of contents alone, and it seems to work very well!

Great, happy to hear it!

> If I understand correctly, you've added/altered the \subfile variants to be treated in the same way as the \import, and then get their location and use that to parse the files relative to the file they're included from(?).

Yes, exactly. There are two main changes: One for the TeX parser to allow `\subfile{path}` to be relative to the current file, and a second to do the same for the table-of-contents parser.

> I thought I was able to compile the parts/chapter.tex in standalone mode, but I can't seem to, not via vimtex or with latexmk directly from within the root folder. I'm a bit confused where my thoughts were or what I did in addition, so I'll have to fiddle around with that again. I thought I could, but I can't, thus I must've misunderstood something essential.

Ah, yes. I never considered you might have made a wrong basic assumption. I think the problem is that we need a consistent root path. When we run `latexmk` from the `main.tex` directory, then `sections/first.tex` is not properly rooted. If we were to run `latexmk` from the `chapter.tex` directory, then the `main.tex` in `\documentclass[main.tex]{subfiles}` is not properly rooted. I think the conclusion is that all the `\subfile` commands should have either absolute paths or paths relative to the `main.tex` file.

Note that the `import` package works differently here, in that it "changes the CWD" to the imported package. This is quite well explained in the [short, two page manual](http://texdoc.net/texmf-dist/doc/latex/import/import.pdf).

I've looked into the [`subfiles` manual](http://texdoc.net/texmf-dist/doc/latex/subfiles/subfiles.pdf) as well, which seems to claim that this exact workflow _should_ work. However, it seems my above paragraph is still correct. That is, you would have to use `\documentclass[../main.tex]{subfiles}` in `chapter.tex` to make things compile correctly. So, I think an appropriate change is something like this:

1. Update the docs to explain this behaviour.

    ```
    It is important to note that the main-file path in
    `\documentclass[path]{subfiles}` implicitly indicates which root directory
    should be used for compilation. This is important, because the root
    directory in turn specifies the parent path for nested imports, includes,
    subfiles, includegraphics, and similar. See Section 3.1 in the manual of
    subfiles [1] for an example of a project with nested root directories.

    Vimtex will automatically try to parse the root path from the
    `\documentclass` macro to be consistent with this behaviour of the subfiles
    package.
    ```

2. Update the code to properly specify the root path for the subfile mode.

What do you think?

thor:
>  I think the conclusion is that all the `\subfile` commands should have either absolute paths or paths relative to the main.tex file.

To clarify that I do not misunderstand: that's absolute paths being relative paths seen from `main.tex` vs. relative paths being the path to `main.tex` as seen from each individual `*.tex` file.

---

Provided I have not miserably failed at my working example --- I'm not sure it is possible to only use absolute paths for multi-level hierarchies, although I would much prefer it.

In an attempt to create a setup where I can execute `latexmk` from `main.tex` and the target can be any of the documents, I am not able to get more than one of the two: I can't find a way where I can compile `parts/chapter.tex` and `main.tex` using `latexmk` from the root directory.

If I use absolute paths then 



> > I thought I was able to compile the parts/chapter.tex in standalone mode, but I can't seem to, not via vimtex or with latexmk directly from within the root folder. I'm a bit confused where my thoughts were or what I did in addition, so I'll have to fiddle around with that again. I thought I could, but I can't, thus I must've misunderstood something essential.
> 
> Ah, yes. I never considered you might have made a wrong basic assumption. I think the problem is that we need a consistent root path. When we run `latexmk` from the `main.tex` directory, then `sections/first.tex` is not properly rooted. If we were to run `latexmk` from the `chapter.tex` directory, then the `main.tex` in `\documentclass[main.tex]{subfiles}` is not properly rooted. I think the conclusion is that all the `\subfile` commands should have either absolute paths or paths relative to the `main.tex` file.

Right, indeed. If the current working directory remains the root folder, then one would assume all paths should be relative to the current working directory. However, I'm not sure that is the case.

> Note that the `import` package works differently here, in that it "changes the CWD" to the imported package. This is quite well explained in the [short, two page manual](http://texdoc.net/texmf-dist/doc/latex/import/import.pdf).
> 
> I've looked into the [`subfiles` manual](http://texdoc.net/texmf-dist/doc/latex/subfiles/subfiles.pdf) as well, which seems to claim that this exact workflow _should_ work. However, it seems my above paragraph is still correct. That is, you would have to use `\documentclass[../main.tex]{subfiles}` in `chapter.tex` to make things compile correctly. So, I think an appropriate change is something like this:

From the `subfiles` documentation it is mentioned that `\subfile{}` itself uses `import`. I interpret that to mean that any use of `\subfile{}` will be rooted in the directory of the file including it. Assuming that is the case the only documents that can use a path relative to `main.tex` are those in the same folder as `main.tex`, whereas all others must use a path relative to their own location.

A somewhat messy execution table highlighting how this works for the example files in #1702:

| CWD      | Root `parts/chapter.tex` | Subfiles `parts/chapter.tex` | `latexmk chapter.tex>` | `latexmk main.tex` |
|----------|--------------------------|------------------------------|:----------------------:|:------------------:|
| `/`      | `main.tex`               | `sections/*.tex`             | ❌                     | ✅                 |
| `/`      | `main.tex`               | `parts/sections/*.tex`       | ✅                     | ❌                 |
| `/`      | `../main.tex`            | `sections/*.tex`             | ❌                     | ✅                 |
| `/`      | `../main.tex`            | `parts/sections/*.tex`       | ✅                       | ❌                 |
| `/parts` | `main.tex`               | `sections/*.tex`             | ❌                     | ❌                 |
| `/parts` | `main.tex`               | `parts/sections/*.tex`       | ❌                     | ❌                 |
| `/parts` | `../main.tex`            | `sections/*.tex`             | ✅                     | ❌                 |
| `/parts` | `../main.tex`            | `parts/sections/*.tex`       | ❌                     | ❌                 |

To me this seems to suggest that the only approach is to execute `latexmk` in the folder of the current file (if `main parser` is *local file* and not `subfiles`, as a result of switching target). However, I feel as though I'm missing something. I thought it would be possible to use absolute paths as you mentioned, but since `subfiles` uses `import` and case no. 2 only works for `chapter.tex`, it seems like that cannot be the case? :shrug: 

> 1. Update the docs to explain this behaviour.
>    ```
>    It is important to note that the main-file path in
>    `\documentclass[path]{subfiles}` implicitly indicates which root directory
>    should be used for compilation. This is important, because the root
>    directory in turn specifies the parent path for nested imports, includes,
>    subfiles, includegraphics, and similar. See Section 3.1 in the manual of
>    subfiles [1] for an example of a project with nested root directories.
>    
>    Vimtex will automatically try to parse the root path from the
>    `\documentclass` macro to be consistent with this behaviour of the subfiles
>    package.
>    ```
> 2. Update the code to properly specify the root path for the subfile mode.
> 
> What do you think?

Per above, making the optimistic and simultaneously pessimistic assumption that I haven't messed up in the testing, I think *"This is important, because the root directory in turn specifies the parent path for nested imports, includes, subfiles, includegraphics, and similar."* might not be correct --- at least for `\import{}` and `\subfiles{}`, perhaps not for `\includegraphics{}` either per `subfile2.tex` in the `subfiles` documentation sec. 3.1.

lervag:
Thanks for the thorough post. I've tried to respond in depth. I hope you might find time to test my below table of commands and verify (or falsify) my expectations.

> To clarify that I do not misunderstand: that's absolute paths being relative paths seen from `main.tex` vs. relative paths being the path to `main.tex` as seen from each individual `*.tex` file.

No, with absolute path I mean paths like `/home/user/path/main.tex`, i.e. paths that begin with `/`. Relative paths are paths that are not absolute, e.g. `../main.tex`. This latter form requires a well defined starting point.

----

So, to be pragmatic: we need to figure out how to handle things within vimtex. A couple of thoughts related to your comments first.

> Right, indeed. If the current working directory remains the root folder, then one would assume all paths should be relative to the current working directory. However, I'm not sure that is the case.

Yes, this was what I tried to say in my previous comment. Still, ...

> From the `subfiles` documentation it is mentioned that `\subfile{}` itself uses `import`. I interpret that to mean that any use of `\subfile{}` will be rooted in the directory of the file including it. Assuming that is the case the only documents that can use a path relative to `main.tex` are those in the same folder as `main.tex`, whereas all others must use a path relative to their own location.

I agree, from the `subfiles` docs it would seem like one should have a similar behaviour. However, as you show quite clearly, this is not the case. `subfiles` seems to be strongly connected to the original working directory. Some remarks:

* In your tests, with `CWD` set to `/`, did you use `latexmk parts/chapter.tex`? And similar, with `CWD` set to `/parts`, did you run `latexmk ../main.tex`? In general, I think the latter variant would not work.

* Based on my previous comment, I would propose the following table of tests:

    | CWD      | Root          | Subfiles               | Command 1             | Command 2                   |
    | ---      | ---           | ---                    | ---                   | ---                         |
    | `/`      | `main.tex`    | `sections/*.tex`       | `latexmk main.tex`    | `latexmk parts/chapter.tex` |
    | `/`      | `main.tex`    | `parts/sections/*.tex` | `latexmk main.tex`    | `latexmk parts/chapter.tex` |
    | `/parts` | `main.tex`    | `sections/*.tex`       | `latexmk ../main.tex` | `latexmk chapter.tex`       |
    | `/parts` | `main.tex`    | `parts/sections/*.tex` | `latexmk ../main.tex` | `latexmk chapter.tex`       |
    | `/parts` | `../main.tex` | `sections/*.tex`       | `latexmk ../main.tex` | `latexmk chapter.tex`       |
    | `/parts` | `../main.tex` | `parts/sections/*.tex` | `latexmk ../main.tex` | `latexmk chapter.tex`       |

  Here I would expect the commands in rows 2 and 5 to both succeed, and all other commands to fail.

> Per above, making the optimistic and simultaneously pessimistic assumption that I haven't messed up in the testing, I think _"This is important, because the root directory in turn specifies the parent path for nested imports, includes, subfiles, includegraphics, and similar."_ might not be correct --- at least for `\import{}` and `\subfiles{}`, perhaps not for `\includegraphics{}` either per `subfile2.tex` in the `subfiles` documentation sec. 3.1.

I don't quite see why my description is wrong. That is, if my above table of suggested commands with my expactations of the outcome is correct, then I think it confirms my text. Except for `\import{}`, which I think you are right about. As we've already discussed, `\import` does change the CWD (it is the entire point of the package and command).

thor:
> Thanks for the thorough post. I've tried to respond in depth. I hope you might find time to test my below table of commands and verify (or falsify) my expectations.

No worries, it's the least I can as someone who opened an issue on your issue tracker. Both for the previous post and this one I'd like to apologise for the occasional misunderstanding due to excessive brewity or incorrect assumptions. I think it's pretty clear I've got some terminology mixed up.

> > To clarify that I do not misunderstand: that's absolute paths being relative paths seen from `main.tex` vs. relative paths being the path to `main.tex` as seen from each individual `*.tex` file.
> 
> No, with absolute path I mean paths like `/home/user/path/main.tex`, i.e. paths that begin with `/`. Relative paths are paths that are not absolute, e.g. `../main.tex`. This latter form requires a well defined starting point.

That's my usual understanding of absolute vs relative paths too. In the case of the issue #155 it seemed to be used to distinguish between paths relative to themselves and paths relative to the root document. To be clear on my own use, any further use of root path on my end refers to relative paths from the location/working directory of the *main document*.

>So, to be pragmatic: we need to figure out how to handle things within vimtex. A couple of thoughts related to your comments first.
>
>> Right, indeed. If the current working directory remains the root folder, then one would assume all paths should be relative to the current working directory. However, I'm not sure that is the case.
>
>Yes, this was what I tried to say in my previous comment. Still, ...

Ah, my apologies for not catching that.

>
>> From the `subfiles` documentation it is mentioned that `\\subfile{}` itself uses `import`. I interpret that to mean that any use of `\\subfile{}` will be rooted in the directory of the file including it. Assuming that is the case the only documents that can use a path relative to `main.tex` are those in the same folder as `main.tex`, whereas all others must use a path relative to their own location.
>
>I agree, from the `subfiles` docs it would seem like one should have a similar behaviour. However, as you show quite clearly, this is not the case. `subfiles` seems to be strongly connected to the original working directory. Some remarks:

A quick reply here: I think it may be connected (primarily) to the original working directory for the primary compilation target, but not equally so for the included subfiles, as ...

>* In your tests, with `CWD` set to `/`, did you use `latexmk parts/chapter.tex`? And similar, with `CWD` set to `/parts`, did you run `latexmk ../main.tex`? In general, I think the latter variant would not work.
>
>* Based on my previous comment, I would propose the following table of tests:
>
> ... [snip] ...

I'm not sure if it does show that it is not the case. Firstly, to clear up the table in the previous post. The invocations of `latex main.tex` and `latex chapter.tex` were with their relative paths, thus `latex ../main.tex` when `cwd` was `/parts`, and `latex parts/chapter.tex` when `cwd` was `/`.
As such I believe the table in the previous post covers all of the six invocations requested.
(I agree that `latex ../main.tex` from within `parts` was expected to fail, but I thought I might as well included it for the sake of the structure in the table, heh.)

The ❌ specifically meant that compilation failed either due to not finding `main.tex` from the `subfiles` document class parameter or `first.tex`. The latter was the primary cause.

>Here I would expect the commands in rows 2 and 5 to both succeed, and all other commands to fail.

As far as I am able to reproduce it correctly they do not succeed (for both command 1 and 2).

| CWD      | `chapter.tex` root          | Subfiles               | Command 1             | Command 2                   | C1 | C2 |
| ---      | ---           | ---                    | ---                   | ---                         |:--:|:--:|
| `/`      | `main.tex`    | `parts/sections/*.tex` | `latexmk main.tex`    | `latexmk parts/chapter.tex` | ❌ | ✔ |

This is the invocation from row 2: building `parts/chapter.tex` succeeds, but `main.tex` does not.

The only way I can rationalise that (absent other things interferring) is that the inclusion of `parts/chapter.tex` from `main.tex` results in `cwd` being changed to `/parts` when `\subfile` is used within `parts/chapter.tex`, thus unable to find `parts/sections/*.tex` from its location. However, quite why that does not happen when `parts/chapter.tex` is built directly is something I can't quite explain. Does `import` possibly work differently depending on whether it's building a subfile independently or whether it is being included? That *seems* sensible to me, but I'm not sure. (A worthwhile idea here would be to replace the `\subfile` invocations in `parts/chapter.tex` to see if it is `import` specific or `subfile` specific.)

| CWD      | `chapter.tex` root          | Subfiles               | Command 1             | Command 2                   | C1 | C2 |
| ---      | ---           | ---                    | ---                   | ---                         |:--:|:--:|
| `/parts` | `../main.tex` | `sections/*.tex`       | `latexmk ../main.tex` | `latexmk chapter.tex`       | ❌ | ✔ |

This in the invocation from row 7: building `chapter.tex` succeeds, but `main.tex` does not (as expected).

(In the event row 2 and 5 included the header row then, sans header row, row 1 and row 6 are those invocations, where for the first case only `latex main.tex` works, and neither works for the second.)

>> Per above, making the optimistic and simultaneously pessimistic assumption that I haven't messed up in the testing, I think _"This is important, because the root directory in turn specifies the parent path for nested imports, includes, subfiles, includegraphics, and similar."_ might not be correct --- at least for `\import{}` and `\subfiles{}`, perhaps not for `\includegraphics{}` either per `subfile2.tex` in the `subfiles` documentation sec. 3.1.
>
>I don't quite see why my description is wrong. That is, if my above table of suggested commands with my expactations of the outcome is correct, then I think it confirms my text. Except for `\import{}`, which I think you are right about. As we've already discussed, `\import` does change the CWD (it is the entire point of the package and command).

I find it somewhat strange myself, but if you'd like I'll whip up a repository for attempting to reproduce it with the various variants if you'd like? Might uncover what's causing the different experience.

lervag:
> No worries, it's the least I can as someone who opened an issue on your issue tracker. Both for the previous post and this one I'd like to apologise for the occasional misunderstanding due to excessive brewity or incorrect assumptions. I think it's pretty clear I've got some terminology mixed up.

No problem, I mix up things all the time :)

> That's my usual understanding of absolute vs relative paths too. In the case of the issue #155 it seemed to be used to distinguish between paths relative to themselves and paths relative to the root document. To be clear on my own use, any further use of root path on my end refers to relative paths from the location/working directory of the _main document_.

Ok.

> > > From the `subfiles` documentation it is mentioned that `\\subfile{}` itself uses `import`. I interpret that to mean that any use of `\\subfile{}` will be rooted in the directory of the file including it. Assuming that is the case the only documents that can use a path relative to `main.tex` are those in the same folder as `main.tex`, whereas all others must use a path relative to their own location.
> > 
> > 
> > I agree, from the `subfiles` docs it would seem like one should have a similar behaviour. However, as you show quite clearly, this is not the case. `subfiles` seems to be strongly connected to the original working directory. Some remarks:
> 
> A quick reply here: I think it may be connected (primarily) to the original working directory for the primary compilation target, but not equally so for the included subfiles, as ...

Ah, yes. I think I see now. It is not so hard, and it is exactly as stated in the docs: `\subfile` works exactly the same as specified in the `import` package. Each `\subfile` command will change the CWD for the subfile, and so the main path must be relative to the subfile. The example on page 3 in [the docs for `subfiles`](https://ctan.uib.no/macros/latex/contrib/subfiles/subfiles.pdf) is actually quite clear.

This is the effect on your example:

* `main.tex` says `\subfile{parts/chapter.tex}`, this implies that within `chapter.tex` the CWD for further `\include` type commands will be `parts/`. Thus, the subsequent `\subfile{sections/first.tex}` works as expected when we compile `main.tex`.
* However, this also means that we need to compile `chapter.tex` from `parts/`. Thus, for things to work consistently, it is also necessary to specify the `main.tex` path relative to the current subfile. That is, we need `\documentclass[../main.tex]{subfiles}` in `chapter.tex`.

The following update to the docs should then make things correct and consistent with the subfiles documentation, I think:

```
  Absolute paths to the main file (e.g. `/home/user/main.tex`) will work as
  expected. Relative paths must be relative to the current subfile. So, for
  a project structured like this:

    main.tex
    sub/sub.tex

  We need `\documentclass[../main.tex]{subfiles}` in `sub.tex`.
```

Note: This also requires an update of vimtex. Before I look into the code, it would be good to hear if you agree with me.

thor:
> > > > From the `subfiles` documentation it is mentioned that `\\subfile{}` itself uses `import`. I interpret that to mean that any use of `\\subfile{}` will be rooted in the directory of the file including it. Assuming that is the case the only documents that can use a path relative to `main.tex` are those in the same folder as `main.tex`, whereas all others must use a path relative to their own location.
> > > 
> > > 
> > > I agree, from the `subfiles` docs it would seem like one should have a similar behaviour. However, as you show quite clearly, this is not the case. `subfiles` seems to be strongly connected to the original working directory. Some remarks:
> > 
> > 
> > A quick reply here: I think it may be connected (primarily) to the original working directory for the primary compilation target, but not equally so for the included subfiles, as ...
> 
> Ah, yes. I think I see now. It is not so hard, and it is exactly as stated in the docs: `\subfile` works exactly the same as specified in the `import` package. Each `\subfile` command will change the CWD for the subfile, and so the main path must be relative to the subfile. The example on page 3 in [the docs for `subfiles`](https://ctan.uib.no/macros/latex/contrib/subfiles/subfiles.pdf) is quite clear.
Right! Every invocation of `\subfile{}` changes it; the contents of the document class naturally does not matter to the inclusion of the file itself.

> This is the effect on your example:
> 
>     * `main.tex` says `\subfile{parts/chapter.tex}`, this implies that within `chapter.tex` the CWD for further `\include` type commands will be `parts/`. Thus, the subsequent `\subfile{sections/first.tex}` works as expected when we compile `main.tex`.
> 
>     * However, this also means that we need to compile `chapter.tex` from `parts/`. Thus, for things to work consistently, it is also necessary to specify the `main.tex` path relative to the current subfile. That is, we need `\documentclass[../main.tex]{subfiles}` in `chapter.tex`.

Indeed, that seems to fit rather well with my earlier experience of trying to compile both documents.
(A note is that is also how the LaTeX Workshop extension for VSCode seems to do it.)

> The following update to the docs should then make things correct and consistent with the subfiles documentation, I think:
> 
> ```
>   Absolute paths to the main file (e.g. `/home/user/main.tex`) will work as
>   expected. Relative paths must be relative to the current subfile. So, for
>   a project structured like this:
> 
>     main.tex
>     sub/sub.tex
> 
>   We need `\documentclass[../main.tex]{subfiles}` in `sub.tex`.
> ```
> 
> Note: This also requires an update of vimtex. Before I look into the code, it would be good to hear if you agree with me.

I agree. FWIW I think there might be some curious or minor frustrations when compiling from different working directions. For instance, how to deal with `latexmkrc` in the root folder? Then there's `build_dir` for the `vimtex_compiler_latexmk`; should it always be presumed to be relative to the root location? While I think the latter question is easy, I imagine there might be other issues or things that might jam the gears a bit for a "simple" or entirely non-invasive change?

All the better if that's not the case though. :)

lervag:
> I agree. FWIW I think there might be some curious or minor frustrations when compiling from different working directions. For instance, how to deal with latexmkrc in the root folder? Then there's build_dir for the vimtex_compiler_latexmk; should it always be presumed to be relative to the root location? While I think the latter question is easy, I imagine there might be other issues or things that might jam the gears a bit for a "simple" or entirely non-invasive change?

Yes, definitely. I agree. However, I think we should take things one step at a time. That is: I'll now fix the code and make things work as I've proposed. I think that should make things consistent with the `subfiles` and `import` packages.

Then, afterwards, you or others can open new issues to raise these minor frustrations that follow. Some of them may be difficult to handle (e.g. dealing with `latexmkrc` in the root folder), but others might be possible to find solutions for.

thor:
That sounds good. A quick follow-up is that after playing around with https://github.com/James-Yu/LaTeX-Workshop I wonder _not_ changing the directory is the ideal approach for the `latexmk` engine specifically. I'm not sure if perhaps we should move this to a dedicated issue as this might not be quite the place, but in the spirit of brewity: `latexmk` has a `-cd`/`$do_cd` alternative that kind of solves parts of the issue.

It's quite neat, but creates a kind of dependency on `latexmk`.It changes the directory when building the file, so you do not have to fiddle or struggle with getting the `latexmkrc` loaded depending on where you are in the tree.

lervag:
> That sounds good.

Great, I'll update the PR when I get the time.

> A quick follow-up is that after playing around with https://github.com/James-Yu/LaTeX-Workshop I wonder _not_ changing the directory is the ideal approach for the `latexmk` engine specifically. I'm not sure if perhaps we should move this to a dedicated issue as this might not be quite the place, but in the spirit of brewity: `latexmk` has a `-cd`/`$do_cd` alternative that kind of solves parts of the issue.

> It's quite neat, but creates a kind of dependency on `latexmk`.It changes the directory when building the file, so you do not have to fiddle or struggle with getting the `latexmkrc` loaded depending on where you are in the tree.

That could be neat, but I'm not quite sure. In any case, I think it best to first make things work in general, then we can look into improvements at the next stage. So, feel free to open a new issue after this PR is finished and merged, and we can continue the discussion then. Sounds good?

thor:
Sounds good! I'll follow-up with an attempt at adequately described issues.
:)

lervag:
I think the newly updated (force pushed) branch should work according to the updated docs. I would be happy if you could test if I've missed anything, then I'll merge to master if things work as expected now.

> Sounds good! I'll follow-up with an attempt at adequately described issues.

Great. And no worries :)

thor:
While it, unfortunately, breaks my current workflow by changing the directory it seems to otherwise work just as expected! :+1:

lervag:
Good. Then I'll merge shortly.

lervag:
Done. Feel free to open follow up issues. :)

PR #1705, opened 2020-05-28
Add markdown to verbatim environments for indent

Carltoffel:


lervag:
It would be good to have a sample and a more detailed description here. Why is this useful/necessary?

Carltoffel:
The markdown package allows you to include markdown code in LaTeX.
One way to do this is to write it inline between `\begin/end{markdown}`.
The problem with the indentation is that the code inside the markdown environment is sensitive to the level of indent.
Therefore I added `markdown` to the list of verbatim environments so the automatic indentation `gg=G` does not messes with that.

```
\begin{markdown}
no leading white spaces allowed
\end{markdown}
```

To be honest, I don't know much about vim script, and it appears to be working with only the second line changed for me.
How to use the markdown package -> [link](https://www.overleaf.com/learn/how-to/Writing_Markdown_in_LaTeX_Documents)

lervag:
Thanks! I've updated the code now. I basically applied your suggestion, but I did some refactoring on top and I added a test for it.

PR #1709, opened 2020-05-31
Textobj variant targets: map ie,ae also explicitly

kiryph:
For consistency with the behavior for ic,ac.
Users of vim-textobj-entire are affected.
Motivated by https://vi.stackexchange.com/q/25633

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -586,11 +586,18 @@ function! s:init_default_mappings() abort " {{{1
     if vimtex#text_obj#targets#enabled()
       call vimtex#text_obj#targets#init()
 
-      " These are handled explicitly to avoid conflict with gitgutter
+      " These are handled explicitly to overwrite plugins
+      " which define these text objects gloablly
+      " e.g. gitgutter (ic, ac) and vim-textobj-entire (ie, ae)
       call s:map('x', 'ic', '<plug>(vimtex-targets-i)c')
       call s:map('x', 'ac', '<plug>(vimtex-targets-a)c')
       call s:map('o', 'ic', '<plug>(vimtex-targets-i)c')
       call s:map('o', 'ac', '<plug>(vimtex-targets-a)c')
+      call s:map('x', 'ie', '<plug>(vimtex-targets-i)e')
+      call s:map('x', 'ae', '<plug>(vimtex-targets-a)e')
+      call s:map('o', 'ie', '<plug>(vimtex-targets-i)e')
+      call s:map('o', 'ae', '<plug>(vimtex-targets-a)e')
+      
     else
       if g:vimtex_text_obj_variant ==# 'targets'
         call vimtex#log#warning(


lervag:
Thanks!

MahbubAlam231:
This didn't fix the problem.

lervag:
Yes, perhaps I was too quick to merge this. @kiryph did you test this yourself?

kiryph:
@lervag 
Sorry, I did not test it. As mentioned it was intended for users of `targets.vim`.

@MahbubAlam231 Have you installed `targets.vim` as well?  You did not mention this on vi.stackexchange.com

But this does not really matter, since this PR does not fix it for a user of all three plugins as well.
The answer by filibranden should solve your problem.

@lervag Can you undo my PR?

The code change does not introduce any regression, but the commit message and in-source comments are incorrect. Futhermore, the code change does not help anybody right now (except there is a plugin similar to vim-gitgutter).

Sorry again for not testing first.

---

A short elaboration why there is a specialization for `vim-gitgutter`:

When I read the question by @MahbubAlam231, I recalled that we specialized `ic` and `ac` but not `ie` and `ae`. However, I did not recall properly the reasoning of the specialization.

My first thought was that ftplugins take higher priority than global plugins. But this is not the case.

After a little bit of research about how `vim-gitgutter` installs it mappings, I figured it out:
`vim-gitgutter` installs it mappings when the autocommand `BufEnter` is executed which is later than when vimtex installs its mappings. This particular behavior of `vim-gitgutter` resulted in an inconsistency between the two text object variants `'vimtex'` and `'targets'` of vimtex which was fixed by the specialization.

Since `vim-textobj-entire` installs its mappings on source-time, this PR does not change which mapping is active when vimtex, targets.vim and vim-textobj-entire are installed.

There might be or come a plugin with ie/ae similar to vim-gitgutter which could take advantage of this PR. But right now I am not aware of any.

kiryph:
Just another thought about the matter of having `vim-textobj-entire` and `vimtex` installed.
Ideally, one would want to use both. So IMHO the actual question would be finding an alternative letter for one of them. Two suggestions from me would be:

1. Redefine vim-textobj-entire to `ia` and `aa` (mnemonic `a` for all)
There would be clash with the typical plugin text object function argument, e.g. available by targets.vim. But if you do not use `a` for argument, this sounds like a perfect candidate.

2. Redefine vimtex text object environment to `iv`and `av`. https://github.com/kana/vim-textobj-user/wiki lists only vim-textobj-continuous-line which uses `iv` and `av`. However, I think there could be a better one for continuous line such as `il` and `al`.

lervag:
> @lervag Can you undo my PR?

Done.

> Sorry again for not testing first.

No problem, sorry for not testing before merging! :)

> Redefine vimtex text object environment to `iv` and `av`.
> ...
> (Clarification: these suggestions are intended for the vimrc of a user, not a change for vimtex)

Thanks, I was about to comment that I don't want to change the vimtex defaults here, as I think they are natural and already "established".

@MahbubAlam231 Again, sorry for the inconvenience. I don't see a good solution on the vimtex side to the issue raised in your SE thread. I think the solution by @filbranden to explicitly add the mappings in your `after/ftplugin/tex.vim` file should work and should be an adequate solution here.

@filbranden also writes:

> Local mappings (created with the `<buffer>` modifier to `:map` and related commands) override global ones, so the expected behavior is that the vimtex mappings should already supercede the key-bindings from vim-textobj-entire.

I think it would be possible to change the `s:map` function in vimtex so that it will override global mappings, but not buffer local mappings. However, although I agree in principle, I don't think this is a good thing to enforce. The reason: Most people are not fully consistent in their mappings. E.g., they would make their own custom mapping for something, either globally or for LaTeX files (they might not know to use the `<buffer>` modifier). In both cases, they would not expect the vimtex mappings to override their custom mappings.

I'm obviously interested to hear differing opinions on the matter!

filbranden:
My thought was to possibly introduce a `g:vimtex_override_global_mappings` or similar configuration variable and check that in `s:map()` before the `mapcheck()`. There's already a local forcing override with `a:0 > 0` but adding a global one might not be a bad idea. I can send a PR for that if you'd like.

lervag:
Yes, this sounds reasonable. I would call the option `g:vimtex_mappings_override_global` and I would set the default value so that it does not change the current behaviour.

MahbubAlam231:
Answer by filbranden does solve my problem.

Introducing `g:vimtex_mappings_override_global` variable is a good idea. Maybe further introduce an option to select which global mappings to override?

As @lervag said "Most people are not fully consistent in their mappings. E.g., they would make their own custom mapping for something, either globally or for LaTeX files (they might not know to use the <buffer> modifier). In both cases, they would not expect the vimtex mappings to override their custom mappings." People might want some mappings of `vimtex` and some of their custom mappings, and in that case setting `g:vimtex_mappings_override_global = 1` would override all their mappings without `<buffer>` modifier. If there could be an option to select which global mappings to override, that would be great.

lervag:
> As @lervag said "Most people are not fully consistent in their mappings. E.g., they would make their own custom mapping for something, either globally or for LaTeX files (they might not know to use the `<buffer>` modifier). In both cases, they would not expect the vimtex mappings to override their custom mappings." People might want some mappings of `vimtex` and some of their custom mappings, and in that case setting `g:vimtex_mappings_override_global = 1` would override all their mappings without `<buffer>` modifier. If there could be an option to select which global mappings to override, that would be great.

No, I disagree. First, I generally want to keep the number of options as low as possible. Adding _two_ options for this seems overkill. So, ok, we could combine them. But still, if you were to explicitly declare each global mapping to override, then you could as well just override them the manual way in your `after/ftplugin/tex.vim` anyway. The number of config lines would also be similar. I think the reason for such an option is to make vimtex behave the way some people expect (i.e., if you expect vimtex to override global mappings).

```vim
" This feels simple and "right"
let g:vimtex_mappings_override_global = 1

" This feels less simple and less "right" to me
let g:vimtex_mappings_override_global = [
      \ '<plug>(vimtex-ie)',
      \ '<plug>(vimtex-ae)',
      \]
```

OTOH, if others agree with you I'm not going to be stubborn! :)

filbranden:
Follow up on #1711. 

Thanks!
Filipe

lervag:
Great, thanks a bunch, @filbranden. Let's continue the discussion there to avoid multiple threads on a single topic.

lervag:
#1711 is now merged and the option `g:vimtex_mappings_override_existing` has been added. If it is set to 1, vimtex will override existing mappings on conflicts, which should solve the original "issue" described on stackexchange.

PR #1711, opened 2020-06-01
Allow vimtex to override global mappings that use the same key sequences.

filbranden:
For example, vim-textobj-entire defines `ie` and `ae` which conflict with vimtex's mappings for an environment, so by default vimtex doesn't define its own `ie` and `ae`.

Use:

```vim
let g:vimtex_mappings_override_global = 1
```

In your vimrc to have vimtex define its own mappings even though they will shadow those from a different package.

Tested with vim-textobj-entire, after setting `g:vimtex_mappings_override_global` in my vimrc I was able to use `vie` and `vae` to select a LaTeX environment and not the entire contents of the buffer. The vim-textobj-entire mappings still worked outside of this context.

Also tested the operator-pending mapping with `dae`.

And inspected the mappings to confirm the buffer-local one from vimtex was shadowing the one from vim-textobj-entire:

```vim
:xmap ie
x  ie           @<Plug>(vimtex-ie)
x  ie            <Plug>(textobj-entire-i)
```

cc @kiryph, since this PR is a follow up for #1709.

Also motivated by https://vi.stackexchange.com/q/25633.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -204,6 +204,7 @@ function! vimtex#init_options() abort " {{{1
 
   call s:init_option('vimtex_mappings_enabled', 1)
   call s:init_option('vimtex_mappings_disable', {})
+  call s:init_option('vimtex_mappings_override_existing', 0)
 
   call s:init_option('vimtex_matchparen_enabled', 1)
   call s:init_option('vimtex_motion_enabled', 1)
@@ -470,7 +471,8 @@ function! s:init_default_mappings() abort " {{{1
   function! s:map(mode, lhs, rhs, ...) abort
     if !hasmapto(a:rhs, a:mode)
           \ && index(get(g:vimtex_mappings_disable, a:mode, []), a:lhs) < 0
-          \ && (empty(maparg(a:lhs, a:mode)) || a:0 > 0)
+          \ && (a:0 > 0 || get(g:, 'vimtex_mappings_override_existing', 0) > 0
+                \ || empty(maparg(a:lhs, a:mode)))
       silent execute a:mode . 'map <silent><nowait><buffer>' a:lhs a:rhs
     endif
   endfunction
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1782,6 +1782,14 @@ OPTIONS                                                        *vimtex-options*
 <
   Default value: {}
 
+*g:vimtex_mappings_override_existing*
+  Control behaviour on mapping conflicts, in particular whether or not to
+  override pre-existing mappings. By default, vimtex does not override existing
+  mappings. If this option is enabled, then vimtex will override existing
+  mappings on conflict.
+
+  Default value: 0
+
 *g:vimtex_matchparen_enabled*
   Enable highlighting of matching delimiters.
 


autoload/vimtex.vim
```diff
@@ -470,7 +470,8 @@ function! s:init_default_mappings() abort " {{{1
   function! s:map(mode, lhs, rhs, ...) abort
     if !hasmapto(a:rhs, a:mode)
           \ && index(get(g:vimtex_mappings_disable, a:mode, []), a:lhs) < 0
-          \ && (empty(maparg(a:lhs, a:mode)) || a:0 > 0)
+          \ && (a:0 > 0 || get(g:, 'vimtex_mappings_override_global', 0) > 0
+                \ || empty(maparg(a:lhs, a:mode)))
```
lervag, line 474: This does not check if the mapping to overwrite is really global, does it? The `maparg()` function does allow to fetch this kind of information (it seems this is available since Vim 7.3, vimtex currently aims to support Vim 7.4+). So, an improvement here would be to use something like this:

```vim
    let l:map = maparg(a:lhs, a:mode, '', 1)
    if !hasmapto(a:rhs, a:mode)
          \ && index(get(g:vimtex_mappings_disable, a:mode, []), a:lhs) < 0
          \ && (a:0 > 0 || empty(l:map)
          \     || (g:vimtex_mappings_override_global && ! l:map.buffer))
```

I also like to initialize my options explicitly in [vimtex.vim](https://github.com/lervag/vimtex/blob/250fe4cf97090faf18db44c2f036103aae1fb501/autoload/vimtex.vim#L26) (which is mostly sorted alphabetically, although probably not perfectly).

lervag:
Thanks! This seems very close to what I was thinking of. I suggested an update. And, of course, we should make sure the test suite still runs. I can do that on my end if it does not immediately run on your end.

autoload/vimtex.vim
filbranden: I can see your point about the variable name being potentially misleading about overriding actually global options or not...

But I'd argue that the most useful behavior is to check for *any* mappings and always override them.

If the user wants to keep the buffer-local mappings of a different plug-in, they can always order the plug-ins to load such that vimtex loads first and the second plug-in loads second, therefore overriding the mappings. They can also use `g:vimtex_mappings_disable` to disable some mappings in specific that they don't want to clobber.

On the other hand, if they *did* want vimtex to override the local mappings from a different plug-in, this option would not help them and they wouldn't have an easy way to fix that other than redefining the mappings themselves... (Which is what we're fixing with this option.)

Perhaps we could rename this option, but I think I'm fine with `global` since that's really the most common case when you'll use it. And since I'd imagine most users might not even really known the difference between global and buffer-local mappings, from their point of view they might be "global" anyways...

But perhaps we should rename it? What do you think?

filbranden:
Regarding testing... I started looking at setting up a Travis-CI for this, but then I noticed you seem to have a GitHub Actions setup for running tests. Why don't tests run on open PRs though? That would be very useful!

filbranden:
So for testing, I added Travis-CI configuration in #1712 and here's a successful run that includes both that PR and this one, all green: https://travis-ci.org/github/filbranden/vimtex/builds/693636809.

Let me know if you think this setting should have tests itself, something like adding an `ie` global mapping first and then testing this with/without the global variable set and verifying that the right text gets selected or acted upon.

Cheers!
Filipe

lervag:
I think I've fixed my Github Actions setup so it also runs on PR's. I guess you must push a commit for it to run (not sure, never tested this before).

filbranden:
I'll do a new push. What's your opinion on the logic of the check? And keep variable name or rename it?

lervag:
Sorry, was too busy to reply to the other parts earlier.

> If the user wants to keep the buffer-local mappings of a different plug-in, they can always order the plug-ins to load such that vimtex loads first and the second plug-in loads second, therefore overriding the mappings. They can also use g:vimtex_mappings_disable to disable some mappings in specific that they don't want to clobber.

Yes, good point.

> But perhaps we should rename it? What do you think?

I think the naming should be consistent with the behaviour. As such, I think we should rename. How about `g:vimtex_mappings_override_existing`? The docs should also be made consistent, e.g. something like this:

```
*g:vimtex_mappings_override_existing*
  Control behaviour on mapping conflicts, in particular whether or not to
  override pre-existing mappings. By default, vimtex does not override existing
  mappings. If this option is enabled, then vimtex will override existing
  mappings on conflict.

  Default value: 0
```

filbranden:
Pushed a new version with the rename of the variable and the update to the docs. Looks like CI is running now, which is nice!

lervag:
Great, thanks!

PR #1713, opened 2020-06-02
Install bibtex and bibparse from Ubuntu packages rather than CPAN

filbranden:
This should simplify and speed up initialization of the test container/VM.

Also simplify the Dockerfile a little further by using a single RUN command that does both `apt update` and `apt install`.

Tested the Dockerfile use case using podman:

```text
$ podman build -t vimtex-test docker/
$ podman run --volume "${PWD}":/vimtex vimtex-test make -C test/tests
```

The GitHub Actions case should be tested when the PR is opened.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,14 +22,14 @@ jobs:
     - name: Install packages
       run: |
         sudo apt update
-        sudo apt install neovim
-        sudo apt install texlive texlive-latex-extra texlive-extra-utils
-        sudo apt install texlive-bibtex-extra libtext-bibtex-perl
-        sudo apt install latexmk
-        sudo apt install gcc moreutils
-        sudo apt install libmodule-build-perl
-        sudo apt install libconfig-autoconf-perl
-        sudo apt install libextutils-libbuilder-perl
+        sudo apt install neovim \
+          texlive texlive-latex-extra texlive-extra-utils \
+          texlive-bibtex-extra libtext-bibtex-perl \
+          latexmk \
+          gcc moreutils \
+          libmodule-build-perl \
+          libconfig-autoconf-perl \
+          libextutils-libbuilder-perl
     - uses: actions/checkout@master
       with:
         fetch-depth: 1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,19 +24,12 @@ jobs:
         sudo apt update
         sudo apt install neovim
         sudo apt install texlive texlive-latex-extra texlive-extra-utils
+        sudo apt install texlive-bibtex-extra libtext-bibtex-perl
         sudo apt install latexmk
         sudo apt install gcc moreutils
         sudo apt install libmodule-build-perl
         sudo apt install libconfig-autoconf-perl
         sudo apt install libextutils-libbuilder-perl
-    - name: Install bibparse
-      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -22,21 +22,14 @@ jobs:
...
-        ./Build
-        sudo ./Build install
+        sudo apt install neovim \
+          texlive texlive-latex-extra texlive-extra-utils \
+          texlive-bibtex-extra libtext-bibtex-perl \
+          latexmk \
+          gcc moreutils \
+          libmodule-build-perl \
+          libconfig-autoconf-perl \
+          libextutils-libbuilder-perl
```
lervag, line 32: I'm curious: Is there a noticable benefit in keeping it as a single command?

docker/Dockerfile
```diff
@@ -16,20 +16,14 @@ FROM ubuntu
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-RUN apt -qq update
-RUN apt -qq -y install \
+RUN apt -qq update && apt -qq -y install \
```
lervag, line 19: As above: Is there a benefit in keeping it as a single `RUN` command?

.github/workflows/main.yml
filbranden: I haven't made any measurements, but `apt` will spend some time loading the package database and doing dependency resolution (which can be a somewhat expensive process), so doing that once for "n" packages rather than "n" times for a single package (or a few packages) each time should be beneficial.

docker/Dockerfile
filbranden: This one is actually a bit more clear cut in the impact, since each `RUN` command in a Dockerfile will create a separate layer and there's definitely overhead for each extra layer that you carry. If you look for general guides on optimizing Dockerfiles, reducing the number of layers is one recommendation you'll typically find.

(There are exceptions though, when the result of a command can be cached separately from another, then having two separate layers might be better during development since the lower, expensive layer can be cached. But in this case either everything gets cached or nothing does, so merging these two makes a lot of sense.)

filbranden:
Thanks for your prompt and insightful code reviews!

lervag:
> I haven't made any measurements, but apt will spend some time loading the package database and doing dependency resolution (which can be a somewhat expensive process), so doing that once for "n" packages rather than "n" times for a single package (or a few packages) each time should be beneficial.

Agreed, but I would guess this is a minor gain. In any case, I don't see any good reasons not to use a single command in this case.

> This one is actually a bit more clear cut in the impact, since each RUN command in a Dockerfile will create a separate layer and there's definitely overhead for each extra layer that you carry. If you look for general guides on optimizing Dockerfiles, reducing the number of layers is one recommendation you'll typically find.

Thanks! I've currently only used Docker to reproduce the CI environments locally for debugging (sometimes I have issues that I can't reproduce on my end). I've learned enough to make it work, but that's about it. I appreciate your input and suggestion for how to improve things!

> (There are exceptions though, when the result of a command can be cached separately from another, then having two separate layers might be better during development since the lower, expensive layer can be cached. But in this case either everything gets cached or nothing does, so merging these two makes a lot of sense.)

Yes, that makes sense.

> Thanks for your prompt and insightful code reviews!

My pleasure, and thank you for contributing :)

Issue #1395, opened 2019-05-27
New feature request (vimtex): equation number in the table of contents


zhuosongz:
This plugin provides the command `:VimtexTocOpen` to open the table of contents. Here is my tex file:

```tex
\documentclass{article}
\begin{document}
\section{Section One}
\label{sec:section_1}

This is the first equation:
\begin{equation}
1 + 1 = 2.
\label{eq:1}
\end{equation}

\noindent
This is the second equation:
\begin{equation}
1 \times 1 = 1.
\label{eq:2}
\end{equation}
\end{document}
```

When use `\lt`, I can get the following toc:

```
<Esc>/q  Close
<Space>  Jump
<Enter>  Jump and close
      r  Refresh
      h  Toggle help text
      t  Toggle sorted TODOs
    -/+  Decrease/increase ToC depth (for content layer)
    f/F  Apply/clear filter
      s  Hide numbering

Layers:  L  label+
         I  include+
         T  todo+
         C  content+

           Preamble
1          Section One
           sec:section_1
           eq:1
           eq:2
```

However, when I am editing a large file, there might be a lot of labels, so that I cannot remember the label of each equation, but can only see the equation number on the pdf file. Can I request a new feature that the equation numbers can also be displayed in the toc part, for example:

```
<Esc>/q  Close
<Space>  Jump
<Enter>  Jump and close
      r  Refresh
      h  Toggle help text
      t  Toggle sorted TODOs
    -/+  Decrease/increase ToC depth (for content layer)
    f/F  Apply/clear filter
      s  Hide numbering

Layers:  L  label+
         I  include+
         T  todo+
         C  content+

           Preamble
1          Section One
           sec:section_1
           eq:1                           [Eq. (1.1)]
           eq:2                           [Eq. (1.2)]
           thm:1                         [Thm. (1.1)]
```

Thank you a lot for your attention!

clason:
Right, I forgot to open that issue after #1191 was merged!

lervag:
Hmm. My initial reaction is "this is too difficult". However, I've looked a little bit at the code and think I might be able to make it work. I'll prioritize some other issues first, but here's a TODO list for how to work on this:

- [ ] Move `parse_aux_files` from `complete.vim` to `parse/...`
- [ ] Use the new `parse/...` as a backend for mapping labels to "title"/correct numbering

lervag:
[...]

PR #1738, opened 2020-07-07
Add label numbering to ToC

lervag:
This is a first version of an implementation for #1395 (see also #1191). Things seem to work well.

@clason 
@zhuosongz

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -366,7 +366,7 @@ function! s:init_highlights() abort " {{{1
         \ ['VimtexInfoTitle', 'PreProc'],
         \ ['VimtexInfoKey', 'PreProc'],
         \ ['VimtexInfoValue', 'Statement'],
-        \ ['VimtexInfoOther', 'Normal'],
+        \ ['VimtexInfoOther', ''],
         \ ['VimtexMsg', 'ModeMsg'],
         \ ['VimtexSuccess', 'Statement'],
         \ ['VimtexTocHelp', 'helpVim'],
@@ -376,7 +376,7 @@ function! s:init_highlights() abort " {{{1
         \ ['VimtexTocTodo', 'Todo'],
         \ ['VimtexTocNum', 'Number'],
         \ ['VimtexTocSec0', 'Title'],
-        \ ['VimtexTocSec1', 'Normal'],
+        \ ['VimtexTocSec1', ''],
         \ ['VimtexTocSec2', 'helpVim'],
         \ ['VimtexTocSec3', 'NonText'],
         \ ['VimtexTocSec4', 'Comment'],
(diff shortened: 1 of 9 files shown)


clason:
Very nice! I almost didn't notice it worked since the split is too narrow on my terminal -- maybe it makes sense to increase the width if the labels are shown?

lervag:
> The performance difference can be notable, though -- for my usual stress test project, it's 25s with labels vs. 5s without (on first call, after that it's instant thanks to caching). For "normal" projects it makes almost no difference, though.

Ah, yes, this is not so unexpected. I've tried to improve this by only loading the label dictionary once, instead of each time the parser function for label entries is called. It might improve things notably.

> Very nice! I almost didn't notice it worked since the split is too narrow on my terminal -- maybe it makes sense to increase the width if the labels are shown?

Yes, I could use `winwidth()` in the `printf` expression. Will push an update for this soon. Let me know what you think.

clason:
> Ah, yes, this is not so unexpected. I've tried to improve this by only loading the label dictionary once, instead of each time the parser function for label entries is called. It might improve things notably.

Yes, now it's down to 4-5s again 👍

> Yes, I could use winwidth() in the printf expression. Will push an update for this soon. Let me know what you think.

Hmm, this didn't change the width of the toc for me, but it does remove the spaces (and alignment) between the labels and the formatted references?

lervag:
> Hmm, this didn't change the width of the toc for me, but it does remove the spaces (and alignment) between the labels and the formatted references?

Ah, I misunderstood you. The width of the toc window is specified by `g:vimtex_toc_config`, specifically the `split_width` key. See the docs. Personally, I prefer to use `split_pos` equal to `full`, as I only use the toc for navigation (I don't keep it open anyway).

clason:
> Ah, I misunderstood you. The width of the toc window is specified by g:vimtex_toc_config, specifically the split_width key. See the docs. Personally, I prefer to use split_pos equal to full, as I only use the toc for navigation (I don't keep it open anyway).

Oh, right, should have read the doc 🤦 That's a good point! (Still think the default could be increased a bit :))

lervag:
Yes, I could increase the default. Noone has suggested it before, but I guess 30 characters is somewhat "thin" for this kind of window.

clason:
Especially now that it also shows the numbers. Not critical though, I just wanted to point out that you're hiding a shiny new feature ;)

lervag:
Ok, I've merged this now. Feel free to open issues if you find any!

lervag:
> Especially now that it also shows the numbers. Not critical though, I just wanted to point out that you're hiding a shiny new feature ;)

Thanks! I've changed to width 50, I think that should be better.

PR #1758, opened 2020-07-26
Add 'clip' to list of tikz commands

gustaphe:
Minor change, make 'clip' indent the same way 'path' does.


lervag:
Thanks! I've added this now (no merge, since you've misaligned the added line).

PR #1759, opened 2020-07-29
Compiler vlty: add test of Python version

matze-dd:
On some systems, the default Python is still version 2. We test on version \>= 3.6.


lervag:
Thanks!

PR #1762, opened 2020-07-30
New event: VimtexEventCompileSuccess

pseewald:
As discussed in #1760

### DIFF
--- a/autoload/vimtex/compiler.vim
+++ b/autoload/vimtex/compiler.vim
@@ -85,6 +85,12 @@ function! vimtex#compiler#callback(status) abort " {{{1
     endif
   endfor
 
+  if a:status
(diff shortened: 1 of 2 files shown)


lervag:
Thanks! However, I think the idea is that the event can do everything that the callback hook can and that it may be considered "simpler" and more clean. Thus, I want to make a breaking change and remove the old callback hook. This means a bigger change. I'll work on that myself after merging this. :)

lervag:
I've made the changes now. And, as always, there may be bugs. So don't hesitate to let me know if you find any!

pseewald:
I did not dare to remove callback hook myself, so thank you for taking care of this!

lervag:
I can understand that! :) I made the changes last night and also made some similar changes to other parts. Thanks for the suggestion and contribution!

PR #1789, opened 2020-08-30
Added QoL quickfix window resize options

Kimplul:
Hey there, first off, really cool plugin, thanks for all your work maintaining it. The one issue that I have with it is that the quickfix window is always 10 lines tall, even if the error or warning only takes up one line.

This is a fix for that issue, the quickfix window now resizes itself to fit the content up to a maximum height. The fix can be turned on or off.

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -245,6 +245,8 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_quickfix_open_on_warning', '1')
   call s:init_option('vimtex_quickfix_blgparser', {})
   call s:init_option('vimtex_quickfix_autoclose_after_keystrokes', '0')
+  call s:init_option('vimtex_quickfix_resize_to_content', '0')
+  call s:init_option('vimtex_quickfix_max_size', '10')
 
   call s:init_option('vimtex_syntax_enabled', 1)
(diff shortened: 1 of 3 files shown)


lervag:
> Hey there, first off, really cool plugin, thanks for all your work maintaining it.

Thanks!

> The one issue that I have with it is that the quickfix window is always 10 lines tall, even if the error or warning only takes up one line.
> 
> This is a fix for that issue, the quickfix window now resizes itself to fit the content up to a maximum height. The fix can be turned on or off.

I very much appreciate that you look into and propose a fix for what you find as an isue with vimtex!

However, I actually don't agree that the solution to your problem is to change vimtex (except perhaps that we should update the docs). The quickfix window is an ubiquitous part of Vim, and changing its behaviour similar to what you propose is really something you should do at a global level. In fact, there are plugins dedicated to this, such as these:

* https://github.com/romainl/vim-qf
* https://github.com/blueyed/vim-qf_resize

lervag:
Note, I personally use some custom configuration for this:

- https://github.com/lervag/dotvim/blob/master/personal/ftplugin/qf.vim
- https://github.com/lervag/dotvim/blob/master/personal/autoload/personal/qf.vim

Kimplul:
Oh cool, didn't know those existed. Thanks!

lervag:
No problem! :)

PR #1791, opened 2020-09-02
Minor (one-letter) change in vimtex.txt

krissen:
If content and *labels* are to be shown, the run variables ought to be 'cl', not 'ct'.


lervag:
Thanks! :)

PR #1794, opened 2020-09-12
Add option `g:vimtex_enable_mutate_iskeyword`

Iambecomeroot:
I HATE the way vimtex changes `iskeyword` without my permission. I don't
want `\textbf{test}` to be one word. I want it to behave as I expect. I
have shot myself in the foot with this so many times hitting ^W in
insert mode and deleting way too much.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -70,7 +70,9 @@ function! s:init_buffer() abort " {{{1
   " Apply some filetype specific settings
   if &filetype ==# 'tex'
     setlocal suffixesadd=.tex,.sty,.cls
-    setlocal iskeyword+=:
+    if g:vimtex_enable_mutate_iskeyword
+      setlocal iskeyword+=:
+    endif
     setlocal includeexpr=vimtex#include#expr()
     let &l:include = g:vimtex#re#tex_include
     let &l:define  = '\\\([egx]\|char\|mathchar\|count\|dimen\|muskip\|skip'
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -355,6 +355,8 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_view_zathura_options', '')
   call s:init_option('vimtex_view_zathura_check_libsynctex', 1)
 
+  call s:init_option('vimtex_enable_mutate_iskeyword', 1)
+
   let s:initialized = v:true
 endfunction
 


lervag:
> I HATE the way vimtex changes `iskeyword` without my permission.

You can easily change it back to whatever you want with an `after/ftplugin/tex.vim` file.

> I don't want `\textbf{test}` to be one word. I want it to behave as I expect.
> I have shot myself in the foot with this so many times hitting ^W in
> insert mode and deleting way too much.

vimtex changes the `iskeyword` by adding `:` to the default setting (see [here](https://github.com/lervag/vimtex/blob/ec219646136de3040356b00d0dc74520950bd761/autoload/vimtex.vim#L73)). Nothing else. The reason for this is that it is common to write labels like `\label{fig:...}` and `\label{eq:...}`, and it makes (some) sense to match entire labels as a single word.

So, the change to `iskeyword` by vimtex is not relevant to your example `\textbf{test}`. In fact, if I open `test.tex`, write `\textbf{test` then `<c-w>` (in insert mode), only `test` is removed. As expected.

Conclusion: I won't merge this, unless you can be more convincing.

Iambecomeroot:
My apologies. One of my plugins is messing with my `iskeyword` and vimtex is the only one I could find that changes that setting. I should have done more research. 

I did try changing the setting back in `after/ftplugin/tex/.vim`, but it did not work. I think the culprit plugin is setting it with an autocmd.

lervag:
> My apologies. One of my plugins is messing with my `iskeyword` and vimtex is the only one I could find that changes that setting. I should have done more research.

Apology accepted, thanks! Still, just FYI:

> I HATE the way vimtex changes iskeyword without my permission.

My first thought when I read this was "I never forced you to use vimtex, and if you HATE it then don't use it". I spend quite some time developing and maintaining it, and perhaps even more time responding to issue requests and helping people get into both Vim and vimtex. I hope you might think twice the next time you're writing an issue report or pull request. For instance, in this case you could have made the same point with something like this:

> I don't like that vimtex changes `iskeyword` during initialization.

Finally, you could try to use `verbose set iskeyword` to find where it was last set. It might help locate which line and file of vimscript that modifies the option (but not always, unfortunately).

Iambecomeroot:
Very good points @lervag. I apologize if my trying to get my point across came off as being rude. Very true that nobody is forcing me to use vimtex. I will try to do better next time.

Iambecomeroot:
I don't want you to think that I take your work for granted. Can I buy you a coffee?

lervag:
> Very good points @lervag. I apologize if my trying to get my point across came off as being rude. Very true that nobody is forcing me to use vimtex. I will try to do better next time.

Thanks, that's appreciated. I do understand that you meant no offense, but since I did get that immediate "bad feeling vibe" I thought it was OK to let you know.

> I don't want you to think that I take your work for granted. Can I buy you a coffee?

Sure, I do accept donations (see paypal link in README). However, I'm not poor and I don't really need the money, so please feel free to give to your favorite charity instead.

And please, don't be afraid to open new issues/PRs; I don't mind constructive criticisms and good discussions. Vimtex is by no means perfect, and I think it is constantly improving because people like you spend their time addressing weaknesses and suggesting improvements. :)

Iambecomeroot:
I think it might actually be a bug in vim, caused by vimtex. Sometimes, after compiling and getting an error, the behaviour of <kbd>w</kbd>, <kbd>b</kbd> etc. changes to jump basically everything but spaces (including `\n`). Running `verbose set iskeyword` shows `iskeyword=@,48-57,_,192-255,:/nvim/plugged/vimtex/autoload/vimtex.vim line 74`. But that `iskeyword` should not have that behaviour. And running `set iskeyword @,48-57,_,192-255,:` (exactly what vim reports it already is) works as expected.

I don't know if I should open a bug report because it's very hard to reproduce.

Also, I don't see a paypal link in the README.

lervag:
> I think it might actually be a bug in vim, caused by vimtex. Sometimes, after compiling and getting an error, the behaviour of w, b etc. changes to jump basically everything but spaces (including `\n`). Running `verbose set iskeyword` shows `iskeyword=@,48-57,_,192-255,:/nvim/plugged/vimtex/autoload/vimtex.vim line 74`. But that `iskeyword` should not have that behaviour. And running `set iskeyword @,48-57,_,192-255,:` (exactly what vim reports it already is) works as expected.
> 
> I don't know if I should open a bug report because it's very hard to reproduce.

Huh, that's strange. Have you remapped `w` somehow? Or are you typing `W` (capital `w`)? Note the difference between a `word` and a `WORD`, see `:help word`. Perhaps you're hitting caps lock?

> Also, I don't see a paypal link in the README.

It's on the very top in the form of a "badge". Anyway, here it is:
https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=5N4MFVXN7U8NW

Iambecomeroot:
Definitely not caps lock since that is bound to escape. Also not hitting shift. In any case. Resetting `iskeyword` to its current value would not affect caps, right? There is definitely something weird going on.

lervag:
> There is definitely something weird going on.

Yes, it seems so. What is the output of `:map w` when you're in the weird situations where `w` moves over too many characters?

Iambecomeroot:
Sorry for the wait; I had to wait for it to happen again.

`:map w` shows `No mapping found`.

lervag:
Ok. Do I understand correctly that:

1. Things work "fine" when you start Vim. That is, the problem only appears after some time.
2. The problem, when it appears, is that motions such as `w` and `e` (and similar) start to move as if `iskeyword` contains all non white space characters?

Further, you are not able to specify a simple way to reproduce the issue?

If possible, could you show a gif/video/screencast of how things look when it is does not work?

Which version of Vim are you on? Perhaps you could try if the problem is also present on neovim?

Iambecomeroot:
Yes, that is correct. No, I do not know a simple way to reproduce. All I know is "wait for it to happen". Next time it happens I will take a video.

I am on neovim actually. Version 0.4.4.

lervag:
> Yes, that is correct. No, I do not know a simple way to reproduce. All I know is "wait for it to happen". Next time it happens I will take a video.

Ok, good. I'm not really sure it will help, but who knows :)

> I am on neovim actually. Version 0.4.4.

Ah, ok. Then the opposite question: would the same issue occur in Vim? Perhaps
you could try to update to neovim nightly (0.5..)?

> Sorry for turning a closed pr into a bug report...

No problem! My main concern about issues/PRs is that I want to keep the discussion "single threaded" and consise, because I don't get to work that many hours on vimtex. Having the discussions "on topic" and clear makes it much easier to pick up the thread after some time off.

Iambecomeroot:
I have been using vim for a little while and have not been able to reproduce. So maybe it is a bug in neovim. Its certainly seems possible; the fix of setting `iskeyword` to itself certainly seems like buggy behaviour.

I will try neovim nightly next.

Iambecomeroot:
After sitting in nvim v0.5.0 for a while, I finally got the bug to show itself. [Here is a screencast.](https://i.imgur.com/oJMfSrB.mp4) I forgot to record it, but `verbose set iskeyword` shows
```
iskeyword=@,48-57,_,192-255,:/nvim/plugged/vimtex/autoload/vimtex.vim line 73
```
and `map w` is
```
No mapping found
```

lervag:
I'm sorry to say I can't really help further with this. My suggestion would be to look into if it is possible to reproduce this with a fully minimal vimrc file, and if so, open an issue with the neovim repo. But even then, I think it is really hard to work with this since there is no easy way to reliably reproduce this. :\

Iambecomeroot:
Ok. Its doesn't occur very often anymore and now I know the fix.

Issue #1799, opened 2020-09-18
[Feature] Full syntax support in vimtex


lervag:
This action was discussed in #1781. Here I summarize the tasks and actions for moving forward. There are two phases.

## Phase 1

The changes in phase 1 should not require any significant changes to the current vimtex syntax code. During development, the option `g:vimtex_syntax_alpha` will be defined as a temporary option that users can enable (set to 1) to test the WIP code.

- [x] Check if it is possible to use [vim-tex-syntax](https://github.com/gi1242/vim-tex-syntax) as a starting point.
  * Is there a licensing problem?
  * Will it require significant changes to current vimtex syntax scripts?
- [x] Use current version of the TeX syntax scripts (or vim-tex-syntax) as a starting point. That is, fork previous work and adopt into vimtex.
- [x] Clean up code, make things consistent with vimtex styles.
- [x] Possibly remove legacy stuff.

## Phase 2

This phase brings breaking changes. The `g:vimtex_syntax_alpha` option will be removed, and the development should be a separate branch and a PR.

- [ ] Update configuration -> vimtex configuration schemes. E.g. `let g:vimtex_syntax_config = {'conceal': ...}`.
- [x] Ensure syntax script is loaded _after_ filetype scripts (cf. #1809).
- [x] Move current vimtex syntax scripts from being `after/syntax/...` to being loaded properly on startup.
- [ ] More clean up and improvements to code.
- [ ] Optimize things and try to improve efficiency as much as possible.

krishnakumarg1984:
I really think you should leave the syntax support to a language server such as `texlab`.  The syntax support of LSP is based on the semantics!

lervag:
I partly agree, in that I also see the benefits of relying on the tree-sitter paradigm for syntax. However, I plan to keep support for Vim, and tree-sitter is currently only available for neovim (nightly). Thus I don't think that this is currently a possible option.

clason:
Note that
1. Semantic highlighting is not yet part of the LSP specification, and texlab doesn't support it (yet). And even if it did, you'd need a client that supports it (which the neovim built-in client doesn't, until it becomes part of the spec -- and even then it might not in favor of tree-sitter).
[...]

PR #1805, opened 2020-09-25
Textidote: translate localized &spelllang for textidote

bratekarate:
Maps `&spelllang` to the format that textidote understands. Allows the use of localized languages such as `en_gb`. Handles all languages that are supported by textidote as of the current version.

### DIFF
--- a/compiler/textidote.vim
+++ b/compiler/textidote.vim
@@ -17,7 +17,7 @@ endif
 let &l:makeprg = 'java -jar ' . shellescape(fnamemodify(s:cfg.jar, ':p'))
       \ . (exists(s:cfg.args) ? ' ' . s:cfg.args : '')
       \ . ' --no-color --output singleline --check '
-      \ . matchstr(&spelllang, '^\a\a') . ' %:S'
+      \ . <SID>GetTextidoteLang(&spelllang) . ' %:S'
 
 setlocal errorformat=
 setlocal errorformat+=%f(L%lC%c-L%\\d%\\+C%\\d%\\+):\ %m
@@ -28,3 +28,13 @@ silent CompilerSet errorformat
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
+
+function! s:GetTextidoteLang(lang)
+  if a:lang is# 'en_gb'
+    return 'en_UK'
+  endif
+
+  let l:start = matchstr(a:lang, '^\a\a')
+  return len(a:lang) == 2 ? l:start
+        \ : l:start . '_' . toupper(matchstr(a:lang, '\a\a$'))
+endfunction


lervag:
Thanks, this looks reasonable. I agree it could be useful or interesting to allow a config dictionary to specify the desired language when it is ambivalent.

bratekarate:
Thank you for including this feature! Also, good to see a better way of handling the varying length of `&spelllang` with only one matchlist execution in your commit  Still learning vimscript.

lervag:
My pleasure. Let me know if you think we need more possibilities for customisation here.

oblitum:
Vim supports multiple spell languages in `&spelllang`, I myself have it set to `set spelllang=en,pt_br`. Isn't that a problem because it seems code here just assumes one language, not too.

lervag:
Good point. Could you open a new issue for it, and I can try to make this general purpose enough to work for all grammar/language parsers?

bratekarate:
True, I did not think of multiple `&spelllang` either. But I don't agree that it should be left out all together -- multiple languages looks rather like an edge case to me, and it is easy enough to just use the first language by default. 

A variable to override `&spelllang` would basically suffice to cover all edge cases. I think it would be rather inconvenient for most users to manage the state of two, basically redundant variables. At least for me it would be reduntant, as I usually switch the spelllang instead of using multiple. If I write in German, I only want German words to be recognized.

On another note, I tested the compiler with multiple languages and found a little bug that causes e.g. `en_ca` to turn into `enCA` (missing underscore). However, this is not related to multiple spelllangs, but rather a general bug for localized languages. Other than this bug, the only thing that makes `get_textidote_lang` fail for multiple languages is the hard check on `en_gb`. 

I used a modified version of the function in [this PR for ale](https://github.com/dense-analysis/ale/pull/3376). The code of the ale linter can possibly be improved, but it may be helpful as it would cover the exact use case that is talked about here.

bratekarate:
@oblitum  You are probably right and more experienced than I am with vim. I still think that picking the first spelllang in the list is a reasonable default. Even if more than 50% of the users were to use multiple spelllangs, the others without multiple languages still profit from the default. The vimmers with multiple languages don't lose anything, having the config variable anyway. Even if you use multiple: Isn't it convenient to use the primary language and override the variable only if needed in buffer scope?

Anyway, I'm also ok with configuring another variable for my spellchecker if that's the consensus. Not a biggie.

oblitum:
@bratekarate I don't disagree with that, an override/fallback scheme as you say is just fine, it just not need to be hard-coded, my previous comment was just to point that it isn't much an edge case.

lervag:
It seems @oblitum has also suggested a solution for this. I hope it's OK that I continue this discussion at #1811 after considering the PR (I'm working on something else right now (#1799), so I won't promise it will happen immediately, sorry).

PR #1812, opened 2020-10-02
handle \input{...} when compiling a selection

mwageringel:
Without this, when compiling a selection, both the line `\input{...}` and the contents of the file are included in the temporary tex file. This was broken in 8b41f09e47fac4ee8fc4bd07a3a717bc69828859.

I am not sure what `parse_current` in `autoload/vimtex/parser/tex.vim` is used for. It might need a similar fix.

### DIFF
--- a/autoload/vimtex/parser/tex.vim
+++ b/autoload/vimtex/parser/tex.vim
@@ -197,11 +197,11 @@ function! s:parse_preamble(file, opts, parsed_files) abort " {{{1
       break
     endif
 
-    call add(l:lines, l:line)
-
     if l:line =~# g:vimtex#re#tex_input
       let l:file = s:input_parser(l:line, a:file, a:opts.root)
       call extend(l:lines, s:parse_preamble(l:file, a:opts, a:parsed_files))
+    else
+      call add(l:lines, l:line)
     endif
   endfor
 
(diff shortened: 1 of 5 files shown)


lervag:
> Without this, when compiling a selection, both the line `\input{...}` and the contents of the file are included in the temporary tex file. This was broken in [8b41f09](https://github.com/lervag/vimtex/commit/8b41f09e47fac4ee8fc4bd07a3a717bc69828859).

Ah, yes, I agree.

> I am not sure what `parse_current` in `autoload/vimtex/parser/tex.vim` is used for. It might need a similar fix.

Sort of. I think you could have the same issue in a similar "compile selected", but in most other use cases I don't think this issue is really a problem. Still, I'll look into if it is possible to fix it also in the non-preamble part.

lervag:
> Still, I'll look into if it is possible to fix it also in the non-preamble part.

Ah, no, sorry. The table of contents parser actually relies on having the information about included files, so removing these lines breaks the ToC.

lervag:
Thanks, I've merged your suggested change. I appreciate the effort and I'm impressed that I did not have to raise any suggestions for changing things. Very well done! I also appreciate that you caught a missing test (`output3`).

mwageringel:
Thank you.

PR #1818, opened 2020-10-06
Jump to previous/next math zone

### LINKED ISSUE #1810: Jump to next/previos math zone
vimtex has mappings for jumping forward and backward to different elements like environments, comments etc. (`[m [/` and family). Would it be possible to have a similar mapping to jump to next math element?

 For example in the following text:

```latex
This is normal text, here comes some math $\alpha + \beta = \gamma$.
Here's some more math $$e^{i\pi} + 1  = 0,$$ and finally an equation
\begin{equation}
a^2 + b^2 = c^2
\end{equation}
```
`]n` (or any other keybinding) should go to ` $\alpha + \beta = \gamma$`, then `$$e^{i\pi} + 1  = 0,$$` and then finally to the `equation` environment.

oponkork:
Fixes #1810.

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -180,6 +180,19 @@ function! s:init_default_mappings() abort " {{{1
     call s:map(1, 'o', '[]', '<plug>(vimtex-[])')
     call s:map(1, 'o', '[[', '<plug>(vimtex-[[)')
 
+    call s:map(1, 'n', ']N', '<plug>(vimtex-]N)')
+    call s:map(1, 'n', ']n', '<plug>(vimtex-]n)')
+    call s:map(1, 'n', '[N', '<plug>(vimtex-[N)')
+    call s:map(1, 'n', '[n', '<plug>(vimtex-[n)')
+    call s:map(1, 'x', ']N', '<plug>(vimtex-]N)')
+    call s:map(1, 'x', ']n', '<plug>(vimtex-]n)')
+    call s:map(1, 'x', '[N', '<plug>(vimtex-[N)')
+    call s:map(1, 'x', '[n', '<plug>(vimtex-[n)')
+    call s:map(1, 'o', ']N', '<plug>(vimtex-]N)')
+    call s:map(1, 'o', ']n', '<plug>(vimtex-]n)')
+    call s:map(1, 'o', '[N', '<plug>(vimtex-[N)')
+    call s:map(1, 'o', '[n', '<plug>(vimtex-[n)')
+
     call s:map(1, 'n', ']M', '<plug>(vimtex-]M)')
     call s:map(1, 'n', ']m', '<plug>(vimtex-]m)')
     call s:map(1, 'n', '[M', '<plug>(vimtex-[M)')
@@ -205,11 +218,6 @@ function! s:init_default_mappings() abort " {{{1
     call s:map(1, 'o', ']*', '<plug>(vimtex-]*)')
     call s:map(1, 'o', '[/', '<plug>(vimtex-[/)')
     call s:map(1, 'o', '[*', '<plug>(vimtex-[*)')
-
-    call s:map(1, 'n', ']N', '<plug>(vimtex-]N)')
-    call s:map(1, 'n', ']n', '<plug>(vimtex-]n)')
-    call s:map(1, 'n', '[N', '<plug>(vimtex-[N)')
-    call s:map(1, 'n', '[n', '<plug>(vimtex-[n)')
   endif
 
   if g:vimtex_text_obj_enabled
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -205,6 +205,11 @@ function! s:init_default_mappings() abort " {{{1
     call s:map(1, 'o', ']*', '<plug>(vimtex-]*)')
     call s:map(1, 'o', '[/', '<plug>(vimtex-[/)')
     call s:map(1, 'o', '[*', '<plug>(vimtex-[*)')
+
+    call s:map(1, 'n', ']N', '<plug>(vimtex-]N)')
+    call s:map(1, 'n', ']n', '<plug>(vimtex-]n)')
+    call s:map(1, 'n', '[N', '<plug>(vimtex-[N)')
+    call s:map(1, 'n', '[n', '<plug>(vimtex-[n)')
   endif
 
   if g:vimtex_text_obj_enabled
--- a/autoload/vimtex/motion.vim
+++ b/autoload/vimtex/motion.vim
@@ -229,7 +229,7 @@ function! vimtex#motion#math(begin, backwards, visual) abort " {{{1
         let l:restore_cursor = 0
         break
       elseif a:begin == 0
-        let l:pos_prev = vimtex#pos#prev(l:pos)
+        let l:pos_prev = vimtex#pos#prev(l:pos[1],l:pos[2]-1)
         if vimtex#syntax#in_mathzone(l:pos_prev[1],l:pos_prev[2],l:pos_prev[3])
             let l:restore_cursor = 0
             break
--- a/autoload/vimtex/motion.vim
+++ b/autoload/vimtex/motion.vim
@@ -225,23 +225,21 @@ function! vimtex#motion#math(begin, backwards, visual) abort " {{{1
       let l:iter += 1
       call search(l:re, l:flags)
       let l:pos = vimtex#pos#get_cursor()
-      if vimtex#syntax#in_mathzone(l:pos[1],l:pos[2],l:pos[3])
-        let l:restore_cursor = 0
-        break
-      elseif a:begin == 0
+      if a:begin == 0
         let l:pos_prev = vimtex#pos#prev(l:pos[1],l:pos[2]-1)
         if vimtex#syntax#in_mathzone(l:pos_prev[1],l:pos_prev[2],l:pos_prev[3])
             let l:restore_cursor = 0
             break
         endif
+      elseif vimtex#syntax#in_mathzone(l:pos[1],l:pos[2],l:pos[3])
+        let l:restore_cursor = 0
+        break
       endif
     endwhile
   endfor
   " Restore cursor position if fail
   if l:restore_cursor
     call vimtex#pos#set_cursor(l:curpos_saved)
-  else
-    call vimtex#pos#set_cursor(l:pos)
   endif
 endfunction
 
--- a/autoload/vimtex/motion.vim
+++ b/autoload/vimtex/motion.vim
@@ -44,10 +44,10 @@ function! vimtex#motion#init_buffer() abort " {{{1
   nnoremap <silent><buffer> <plug>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,0)<cr>
   nnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,0)<cr>
   nnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,0)<cr>
-  xnoremap <silent><buffer> <plug>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,1)<cr>
-  xnoremap <silent><buffer> <plug>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,1)<cr>
-  xnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,1)<cr>
-  xnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,1)<cr>
+  xnoremap <silent><buffer>  <sid>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,1)<cr>
+  xnoremap <silent><buffer>  <sid>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,1)<cr>
+  xnoremap <silent><buffer>  <sid>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,1)<cr>
+  xnoremap <silent><buffer>  <sid>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,1)<cr>
   xmap     <silent><buffer> <plug>(vimtex-]n) <sid>(vimtex-]n)
   xmap     <silent><buffer> <plug>(vimtex-]N) <sid>(vimtex-]N)
   xmap     <silent><buffer> <plug>(vimtex-[n) <sid>(vimtex-]n)
--- a/autoload/vimtex/motion.vim
+++ b/autoload/vimtex/motion.vim
@@ -82,6 +82,12 @@ function! vimtex#motion#init_buffer() abort " {{{1
         \ :execute "normal \<sid>(V)" . v:count1 . "\<sid>(vimtex-[/)"<cr>
   onoremap <silent><buffer> <plug>(vimtex-[*)
         \ :execute "normal \<sid>(V)" . v:count1 . "\<sid>(vimtex-[*)"<cr>
+
+  " Math
+  nnoremap <silent><buffer> <plug>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,0)<cr>
 endfunction
 
 " }}}1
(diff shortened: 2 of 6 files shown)


autoload/vimtex.vim
```diff
@@ -205,6 +205,11 @@ function! s:init_default_mappings() abort " {{{1
     call s:map(1, 'o', ']*', '<plug>(vimtex-]*)')
     call s:map(1, 'o', '[/', '<plug>(vimtex-[/)')
     call s:map(1, 'o', '[*', '<plug>(vimtex-[*)')
+
+    call s:map(1, 'n', ']N', '<plug>(vimtex-]N)')
+    call s:map(1, 'n', ']n', '<plug>(vimtex-]n)')
+    call s:map(1, 'n', '[N', '<plug>(vimtex-[N)')
+    call s:map(1, 'n', '[n', '<plug>(vimtex-[n)')
```
lervag, line 212: I would move these lines above the `<plug>(vimtex-*[mM])` lines to keep the lines somewhat sorted.

autoload/vimtex/motion.vim
```diff
@@ -82,6 +82,12 @@ function! vimtex#motion#init_buffer() abort " {{{1
         \ :execute "normal \<sid>(V)" . v:count1 . "\<sid>(vimtex-[/)"<cr>
   onoremap <silent><buffer> <plug>(vimtex-[*)
         \ :execute "normal \<sid>(V)" . v:count1 . "\<sid>(vimtex-[*)"<cr>
+
+  " Math
+  nnoremap <silent><buffer> <plug>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,0)<cr>
```
lervag, line 90: You should also create the corresponding `xnoremap` and `onoremap` mappings to make the mappings available in visual mode and in operator pending mode.

lervag: Note, you should also create the corresponding default maps in the other file.

autoload/vimtex/motion.vim
```diff
@@ -194,6 +200,45 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+  let l:curpos_saved = vimtex#pos#get_cursor()
+
+  let l:count = v:count1
+  if a:visual
+    normal! gv
+  endif
+
+  " Search for $, $$, \[, \(, \begin
+  " Use syntax to determine if we are inside math region
+  let l:re = g:vimtex#re#not_comment . (a:begin ? '%(\${1,2}|\\\[|\\\(|\\begin\s*\{)' : '%(\${1,2}|\\\]|\\\)|\\end\s*\{)')
```
lervag, line 214: Please avoid lines longer than 80 characters. In this case, I would write it like this:

```vim
let l:re = g:vimtex#re#not_comment . (a:begin
      \ ? '%(\${1,2}|\\\[|\\\(|\\begin\s*\{)'
      \ : '%(\${1,2}|\\\]|\\\)|\\end\s*\{)')
```

autoload/vimtex/motion.vim
```diff
@@ -194,6 +200,45 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+      if a:begin == 0
+          let l:pos = vimtex#pos#prev(l:pos[1],l:pos[2]-1)
+      endif
+      if vimtex#syntax#in_mathzone(l:pos[1],l:pos[2],l:pos[3])
+          let l:restore_cursor = 0
+          break
+      endif
+    endwhile
+  endfor
+  " Restore cursor position if fail
```
lervag, line 237: Add empty line above comment.

autoload/vimtex/motion.vim
```diff
@@ -194,6 +200,45 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+
+    " We need to restore cursor back to it's original position if jumping
+    " to count number of math environments fail.
+    let l:restore_cursor = 1
+    while l:iter <= 5
+      let l:iter += 1
+      call search(l:re, l:flags)
+      let l:pos = vimtex#pos#get_cursor()
+      if a:begin == 0
+          let l:pos = vimtex#pos#prev(l:pos[1],l:pos[2]-1)
```
lervag, line 229: Use two spaces for indents.

lervag: Also: I think you could do this as `vimtex#pos#prev(vimtex#pos#prev(l:pos))`, which would be slightly more robust, because `l:pos[2] - 1` does not make sense for the first column in a line.

autoload/vimtex/motion.vim
```diff
@@ -194,6 +200,45 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+      endif
+      if vimtex#syntax#in_mathzone(l:pos[1],l:pos[2],l:pos[3])
```
lervag, line 231: The `l:pos[3]` is not necessary.

autoload/vimtex/motion.vim
```diff
@@ -194,6 +200,45 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+  let l:re = g:vimtex#re#not_comment . (a:begin ? '%(\${1,2}|\\\[|\\\(|\\begin\s*\{)' : '%(\${1,2}|\\\]|\\\)|\\end\s*\{)')
+  let l:flags = 'W' . (a:backwards ? 'b' : '')
+
+  for l:_ in range(l:count)
+    " Ensure we are not going into infinite loop
+    let l:iter = 0
+
+    " We need to restore cursor back to it's original position if jumping
+    " to count number of math environments fail.
+    let l:restore_cursor = 1
```
lervag, line 223: Perhaps call this variable something like `l:success` (in which case you would also need to reverse the following logic)?

lervag:
I've made some relatively minor comments to the code. As a general comment: Good work! I think you are quite close.

I think it would be very good to add some tests. I think it should not be so hard. Something like this:

1. Create the file `test/tests/test-motions/test-math-motions.tex` and add example content.
2. Create the file `test/tests/test-motions/test-math-motions.vim` and write tests similar to [this](https://github.com/oponkork/vimtex/blob/master/test/tests/test-motions/test-method-motions.vim).

It should suffice, and you should be able to run the tests with `make` (runs all) or `make test-math-motions`. It will use neovim by default, if you want to run tests with Vim, you can do it with `MYVIM="vim -T dumb --not-a-term -n" make ...`.

autoload/vimtex/motion.vim
oponkork: Yes, for testing purposes, I created only the normal mode mappings, will add rest of them.

oponkork:
@lervag I have hit an issue with the following `tex` source
```latex
\begin{equation}
    1 + 1 = 2
\end{equation}

\( 1 + 1 = 2 \)

\[
 1 + 1 = 2
\]
```
Notation used to represent position: (line, col)
- Goto the beginning of the source and press `]N` it correctly takes you to (3,1)
- Press `]N` again and instead of jumping to `\)` on (5,14) it jumps to the `\]` on (9,1)
- If it works correctly, move the cursor around in the same line and try jumping again. 
- I think the problem is with `synstack` call in function `vimtex#syntax#stack`. When a call is made to get the syntax items of the position (5,12) (two columns left to the `\)` on (5,14), it returns an empty list, however when I call synstack manually for (5,12) it returns an integer.

[Here's an asciinema showing the same.](https://asciinema.org/a/OiqHw0WLlPczEJ01uBmJj2dh4)

P.S. I used the following modified version `vimtex#syntax#stack` to print the synstack and I have checked the behavior is the same on both vim and neovim.

```vim
function! vimtex#syntax#stack(...) abort " {{{1
  let l:pos = a:0 > 0 ? [a:1, a:2] : [line('.'), col('.')]
  if mode() ==# 'i'
    let l:pos[1] -= 1
  endif
  call map(l:pos, 'max([v:val, 1])')
  if a:0 > 0
      echo "In syntax#stack" l:pos synstack(l:pos[0], l:pos[1])
  endif
  return map(synstack(l:pos[0], l:pos[1]), "synIDattr(v:val, 'name')")
endfunction
```

autoload/vimtex/motion.vim
```diff
@@ -39,6 +39,28 @@ function! vimtex#motion#init_buffer() abort " {{{1
...
+  nnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,0)<cr>
+  nnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,0)<cr>
+  xnoremap <silent><buffer> <plug>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,1)<cr>
+  xnoremap <silent><buffer> <plug>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,1)<cr>
+  xnoremap <silent><buffer> <plug>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,1)<cr>
+  xnoremap <silent><buffer> <plug>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,1)<cr>
+  xmap     <silent><buffer> <plug>(vimtex-]n) <sid>(vimtex-]m)
+  xmap     <silent><buffer> <plug>(vimtex-]N) <sid>(vimtex-]N)
+  xmap     <silent><buffer> <plug>(vimtex-[n) <sid>(vimtex-]n)
+  xmap     <silent><buffer> <plug>(vimtex-[N) <sid>(vimtex-]N)
```
lervag, line 54: You made some minor mistakes here. It should be:

```vim
xnoremap <silent><buffer> <sid>(vimtex-]n) :<c-u>call vimtex#motion#math(1,0,1)<cr>
xnoremap <silent><buffer> <sid>(vimtex-]N) :<c-u>call vimtex#motion#math(0,0,1)<cr>
xnoremap <silent><buffer> <sid>(vimtex-[n) :<c-u>call vimtex#motion#math(1,1,1)<cr>
xnoremap <silent><buffer> <sid>(vimtex-[N) :<c-u>call vimtex#motion#math(0,1,1)<cr>
xmap     <silent><buffer> <plug>(vimtex-]n) <sid>(vimtex-]n)
xmap     <silent><buffer> <plug>(vimtex-]N) <sid>(vimtex-]N)
xmap     <silent><buffer> <plug>(vimtex-[n) <sid>(vimtex-]n)
xmap     <silent><buffer> <plug>(vimtex-[N) <sid>(vimtex-]N)
```

(I know this part is cryptic, that is, why use the `<sid>` mappings? I'm sorry to say I don't remember, but I think it was necessary for some reason.)

autoload/vimtex/motion.vim
```diff
@@ -194,6 +216,46 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
 endfunction
 
 " }}}1
+function! vimtex#motion#math(begin, backwards, visual) abort " {{{1
+
```
lervag, line 220: Sorry for being so pedantic: Please don't have an empty line as the first line of a function.

autoload/vimtex/motion.vim
```diff
@@ -194,6 +216,46 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
...
+    let l:iter = 0
+    let l:success = 0
+    while l:iter <= 5
+      let l:iter += 1
+      call search(l:re, l:flags)
+      let l:pos = vimtex#pos#get_cursor()
+      if a:begin == 0
+        let l:pos =vimtex#pos#prev(vimtex#pos#prev(l:pos))
+      endif
+      if vimtex#syntax#in_mathzone(l:pos[1],l:pos[2])
```
lervag, line 247: Minor: Please add a space after the comma.

lervag:
> @lervag I have hit an issue with the following `tex` source ...
>
>     * I think the problem is with `synstack` call in function `vimtex#syntax#stack`. When a call is made to get the syntax items of the position (5,12) (two columns left to the `\)` on (5,14), it returns an empty list, however when I call synstack manually for (5,12) it returns an integer.

I'll look into it!

I've also made a couple of very minor, pedantic comments (sorry!).

lervag:
If the current "match" is any of `\]` and `\)` (or similar for beginnings), then we don't need to check the syntax. The syntax check is only necessary for `$` and `\end`. This should fix the issue you mention, or, at least the exact issue. The same problem may still be present, but this is exactly why I think it is necessary/useful to add the tests.

autoload/vimtex/motion.vim
lervag: It seems you noticed the minor error but missed the major one: you must change `<plug>...` to `<sid>...` in the `xnoremap` lines.

oponkork: Fixed. Sorry about that. I only know basics mappings, so while copying mappings from the `environment` section, I made the mistake.

oponkork:
> If the current "match" is any of `\]` and `\)` (or similar for beginnings), then we don't need to check the syntax. The syntax check is only necessary for `$` and `\end`. This should fix the issue you mention, or, at least the exact issue.
 The same problem may still be present, but this is exactly why I think it is necessary/useful to add the tests.

Yes, I was thinking the same. While writing the tests I realized that `\]`, `\)` and `\end` are easy to handle, '$' and '$$' create a lot of corner cases, which
arise mainly because we need to check previous positions for mathzones while trying to find the end of a mathzone. Consider the following text for example (line numbers not part of the text)
```tex
1 $\alpha + \beta$
2 $\gamma$
```
If the cursor is on the last column of line 2 (right delimiter), then `[N` should take us to the last column on line 1 (right delimiter). But what happens is that the search takes us to the first column of line 2 (left delimiter), for which the previous-previous position is (1,15) and it is in mathzone. The function then sends us incorrectly to line 2 column 1. In this case then we need to check first if the current search position is an opening delimiter or not.

Another example
```tex
1 $\alpha + \beta$
2
3 $$\gamma$$
```
In this case if the cursor is on the last column of line 3 (right delimiter), the search takes us one column left because `$$` itself matches, to handle this case then we also need to check that we have atleast moved more than one column left or otherwise continue searching.

I have pushed a commit, where I am handling the edge cases by using regex subgroups and a branching logic, and I think I have done it in a crude way.

If there's a better way of handling these cases, or if there are other edge cases you can think of, please let me know.

oponkork:
@lervag I have added tests as you suggested. When I check them manually, they run fine but when I run `make test-math-motions`, the jump doesn't happen and the assertion fails. I have checked that my mappings exist while the test runs.

lervag:
> Yes, I was thinking the same. While writing the tests I realized that `\]`, `\)` and `\end` are easy to handle. '$' and '$$' on the other hand create some corner cases, ...

Yes, I expected these. These have been head scratchers in many other cases as well. I'm glad to see you've figured out a good way to solve it!

> I have pushed a commit, where I am handling the edge cases by using regex subgroups and a branching logic, and I think I have done it in a crude way.

I appreciate the use of the `p` flag for `search()`, I've never used that before my self! Neat! I don't immediately see any important ways to improve this, except one: The motion does not properly presever the last position in the jump list. That is, as it is now, doing e.g. `3]n`, then `<c-o>`, you may notice that it does not return to the original position. This can be easily fixed by adding ``normal! m``` in the beginning of the function, e.g. after the `if a:visual` block.

> @lervag I have added tests as you suggested. When I check them manually, they run fine but when I run `make test-math-motions`, the jump doesn't happen and the assertion fails. I have checked that my mappings exist while the test runs.

The solution is obvious when you see it: this feature requires `syntax on`. So, simply add `syntax on` after `filetype plugin on` in the test file. :)

This leaves two minor fixes as mentioned above, and then finally the documentation:

* Add motion to feature list in README.md.
* Add motion to the corresponding list in `doc/vimtex.txt`.
* Add the mappings to the default map list and to the "map definitions" section (just search for `]m` to see how it is documented and use the same format).

oponkork:
> I appreciate the use of the `p` flag for `search()`, I've never used that before my self! Neat!

Thanks a lot.

> I don't immediately see any important ways to improve this, except one: The motion does not properly presever the last position in the jump list. That is, as it is now, doing e.g. `3]n`, then `<c-o>`, you may notice that it does not return to the original position. This can be easily fixed by adding ``normal! m``` in the beginning of the function, e.g. after the `if a:visual` block.

Fixed.

> The solution is obvious when you see it: this feature requires `syntax on`. So, simply add `syntax on` after `filetype plugin on` in the test file. :)

Aah, I should have realized that. Fixed.

- [x] Add motion to feature list in README.md.
- [x] Add motion to the corresponding list in `doc/vimtex.txt`.
- [x] Add the mappings to the default map list and to the "map definitions" section (just search for `]m` to see how it is documented and use the same format).

Please let me know if you want me to make anymore changes.

I am doing some fine tuning, will let you know once I think it is ready to be  merged.

:)

autoload/vimtex/motion.vim
```diff
@@ -194,6 +216,82 @@ function! vimtex#motion#comment(begin, backwards, visual) abort " {{{1
 endfunction
 
 " }}}1
+function! vimtex#motion#math(begin, backwards, visual) abort " {{{1
+  let l:curpos_saved = vimtex#pos#get_cursor()
+  let l:count = v:count1
+  if a:visual
+    normal! gv
+  endif
+
```
lervag, line 225: Add ``normal! m` `` above line 225.

lervag:
> > I don't immediately see any important ways to improve this, except one: The motion does not properly presever the last position in the jump list. That is, as it is now, doing e.g. `3]n`, then `<c-o>`, you may notice that it does not return to the original position. This can be easily fixed by adding ``normal! m``` in the beginning of the function, e.g. after the `if a:visual` block.
> 
> Fixed.

Did you fix it in a different manner, perhaps?

> Please let me know if you want me to make anymore changes.

No, this looks good. Thanks!

> I am doing some fine tuning, will let you know once I think it is ready to be merged.

Sounds good.

autoload/vimtex/motion.vim
oponkork: Sorry forgot to push.

oponkork:
@lervag I have fixed a regex bug that I had introduced while changing the implementation.

I have also changed the behavior of the motion when the `count` is larger than the total available math zones. Earlier, the motion didn't occur unless we find the `count` number of math zones. I have modified it to jump to the last available math zone, which is the same as the behavior of `]m` and similar motions. Also, the for loop exists the first time the while loop fails, to prevent unexpected jumps when the count is larger than one.

I have checked the motions on some of my latex files as well, and I think it is ready to be merged now. Please let me know what you think of it.

lervag:
> @lervag I have fixed a regex bug that I had introduced while changing the implementation.
> 
> I have also changed the behavior of the motion when the `count` is larger than the total available math zones. Earlier, the motion didn't occur unless we find the `count` number of math zones. I have modified it to jump to the last available math zone, which is the same as the behavior of `]m` and similar motions. Also, the for loop exists the first time the while loop fails, to prevent unexpected jumps when the count is larger than one.
> 
> I have checked the motions on some of my latex files as well, and I think it is ready to be merged now. Please let me know what you think of it.

Thanks! I've now merged this. It seems to work well also in my tests, and at least it should not break anything.

I've made some minor modifications/changes, nothing essential. Again, thank you very much for this, it seems to be a useful feature and your work is of high quality! Very well done!

oponkork:
> Thanks! I've now merged this. It seems to work well also in my tests, and at least it should not break anything.
> 
> I've made some minor modifications/changes, nothing essential. Again, thank you very much for this, it seems to be a useful feature and your work is of high quality! Very well done!

Your modifications are indeed improvements on what I wrote. Thanks for helping me in adding the feature. :+1:

lervag:
My pleasure :)

PR #1821, opened 2020-10-07
Compiler vlty: support LanguageTool as installed by pacman

matze-dd:
This PR addresses an issue raised in the [discussion of vlty](https://github.com/lervag/vimtex/pull/1714#issuecomment-650387467). Under Arch Linux, one now can use LanguageTool as installed with
```
pacman -S languagetool
```
The minimal vimrc is given in the documentation.
This requires YaLafi version 1.1.5. It has been uploaded to PyPI, such that it is available via pip.

### DIFF
--- a/compiler/vlty.vim
+++ b/compiler/vlty.vim
@@ -40,12 +40,10 @@ if s:vlty.server !=# 'lt'
       call s:installation_error('vlty compiler - lt_command not valid')
       finish
     endif
-  else
-    if !filereadable(fnamemodify(s:vlty.lt_directory
+  elseif !filereadable(fnamemodify(s:vlty.lt_directory
         \ . '/languagetool-commandline.jar', ':p'))
-      call s:installation_error('vlty compiler - lt_directory path not valid')
-      finish
-    endif
+    call s:installation_error('vlty compiler - lt_directory path not valid')
+    finish
   endif
 endif
 
@@ -57,9 +55,8 @@ let s:language = substitute(s:language, '_', '-', '')
 
 let &l:makeprg =
       \ s:python . ' -m yalafi.shell'
-      \ . ' --lt-command "' . s:vlty.lt_command . '"'
       \ . (s:vlty.lt_command != ''
-      \    ? ''
+      \    ? ' --lt-command ' . s:vlty.lt_command
       \    : ' --lt-directory ' . s:vlty.lt_directory)
       \ . (s:vlty.server ==# 'no'
       \    ? ''
--- a/compiler/vlty.vim
+++ b/compiler/vlty.vim
@@ -17,8 +17,7 @@ if !executable(s:python)
   finish
 endif
 
-call system(s:python . ' -c "import sys; assert sys.version_info.major >= 3'
-            \ . ' and sys.version_info.minor >= 6"')
+call system(s:python . ' -c "import sys; assert sys.version_info >= (3, 6)"')
 if v:shell_error != 0
   call s:installation_error('vlty compiler requires at least Python version 3.6')
   finish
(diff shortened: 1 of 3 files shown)


compiler/vlty.vim
```diff
@@ -51,9 +57,12 @@ let s:language = substitute(s:language, '_', '-', '')
 
 let &l:makeprg =
       \ s:python . ' -m yalafi.shell'
-      \ . ' --lt-directory ' . s:vlty.lt_directory
+      \ . ' --lt-command "' . s:vlty.lt_command . '"'
```
lervag, line 60: Does it make sense to _always_ supply the `--lt-command` option, or should this only be supplied if `lt_command` is available?

doc/vimtex.txt
```diff
@@ -1616,6 +1616,8 @@ OPTIONS                                                        *vimtex-options*
 
     lt_directory~
       Path to the `LanguageTool` software.
+    lt_command~
```
lervag, line 1619: I like it better if there's an empty line between the descriptions. Also, I'm curious if `lt_command` perhaps should be the first entry here?

doc/vimtex.txt
```diff
@@ -4072,8 +4079,15 @@ As a minimal example, one could write in |vimrc|: >
   set spelllang=en_gb
 
 The given directory has to contain the `LanguageTool` software, including for
-instance the file `languagetool-server.jar`. A separated `:compiler vlty` will
-raise an error message, if some component cannot be found.
+instance the file `languagetool-server.jar`.
+If `LanguageTool` has been installed via a packet manager as given above,
```
lervag, line 4083: I prefer the paragraphs to be full "blocks" of text, so move `If ...` back to the previous line. If you meant to start a new paragraph, then add an empty line.

compiler/vlty.vim
```diff
@@ -36,10 +35,17 @@ if s:vlty.server !=# 'lt'
...
   endif
 
-  if !filereadable(fnamemodify(s:vlty.lt_directory
+  if s:vlty.lt_command != ''
+    if !executable(s:vlty.lt_command)
+      call s:installation_error('vlty compiler - lt_command not valid')
+      finish
+    endif
+  else
+    if !filereadable(fnamemodify(s:vlty.lt_directory
```
lervag, line 44: I think it is better to use `elseif` here.

lervag:
> This PR addresses an issue raised in the [discussion of vlty](https://github.com/lervag/vimtex/pull/1714#issuecomment-650387467). Under Arch Linux, one now can use LanguageTool as installed with
> 
> ```
> pacman -S languagetool
> ```
> 
> The minimal vimrc is given in the documentation.
> This requires YaLafi version 1.1.5. It has been uploaded to PyPI, such that it is available via pip.

Great, thanks! I appreciate the update, I only had a few minor comments!

compiler/vlty.vim
matze-dd: Yes, this is just the version with the least key strokes. I will change that.

doc/vimtex.txt
matze-dd: > I like it better if there's an empty line between the descriptions.

OK, I will insert a blank line.

> Also, I'm curious if lt_command perhaps should be the first entry here?

Don't know, which key is more important. But I assume that most people will install LanguageTool "manually".

doc/vimtex.txt
matze-dd: OK.

compiler/vlty.vim
matze-dd: My first version was this way, and it is in fact synonymous here. But then, I thought the current code would be more readable. I will change that back.

matze-dd:
> Great, thanks! I appreciate the update, I only had a few minor comments!

My thanks go back for your great project :)

I will incorporate the changes and make a new comment afterwards.

matze-dd:
The changes are included. Only for the order of lt\_directory and lt\_command, things are kept. I think, most people will use the manual LanguageTool installation, although this is only a guess.

lervag:
Thanks! I merging this now.

lervag:
> The changes are included. Only for the order of lt_directory and lt_command, things are kept. I think, most people will use the manual LanguageTool installation, although this is only a guess.

Seems sensible. Again, thanks! :)

matze-dd:
Thank you for merging and the final adjustments! These are all improvements.

lervag:
My pleasure :)

PR #1830, opened 2020-10-15
Enhance fold labels on folded beamer frames

bk90:
Hello,

this PR enhances (IMHO) the labels on folded frames in LaTeX beamer documents.

1. The first commit changes the regex parsing comments on the `\begin{frame}` line. So far the comments where only found if they were separated from the `\begin{frame}` only by whitespace. Now optional `[...]` are allowed, as in `\begin{frame}[noframenumbering]  % Title page`.
2. The second commit extends the search for `\frametitle()` by looking for `\framesubtitle()` on the following line and, if both are found, displaying them on the fold as `frametitle: framesubtitle`.
3. The third commit to me is optional due to commit (1). It changes the regex looking for `frametitle`, so that commented out `frametitle`s are found as well. This originally was a workaround for the first commit, but still could be useful.

### DIFF
--- a/autoload/vimtex/fold/envs.vim
+++ b/autoload/vimtex/fold/envs.vim
@@ -171,8 +171,14 @@ function! s:folder.parse_caption_frame(line) abort dict " {{{1
   let i = v:foldstart
   while i <= v:foldend
     if getline(i) =~# '^\s*\\frametitle'
-      return matchstr(getline(i),
+      let frametitle = matchstr(getline(i),
             \ '^\s*\\frametitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
+      if i+1 <= v:foldend && getline(i+1) =~# '^\s*\\framesubtitle'
+        let framesubtitle = matchstr(getline(i+1),
+              \ '^\s*\\framesubtitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
+        return printf('%S: %S', frametitle, framesubtitle)
+      end
+      return frametitle
     end
     let i += 1
   endwhile
--- a/autoload/vimtex/fold/envs.vim
+++ b/autoload/vimtex/fold/envs.vim
@@ -178,7 +178,7 @@ function! s:folder.parse_caption_frame(line) abort dict " {{{1
   endwhile
 
   " If no caption found, check for a caption comment
-  return matchstr(a:line,'\\begin\*\?{.*}\s*%\s*\zs.*')
+  return matchstr(a:line,'\\begin\*\?{.*}\(\[.*\]\)\?\s*%\s*\zs.*')
 endfunction
 
 " }}}1
--- a/test/test-folding/main.tex
+++ b/test/test-folding/main.tex
@@ -201,4 +201,17 @@ this fold worked before issue #1515
     \end{proof}
 \end{enumerate}
 
+% The frames need to be in this order for the test to pass. This is likely a
+% bug in vim, see https://github.com/lervag/vimtex/pull/1830#issuecomment-775527621
+\begin{frame}
+  \frametitle{Title}
+  \framesubtitle{Subtitle}
+
+  hello world
+\end{frame}
+
+\begin{frame}[noframenumbering]  % Title page
+  hello world
+\end{frame}
+
 \end{document}
--- a/test/test-folding/test.vim
+++ b/test/test-folding/test.vim
@@ -56,6 +56,15 @@ call vimtex#test#assert_equal(3, foldlevel(186))
 call vimtex#test#assert_equal(1, foldlevel(190))
 call vimtex#test#assert_equal(2, foldlevel(202))
 
+call vimtex#test#assert_equal(2, foldlevel(206))
+call vimtex#test#assert_equal(
+      \ '\begin{frame}       Title: Subtitle',
+      \ foldtextresult(206))
+call vimtex#test#assert_equal(2, foldlevel(213))
+call vimtex#test#assert_equal(
+      \ '\begin{frame}       Title page',
+      \ foldtextresult(213))
+
 call vimtex#test#assert_equal(1, foldlevel(line('$')-1))
 
 quit!


lervag:
Thanks. Could you provide a couple of complete examples for the various cases?

bk90:
Sure. Given this code

```
\begin{frame}[noframenumbering]  % Title page                                    
  \titlepage                                                                     
\end{frame}

\begin{frame}
  \frametitle{Title}
  \framesubtitle{Subtitle}
\end{frame}

\begin{frame}
  % \frametitle{Title}
\end{frame}
```

I get this folding:

```
\begin{frame}       Title page·····················································

\begin{frame}       Title: Subtitle·················································

\begin{frame}       Title··························································
```

Without the changes the folding is

```
\begin{frame}       ·······························································

\begin{frame}       Title···························································

\begin{frame}       ·······························································
```

The first example changes, because the regex now allows for `[.*]` in between `\begin{frame}` and the comment.

The second example changes, because the code now searches and extracts a `\framesubtitle{Subtitle}` in the line following `\framtitle{Title}`. This could be improved by searching not only the following line but all following lines. Do you think that's necessary?

The third example changes, because commented out `\frametitle{Title}` commands are parsed as well. I don't know if this is desired, what do you think?

lervag:
> Sure. Given this code ...

Thanks for the good examples and before vs after explanations.

> The second example changes, because the code now searches and extracts a `\framesubtitle{Subtitle}` in the line following `\framtitle{Title}`. This could be improved by searching not only the following line but all following lines. Do you think that's necessary?

It seems possible, but I think this should already suffice for a useful improvement. If you want to look into even more advanced parsing, I'll be glad to consider an updated PR or a future PR for that!

> The third example changes, because commented out `\frametitle{Title}` commands are parsed as well. I don't know if this is desired, what do you think?

I'm not sure. In a way, I think we should ignore the commented out title. It does not feel light the right way to control the fold text. Do you have a strong opinion about this?

autoload/vimtex/fold/envs.vim
```diff
@@ -170,15 +170,21 @@ function! s:folder.parse_caption_frame(line) abort dict " {{{1
...
-    if getline(i) =~# '^\s*\\frametitle'
-      return matchstr(getline(i),
-            \ '^\s*\\frametitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
+    if getline(i) =~# '^.*\\frametitle'
+      let frametitle = matchstr(getline(i),
+            \ '^.*\\frametitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
+      if i+1 <= v:foldend && getline(i+1) =~# '^.*\\framesubtitle'
+        let framesubtitle = matchstr(getline(i+1),
+              \ '^.*\\framesubtitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
+        return printf('%S: %S', frametitle, framesubtitle)
```
lervag, line 179: I would simplify this to:

```vim
let frametitle .= ': ' . matchstr(getline(i+1),
      \ '^.*\\framesubtitle\(\[.*\]\)\?{\zs.\{-1,}\ze\(}\s*\)\?$')
```

autoload/vimtex/fold/envs.vim
```diff
@@ -170,15 +170,21 @@ function! s:folder.parse_caption_frame(line) abort dict " {{{1
...
+      end
+      return frametitle
     end
     let i += 1
   endwhile
 
   " If no caption found, check for a caption comment
-  return matchstr(a:line,'\\begin\*\?{.*}\s*%\s*\zs.*')
+  return matchstr(a:line,'\\begin\*\?{.*}\(\[.*\]\)\?\s*%\s*\zs.*')
```
lervag, line 187: I would change `\(\[.*\]\)\?` to `\(\[[^\]]*\]\)\?`.

lervag (commented):
I've looked through the code, and I think it is decent. I have two comments, and we'll have to decide about the commented out frametitles. I also want to add tests, but I can do that myself. If you want to try, look into `test/tests/test-folding` for examples of how to do it.

autoload/vimtex/fold/envs.vim
bk90: I tried this and as far as I can tell this adds `: 0` to the fold text if no subtitle is available, so the `if` condition seems necessary. But I could remove the extra variable and `return` statement, if you prefer.

autoload/vimtex/fold/envs.vim
bk90: I can't get it to work with the suggestion. What would be the improvement? I copied the regex for detecting the brackets from the `\frametitle` regex a few lines above this.

bk90:
> > The third example changes, because commented out `\frametitle{Title}` commands are parsed as well. I don't know if this is desired, what do you think?
> 
> I'm not sure. In a way, I think we should ignore the commented out title. It does not feel light the right way to control the fold text. Do you have a strong opinion about this?

I removed the code for the commented frame titles. I guess if someone comments the frame title its for a good reason, so we should respect this and not show it on the fold. If the title should be shown it can be added as a comment to the`\begin{frame}` line.

> I also want to add tests, but I can do that myself. If you want to try, look into test/tests/test-folding for examples of how to do it.

I added a test, but have some trouble with it. The way it's currently set up the test breaks on recognizing the title comment on line 204. If I switch the order of the two added beamer frames and update the test accordingly (line 204 -> 211, 208 -> 204) the test works. Do you have any idea why? As far as I can tell the order of the frames should not have any effect. When I open the latex file in vim the fold texts are shown as expected either way.

lervag:
> I removed the code for the commented frame titles. I guess if someone comments the frame title its for a good reason, so we should respect this and not show it on the fold. If the title should be shown it can be added as a comment to the`\begin{frame}` line.

Great!

> > I also want to add tests, but I can do that myself. If you want to try, look into test/tests/test-folding for examples of how to do it.
> 
> I added a test, but have some trouble with it. The way it's currently set up the test breaks on recognizing the title comment on line 204. If I switch the order of the two added beamer frames and update the test accordingly (line 204 -> 211, 208 -> 204) the test works. Do you have any idea why? As far as I can tell the order of the frames should not have any effect. When I open the latex file in vim the fold texts are shown as expected either way.

Huh, this is very strange. For some reason, the `v:foldend` and `v:foldlevel` variables are not correct when we call `foldtextresult(xxx)`. I looked into this a bit now and could not find anything wrong with your code, and I suspect it might actually be a bug in Vim. Not fully sure, though. The behaviour is consistent between Vim and neovim.

But I'll accept this regardless; let's just swap the tests to make them pass and add a comment (e.g. refer to the issue tread).

Nice work, thanks! If you are happy with this, then I can probably merge this in a day or two.

bk90:
The test is updated and I think the PR is ready to be merged. Thank you for the great plugin, it has made my life easier for several years now. I'm glad I could make a small contribution :)

lervag:
> The test is updated and I think the PR is ready to be merged.

Merging now.

> Thank you for the great plugin, it has made my life easier for several years now. I'm glad I could make a small contribution :)

I'm happy to hear it is useful to you! And thanks for your contribution, it is appreciated!

Issue #1829, opened 2020-10-14
Respect file encoding with vlty gramma check


petRUShka:
**Issue**
Suppose we have tex-file in non utf-8 encoding and want to check grammar. vimtex together with vlty don't respect fileencoding so I get an error:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 340: invalid start byte
```

**minimal.vim**
  ```vim
  set nocompatible
  let &runtimepath  = '~/.vim/bundle/vimtex,' . &runtimepath
  let &runtimepath .= ',~/.vim/bundle/vimtex/after'
  filetype plugin indent on
  syntax enable

  let g:vimtex_grammar_vlty = {}
  let g:vimtex_grammar_vlty.lt_directory = '/usr/share/java/languagetool/'
  let g:vimtex_grammar_vlty.lt_command = 'languagetool'
  let g:vimtex_grammar_vlty.encoding = 'cp866'

  if has('nvim')
      let g:vimtex_compiler_progname = 'nvr'
  endif

  au Filetype tex
              \ set spell |
              \ set spelllang=ru,en
  ```

**minimal.tex**
  ```tex
    \documentclass{article}
    \usepackage[cp866]{inputenc}
    \usepackage[russian]{babel}

    \begin{document}
    \selectlanguage{russian}

    Случайный текст на русском

    \end{document}
  ```

  don't forget to convert file to cp866 encoding

```
iconv -t cp866 test.tex > test_cp866.tex
```
**Commands/Input**
Setup vim and `text_cp866.tex` like above. Run vim on test_cp866.tex and execute:

```
:compiler vlty
:make
```

**Observed Behaviour**

```
:!python3 -m yalafi.shell --lt-directory /usr/share/java/languagetool/ --language ru --disable "WHITESPACE_RULE" --enable "" --disablecategories "" --enablecat
egories "" --documentclass "" --packages "amsthm,babel,enumerate,inputenc"  'test_cp866.tex'  2>&1| tee /tmp/vijaI15/129
=== test.tex
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.8/site-packages/yalafi/shell/__main__.py", line 3, in <module>
    from . import shell
[...]

PR #1839, opened 2020-10-21
Compiler vlty: respect Vim variable fileencoding

matze-dd:
This relates to issue #1829.

- vlty.vim: According to the doc, `fileencoding` may be empty. Then `encoding` is used.
- options.vim: added default for encoding
- vimtex.txt:
  - As in the previous PR, I'm undecided about the order of the options. Please feel free to change that ;-)
  - I have wrapped some lines that were longer than 80 characters.

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -137,6 +137,7 @@ function! vimtex#options#init() abort " {{{1
         \ 'server': 'no',
         \ 'shell_options': '',
         \ 'show_suggestions': 0,
+        \ 'encoding': 'auto',
         \})
 
   call s:init_option('vimtex_imaps_enabled', 1)
--- a/compiler/vlty.vim
+++ b/compiler/vlty.vim
@@ -61,6 +61,9 @@ let &l:makeprg =
       \ . (s:vlty.server ==# 'no'
       \    ? ''
       \    : ' --server ' . s:vlty.server)
+      \ . ' --encoding ' . (s:vlty.encoding ==# 'auto'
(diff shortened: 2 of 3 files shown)


lervag:
Thanks! I think this is good enough for an immediate merge (I just had to fix a minor typo).

> As in the previous PR, I'm undecided about the order of the options. Please feel free to change that ;-)

The option description comes in a consistent order compared to the default dict, which is nice. In many cases, the order is irrelevant, and I often try to use e.g. alphabetic order. However, in some cases it feels right to put the important things on top. This does result in some things being in semi random order. I don't think it's a problem, but if anyone does I'm glad to discuss it.

> I have wrapped some lines that were longer than 80 characters.

I've often tried to keep things wrapped to 79 or 80 characters on the _visual_ columns. E.g., in lines that have a lot of concealed characters, I tend to allow more than 80 characters. But then I guess there's some/many who don't have conceals enabled at all, so perhaps that's a bad idea.

matze-dd:
Thanks for merging!

> I've often tried to keep things wrapped to 79 or 80 characters on the visual columns. E.g., in lines that have a lot of concealed characters, I tend to allow more than 80 characters. But then I guess there's some/many who don't have conceals enabled at all, so perhaps that's a bad idea.

Ah, I see. For instance, there was a line with 81 characters:
```
  This option is used to configure the `vlty` grammar checker. It is a dictionary 
```
With my simple configuration, in `:help`, the backticks around `vlty` are displayed, so that the line is wrapped in an 80-column terminal.

lervag:
Yes, that makes sense. I'll revert my silly practice of considering the visual columns in future updates to the docs. :)

Issue #1854, opened 2020-11-14
Compatibility of compiler plugins with vim-dispatch


dianlujitao:
**Issue**
To perform grammar check with YaLafi, the documented way is:

```vim
:compiler vlty
:make
:copen
```

This works quite well and properly fills the quickfix window like this:

![image](https://user-images.githubusercontent.com/5774651/99146478-80647200-26b3-11eb-9585-eb475d7faf6f.png)

But with Tim Pope's vim-dispatch plugin, after either:

```vim
:compiler vlty
:Make
```

or `:Dispatch -compiler=vlty`, I got additional boilerplate outputs as follows:

![image](https://user-images.githubusercontent.com/5774651/99146528-f8cb3300-26b3-11eb-8269-267e7bf17da1.png)

I failed to find the explanation from neither vimtex's nor vim-dispatch's doc, how to make vim-dispatch produce the same output as vim's built-in `:make` command?

Cc: @matze-dd

lervag:
Interesting and good question; I hope @matze-dd can look into it and help. I don't really know so much about this stuff myself.

matze-dd:
Thanks for reporting ...

> This works quite well and properly fills the quickfix window like this:

... and for the positive feedback!

The problem you describe is very probably due to a non-correct `errorformat` setting in compiler vlty. I will need some more time to propose a PR, but this quick fix did work for me in `.../vimtex/compiler/vlty.vim` (the commented lines starting with `"` are the ones to be replaced by the following ones).
```
if s:vlty.show_suggestions == 0
" let &l:errorformat .= '%ZMessage: %m,%-G%.%#'
  let &l:errorformat .= '%ZMessage: %m,%-G%.%#,%-G%.%#'
else
" let &l:errorformat .= '%CMessage: %m,%Z%m,%-G%.%#'
  let &l:errorformat .= '%CMessage: %m,%Z%m,%-G%.%#,%-G%.%#'
endif
```
It seems that the current `errorformat` does not consume enough of the message returned by YaLafi, so that vim-dispatch tries to assign it somehow.

BTW: It appears that you do not want to wait while the proofreader is running. Did you already try this option?
```
let g:vimtex_grammar_vlty.server = 'my'
```
The started LanguageTool server is faster thanks to the omitted start-up work at each check, and thanks to its internal sentence caching mechanism.

dianlujitao:
> ```
> if s:vlty.show_suggestions == 0
[...]

PR #1856, opened 2020-11-17
Compiler vlty: ensure compatibility with vim-dispatch

matze-dd:
This relates to issue #1854. An [issue from vim-dispatch](https://github.com/tpope/vim-dispatch/issues/88) loosely relates to ours.

The proposed solution should be less dubious than the quick fix from #1854. The latter just said *twice:* "Please ignore all other lines." No clue, why this works as intended.

In the new `errorformat`, all message lines from YaLafi are explicitly consumed.

The part `%+CSuggestion:%m` is a bit special. If one simply says `%C%m` for picking up LT's replacement suggestions, then this will add the complete remainder of the multi-line message.


matze-dd:
Updated after discussion in #1854.

lervag:
Thanks, I've merged this now.

PR #1860, opened 2020-11-23
No autocmd when opening QuickFix window

### LINKED ISSUE #1595: (Neo)Vim 'w' behaves like 'W' in a certain scenario
**Issue**
Vim word movements like 'w', 'e', 'b', etc. suddenly behave like their capital counterparts 'W', 'E', 'B', etc., when a quickfix list pops out and if the cursor is on the end-of-a-line-`}`.
This issue has been nagging me for a few years, and I finally had a time to investigate it.

`iskeyword` seems to be fine:
```vim
 iskeyword=@,48-57,_,192-255,:
```

This issue is always reproducible with the following `minimal.vim` & `minimal.tex` on my Macs.

**minimal.vim**
  ```vim
let &runtimepath  = '~/.local/share/nvim/plugged/vimtex,' . &runtimepath
let &runtimepath .= ',~/.local/share/nvim/plugged/vimtex/after'
  ```

**minimal.tex**
  ```tex
% Below 'articl' is an intended typo
\documentclass{articl}
\begin{document}
\TeX
\end{document}
  ```

**Commands/Input**
Put the cursor on the closing curly bracket `}` in `{acticl}`.
Then start a continuous compile mode via `\ll`.

**Observed Behaviour**
The missing 'e' opens the quickfix list, and if you try to move using `w`, it would behave like `W`.

This also breaks a 3rd party plugin like [easymotion](https://github.com/easymotion/vim-easymotion), only showing the starts of *W*ords (delimited by spaces).

**Expected Behaviour**
It should still act like a normal `w`(, `e`, `b`, etc.).

**Output from VimtexInfo**
```
System info
  OS: Mac OS X 10.15.3 (19D76)
  Vim version: NVIM v0.4.3
  Has clientserver: true
  Servername: /var/folders/cs/7bfzsfzx7bn67bhxn0pgzk1w0000gn/T/nvim72gwdA/0

vimtex project: minimal
  base: minimal.tex
  root: /Users/jay/vimtex
  tex: /Users/jay/vimtex/minimal.tex
  out: /Users/jay/vimtex/minimal.pdf
  log: /Users/jay/vimtex/minimal.log
  aux: /Users/jay/vimtex/minimal.aux
  fls: /Users/jay/vimtex/minimal.fls
  compiler: latexmk
    backend: nvim
    output: /var/folders/cs/7bfzsfzx7bn67bhxn0pgzk1w0000gn/T/nvim72gwdA/1
    configuration: 
      continuous: 1
      callback: 1
      latexmk options:
        -verbose
        -file-line-error
        -synctex=1
        -interaction=nonstopmode
      latexmk engine: -pdf
  viewer: General
  qf: LaTeX logfile
    config: 
      packages: 
        default: 1
      default: 1
  document class: article
```

**Output from chechhealth**
```markdown
health#vimtex#check
========================================================================
## vimtex
  - OK: Vim version should have full support!
  - OK: General viewer should work properly!
  - WARNING: pstree is not available
    - ADVICE:
      - vimtex#view#reverse_goto is better if pstree is available.
  - WARNING: Compiler callbacks will not work!
    - ADVICE:
      - `neovim-remote` / `nvr` is required for callbacks to work with neovim
      - Please also set |g:vimtex_compiler_progname| = 'nvr'
      - For more info, see :help |vimtex-faq-neovim|

health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=^H
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $TERM_PROGRAM='iTerm.app'
  - INFO: $COLORTERM='truecolor'

## tmux
  - OK: escape-time: 0ms
  - INFO: $TERM: tmux-256color

health#provider#check
========================================================================
## Clipboard (optional)
  - OK: Clipboard tool found: pbcopy

## Python 2 provider (optional)
  - INFO: pyenv: Path: /usr/local/Cellar/pyenv/1.2.16/libexec/pyenv
  - INFO: pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.
  - INFO: pyenv: Root: /Users/jay/.pyenv
  - WARNING: No Python executable found that can `import neovim`. Using the first available executable for diagnostics.
  - ERROR: Python provider error:
    - ADVICE:
      - provider/pythonx: Could not load Python 2:
          /Users/jay/.pyenv/shims/python2 does not exist: pyenv: python2: command not found
          
          The `python2' command exists in these Python versions:
          2.7.17
          2.7.17/envs/neovim-python2
          neovim-python2
          
          
          /Users/jay/.pyenv/shims/python2.7 does not exist: pyenv: python2.7: command not found
          
          The `python2.7' command exists in these Python versions:
          2.7.17
          2.7.17/envs/neovim-python2
          neovim-python2
          
          
          python2.6 not found in search path or not executable.
          /Users/jay/.pyenv/shims/python is Python 3.8 and cannot provide Python 2.
  - INFO: Executable: Not found

## Python 3 provider (optional)
  - INFO: pyenv: Path: /usr/local/Cellar/pyenv/1.2.16/libexec/pyenv
  - INFO: pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.
  - INFO: pyenv: Root: /Users/jay/.pyenv
  - WARNING: No Python executable found that can `import neovim`. Using the first available executable for diagnostics.
  - ERROR: Python provider error:
    - ADVICE:
      - provider/pythonx: Could not load Python 3:
          /Users/jay/.pyenv/shims/python3 does not have the "neovim" module. :help |provider-python|
          /usr/local/bin/python3.7 does not have the "neovim" module. :help |provider-python|
          python3.6 not found in search path or not executable.
          python3.5 not found in search path or not executable.
          python3.4 not found in search path or not executable.
          python3.3 not found in search path or not executable.
          /Users/jay/.pyenv/shims/python does not have the "neovim" module. :help |provider-python|
  - INFO: Executable: Not found

## Ruby provider (optional)
  - INFO: Ruby: ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin19]
  - INFO: Host: /Users/jay/.rbenv/shims/neovim-ruby-host
  - OK: Latest "neovim" gem is installed: 0.8.1

## Node.js provider (optional)
  - INFO: Node.js: v13.7.0
  - INFO: Neovim node.js host: /usr/local/lib/node_modules/neovim/bin/cli.js
  - OK: Latest "neovim" npm/yarn package is installed: 4.8.0
```

andryandrew:
Disables firing of autocommands when opening the QuickFix window.

Fixes #1595. While the bug is ultimately `neovim`'s, firing a `CursorMoved` more often than in `vim` is a `neovim` design choice.

### DIFF
--- a/autoload/vimtex/qf.vim
+++ b/autoload/vimtex/qf.vim
@@ -81,7 +81,7 @@ function! vimtex#qf#open(force) abort " {{{1
 
   if a:force || (g:vimtex_quickfix_mode > 0 && l:errors_or_warnings)
     let s:previous_window = win_getid()
-    botright cwindow
+    noautocmd botright cwindow
     if g:vimtex_quickfix_mode == 2
       call win_gotoid(s:previous_window)
     endif


lervag:
I'm curious if this would prevent running "important" or "desired" autocmds. E.g., I personally have

```vim
augroup quickfix_autocmds
  autocmd!
  autocmd BufReadPost quickfix call personal#qf#adjust_height()
augroup END
```

where `adjust_height` is used to adjust the height of the window. Would not this PR disable this particular autocmd?

lervag:
My simple test for this confirms my suspicion. Thus, I would prefer to not merge this and instead implement a different fix, if possible.

andryandrew:
Oups, sorry :sweat_smile: I did not think this through enough.
We can just ignore the two culprits instead of all autocommands (with `eventignore=CursorMoved,WinEnter`)...

...but maybe it's better to go with a "redraw" just before the `win_gotoid` call? This too seems to fix the important part of the bug (the misbehaving motion keys), but leaves the "ghost" highlighting in the quickfix window.

Just a question though: why is the built-in `matchparen` still active alongside `vimtex`'s own mechanism? Disbling it (or just loading it after `vimtex`'s one) also fixes the bug.

lervag:
> Oups, sorry sweat_smile I did not think this through enough.
> We can just ignore the two culprits instead of all autocommands (with `eventignore=CursorMoved,WinEnter`)...

No problem! And if possible it seems like a good idea to ignore specific autocommands! I think the main autocommands that are relevant that we don't want to ignore are ones such as the one in my example.

> ...but maybe it's better to go with a "redraw" just before the `win_gotoid` call? This too seems to fix the important part of the bug (the misbehaving motion keys), but leaves the "ghost" highlighting in the quickfix window.

That also sounds like a simple and OK solution. It should be added with a relevant comment (e.g. linking to this PR for motivation).

> Just a question though: why is the built-in `matchparen` still active alongside `vimtex`'s own mechanism? Disabling it (or just loading it after `vimtex`'s one, instead of before) ~also fixes the bug~ **unfortunately, not in all situations**.

Good question. I guess the simple answer is "because Vimtex tries not to interfere with other plugins"? Is there a simple way to disable it that would not lead to unexpected situations for the user (e.g. a user may expected matchparen to be active in some buffers, but not the LaTeX buffers)?

lervag:
Feel free to reopen if you have a proposal for an updated version.

andryandrew:
I can submit a PR with the `redraw` workaround, but apparently I cannot reopen this one (or I don't see how). Is it a permission issue?

lervag:
> I can submit a PR with the `redraw` workaround,

Great, thanks, that would be appreciated!

> but apparently I cannot reopen this one (or I don't see how). Is it a permission issue?

I have no idea. I can't reopen it either. So I guess you should just open a new PR. If/When you do, please link back to this PR and to the original issue, #1595 to help maintain the context.

PR #1862, opened 2020-11-24
fix some cchars

tomtomjhj:
* removed the duplicated command for `\leq`
* more aesthetic `cchar`s for `\langle`, `\rangle` and `\models`


lervag:
Thanks! That's both quick and nice!

PR #1853, opened 2020-11-10
Add motion between frames

RiMillo:
I find the provided motions (by section, environment,...) very handy but I think it could be useful to have one by `frame` to make the editing of `beamer` documents easier.

I took the liberty to write a first version. I just took the environment-related functions and modified them a little bit. It seems to work on simple examples but I did not fully test it and there definitely is room to improvement.

I propose `r`-based shortcuts (so, for instance, `]r` or `[R`): I would have gone with `f`, but it is already taken for instance in [tpope/vim-unimpaired](https://github.com/tpope/vim-unimpaired).

No documentation is given.

Hope it could be useful.

(I'm very sorry if I did not follow the contributions rules, please have mercy of a Git newbie :-) )

PR #1869, opened 2020-11-25
doc for frame motions in README

RiMillo:
As per my last comment in #1853 , I have added the new beamer/frame motions in the README as well.
Sorry for the inconvenience.


lervag:
No inconvenience. Thanks!

Issue #1873, opened 2020-11-30
Highlight group texNewcmdArgName


j1-lee:
**Is your feature request related to a problem? Please describe it.**
Suppose one wants to highlight the LHS (a macro to be defined, e.g., `\norm`) in `\newcommand` command, as follows:
```tex
\newcommand{\norm}[1]{\lVert #1 \rVert}
```
One might try, for example,
```vim
highlight link texNewcmdArgName Yellow
```
and sees no difference. It seems that `\norm` is matched to texCmd. As the LHS almost always is in this form, it means that texNewcmdArgName is almost never accessible; one cannot highlight the LHS separately from a regular texCmd.

In contrast, for `\def`, one can separately highlight LHS in by
```vim
highlight link texDefArgName Yellow
```

**Describe the solution you'd like**
One solution might be adding `'contains': ''` pair in the definition of texNewcmdArgName to force vim _not_ to match texCmd (well, anything) in the highlight region:

autoload/vimtex/syntax/core.vim
```vim
call vimtex#syntax#core#new_arg('texNewcmdArgName', {'next': 'texNewcmdOpt,texNewcmdArgBody', 'contains': ''})
```

If you consider this a relevant feature request and find this solution acceptable, then I can modify the code and issue a PR.

**Additional feature**
It might be good to have similar syntax feature for similar commands such as `\let` and `\DeclareMathOperator`; the former is a TeX-native command, and the latter is an amsmath command (so that it is natural to be defined in .../p/amsmath.vim?). I understand we need to draw the line somewhere, but I believe these two are the most frequently used commands that have similar purposes as `\def` and `\newcommand`.

If you find this additional feature plausible, I'll try working on this.

lervag:
> If you consider this a relevant feature request and find this solution acceptable, then I can modify the code and issue a PR.

Yes, absolutely. I agree with your description. Note, if you issue a PR, then
please also add a default link for `texNewcmdArgName` to `texCmd`.

> **Additional feature**
> It might be good to have similar syntax feature for similar commands such as `\let` and `\DeclareMathOperator`; the former is a TeX-native command, and the latter is an amsmath command (so that it is natural to be defined in .../p/amsmath.vim?). I understand we need to draw the line somewhere, but I believe these two are the most frequently used commands that have similar purposes as `\def` and `\newcommand`.
> 
> If you find this additional feature plausible, I'll try working on this.

[...]

PR #1874, opened 2020-12-01
Syntax support for def-like commands

j1-lee:
This PR attempts to enhance the syntax support for def-like commands: `\let`, `\newcommand`, and `\DeclareMathOperator`, as proposed in #1873.

1. As shown in the following screenshot, one may now assign colors to the commands `\let` and `\DeclareMathOperator`, differently from regular commands such as `\widehat` or `\overline`. This has been supported only for `\def` and `\newcommand`.<br>
The relevant hlgroups are `texCmdLet` and `texCmdDeclmathoper`. The default link for the hlgroups is `texCmd`.

2. One may color the terms to be defined (e.g., `\myname`, `\hat`, `\norm`, etc.) in `\let`, `\DeclareMathOperator`, and `\newcommand`. So far this has been supported only for `\def`.<br>
The relevant hlgroups are `texLetArgName`, `texDeclmathoperArgName`, and `texNewcmdArgName`. The default link is `texCmd`.

pr

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -549,6 +549,7 @@ function! s:init_highlights() abort " {{{1
   highlight def link texCmdBib               texCmd
   highlight def link texCmdClass             texCmd
   highlight def link texCmdDef               texCmd
+  highlight def link texCmdLet               texCmd
   highlight def link texCmdEnv               texCmd
   highlight def link texCmdE3                texCmd
   highlight def link texCmdFootnote          texCmd
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -544,24 +544,26 @@ function! s:init_highlights() abort " {{{1
   highlight def texStyleItal gui=italic      cterm=italic
 
   " Inherited groups
+  highlight def link texArgNew               texCmd
   highlight def link texAuthorOpt            texOpt
   highlight def link texCmdAccent            texCmd
   highlight def link texCmdAuthor            texCmd
   highlight def link texCmdBib               texCmd
   highlight def link texCmdClass             texCmd
-  highlight def link texCmdDef               texCmd
+  highlight def link texCmdDef               texCmdNew
   highlight def link texCmdEnv               texCmd
   highlight def link texCmdE3                texCmd
   highlight def link texCmdFootnote          texCmd
   highlight def link texCmdGreek             texCmd
   highlight def link texCmdInput             texCmd
   highlight def link texCmdItem              texCmdEnv
-  highlight def link texCmdLet               texCmd
+  highlight def link texCmdLet               texCmdNew
   highlight def link texCmdLigature          texSpecialChar
   highlight def link texCmdMath              texCmd
   highlight def link texCmdMathEnv           texCmdEnv
   highlight def link texCmdMathText          texCmd
-  highlight def link texCmdNewcmd            texCmd
+  highlight def link texCmdNew               texCmd
+  highlight def link texCmdNewcmd            texCmdNew
   highlight def link texCmdNewenv            texCmd
   highlight def link texCmdNoSpell           texCmd
   highlight def link texCmdPackage           texCmd
@@ -579,7 +581,7 @@ function! s:init_highlights() abort " {{{1
   highlight def link texCmdVerb              texCmd
   highlight def link texCommentAcronym       texComment
   highlight def link texCommentURL           texComment
-  highlight def link texDefArgName           texCmd
+  highlight def link texDefArgName           texArgNew
   highlight def link texDefParm              texParm
   highlight def link texE3Cmd                texCmd
   highlight def link texE3Delim              texDelim
@@ -593,7 +595,7 @@ function! s:init_highlights() abort " {{{1
   highlight def link texFilesOpt             texFileOpt
   highlight def link texGroupError           texError
   highlight def link texLetArgEqual          texSymbol
-  highlight def link texLetArgName           texCmd
+  highlight def link texLetArgName           texArgNew
   highlight def link texLigature             texSymbol
   highlight def link texMathArg              texMathRegion
   highlight def link texMathArrayArg         texOpt
@@ -612,7 +614,7 @@ function! s:init_highlights() abort " {{{1
   highlight def link texMathSub              texMathRegion
   highlight def link texMathSuper            texMathRegion
   highlight def link texMathSymbol           texCmd
-  highlight def link texNewcmdArgName        texCmd
+  highlight def link texNewcmdArgName        texArgNew
   highlight def link texNewcmdOpt            texOpt
   highlight def link texNewcmdParm           texParm
   highlight def link texNewenvArgName        texEnvArgName
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -251,11 +251,12 @@ function! vimtex#syntax#core#init() abort " {{{1
   call vimtex#syntax#core#new_arg('texDefArgBody')
 
   " \let
-  " Note: define texLetArgEqual after texLetArgBody; order matters
-  " E.g. in '\def\eq==' we want: 1st = is texLetArgEqual, 2nd = is texLetArgBody
   syntax match texCmdLet "\\let\>" nextgroup=texLetArgName skipwhite skipnl
   syntax match texLetArgName  contained nextgroup=texLetArgBody,texLetArgEqual skipwhite skipnl "\\[a-zA-Z@]\+"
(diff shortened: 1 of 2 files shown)


clason:
That's a great addition! One suggestion: I would introduce an "umbrella" highlight group to link the four different `tex*Name` syntax groups to (which is then mapped to `texCmd`. This makes it easier for users to override the highlight in a consistent way (since I assume if you want to color one differently, in most cases you want to color _all_ differently).

autoload/vimtex/syntax/core.vim
```diff
@@ -537,6 +549,7 @@ function! s:init_highlights() abort " {{{1
   highlight def link texCmdBib               texCmd
   highlight def link texCmdClass             texCmd
   highlight def link texCmdDef               texCmd
+  highlight def link texCmdLet               texCmd
```
lervag, line 552: Although partially random: I've tried to keep these lines alphabetically sorted... same for the following lines (583 and 584). :)

autoload/vimtex/syntax/p/amsmath.vim
```diff
@@ -30,6 +30,18 @@ function! vimtex#syntax#p#amsmath#load(cfg) abort " {{{1
...
+  syntax match texCmdDeclmathoper nextgroup=texDeclmathoperArgName skipwhite skipnl "\\DeclareMathOperator\>\*\?"
+  call vimtex#syntax#core#new_arg('texDeclmathoperArgName', {
+        \ 'next': 'texDeclmathoperArgBody',
+        \ 'contains': ''
+        \})
+  call vimtex#syntax#core#new_arg('texDeclmathoperArgBody')
+
+  highlight link texCmdDeclmathoper     texCmd
+  highlight link texDeclmathoperArgName texCmd
+
```
lervag, line 44: I like to avoid unnecessary empty lines.

autoload/vimtex/syntax/core.vim
```diff
@@ -247,6 +250,15 @@ function! vimtex#syntax#core#init() abort " {{{1
   syntax match texDefParm contained "#\+\d" containedin=texDefParmPre,texDefArgBody
   call vimtex#syntax#core#new_arg('texDefArgBody')
 
+  " \let
+  " Note: define texLetArgEqual after texLetArgBody; order matters
+  " E.g. in '\def\eq==' we want: 1st = is texLetArgEqual, 2nd = is texLetArgBody
```
lervag, line 255: Should the comment be `E.g. in \let\eq`, not `\def\eq`?

autoload/vimtex/syntax/core.vim
```diff
@@ -247,6 +250,15 @@ function! vimtex#syntax#core#init() abort " {{{1
...
+  syntax match texCmdLet "\\let\>" nextgroup=texLetArgName skipwhite skipnl
+  syntax match texLetArgName  contained nextgroup=texLetArgBody,texLetArgEqual skipwhite skipnl "\\[a-zA-Z@]\+"
+  syntax match texLetArgName  contained nextgroup=texLetArgBody,texLetArgEqual skipwhite skipnl "\\[^a-zA-Z@]"
```
lervag, line 258: It could be nice with an example for both of the ArgName variants.

lervag (commented):
Thanks, this looks very good. I have some minor comments.

lervag:
> That's a great addition! One suggestion: I would introduce an "umbrella" highlight group to link the four different `tex*Name` syntax groups to (which is then mapped to `texCmd`. This makes it easier for users to override the highlight in a consistent way (since I assume if you want to color one differently, in most cases you want to color _all_ differently).

I agree, this is a good suggestion. How about `texCmdNew` for the name?

autoload/vimtex/syntax/core.vim
j1-lee: > > That's a great addition! One suggestion: I would introduce an "umbrella" highlight group to link the four different `tex*Name` syntax groups to (which is then mapped to `texCmd`. This makes it easier for users to override the highlight in a consistent way (since I assume if you want to color one differently, in most cases you want to color _all_ differently).
> 
> I agree, this is a good suggestion. How about `texCmdNew` for the name?

Good suggestion! I tried to add `texCmdNew` and then link `texCmdNewcmd`, `texCmdDef`, `texCmdLet`, and `texCmdDeclmathoper` to it. Then `texCmdNewenv` was staring at me.

I guess `\newenvironment` (or `\newtheorem` if one should want to add it in the future) is kinda different from others in the sense that it does not define a TeX _command_, and for this reason users might want this _not_ to be linked to `texCmdNew`.

1. If this is the case, the names (`texCmdNew` and `texCmdNewenv`) may sound like `texCmdNewenv` is linked to `texCmdNew`. Two options:
  (a) `texCmdNew` is fine.
  (b) Think of another name for this umbrella highlight group to avoid the potential confusion.

2. If this is not the case, then I'll also link `texCmdNewenv` to `texCmdNew`. (But personally I am reluctant to do this...)

j1-lee:
> > That's a great addition! One suggestion: I would introduce an "umbrella" highlight group to link the four different `tex*Name` syntax groups to (which is then mapped to `texCmd`. This makes it easier for users to override the highlight in a consistent way (since I assume if you want to color one differently, in most cases you want to color _all_ differently).
> 
> I agree, this is a good suggestion. How about `texCmdNew` for the name?

Good suggestion! I tried to add `texCmdNew` and then link `texCmdNewcmd`, `texCmdDef`, `texCmdLet`, and `texCmdDeclmathoper` to it. Then `texCmdNewenv` was staring at me.

I guess `\newenvironment` (or `\newtheorem` if one should want to add it in the future) is kinda different from others in the sense that it does not define a TeX _command_, and for this reason users might want this _not_ to be linked to `texCmdNew`.

1. If this is the case, the names (`texCmdNew` and `texCmdNewenv`) may sound like `texCmdNewenv` is linked to `texCmdNew`. Two options:
  (a) `texCmdNew` is fine.
  (b) Think of another name for this umbrella highlight group to avoid the potential confusion.

2. If this is not the case, then I'll also link `texCmdNewenv` to `texCmdNew`. (But personally I am reluctant to do this...)

lervag:
Good question, and thanks for writing out the two possible paths.

I'm a pragmatic, and I don't really aim for perfect. As such, I think perhaps the second solution looks fine. But I don't have a very strong opinion, so I'll be glad to hear @clason's opinion as well (and others who might want to chip in).

Still, it might be that `texArgNew` is a better name regardless?

clason:
It's a pickle, alright :)

To be honest, I don't think there's going to be a group name schema that makes these "discoverable" without documentation, so I wouldn't worry too much about it, as long as there are no collisions (and maybe no groups that only differ in capitalization).

Nevertheless, as another alternative I offer `texDefcmd` as the base name for the umbrella group (with `texDevcmdArg` for the name and possibly -- if that's not too silly for you -- `texDevcmdCmd` for the target); the rationale being that these are special commands for defining commands?

autoload/vimtex/syntax/core.vim
j1-lee: The example pertains to the ordering of the lines (259--260) following these lines, not to the `ArgName`s. (`ArgName` variants are directly copied from `\def`.) I'll probably move the comment to right before these relevant lines, and make it more clear...

j1-lee:
Thanks for the inputs! I'll then let the umbrella group subsume `texCmdNewenv` as well. As for the name of the umbrella group, I personally prefer `texCmdNew` and `texArgNew` as @lervag mentioned. `texDefcmd` also is a nice suggestion (thanks!), but I guess "def" and "cmd" are overloaded (slightly more than "new"), especially since we already have `texCmdDef`. I'll probably proceed in this direction if you find it okay. Thanks again for the feedbacks!

lervag:
> Thanks for the inputs! I'll then let the umbrella group subsume `texCmdNewenv` as well. As for the name of the umbrella group, I personally prefer `texCmdNew` and `texArgNew` as @lervag mentioned. `texDefcmd` also is a nice suggestion (thanks!), but I guess "def" and "cmd" are overloaded (slightly more than "new"), especially since we already have `texCmdDef`. I'll probably proceed in this direction if you find it okay. Thanks again for the feedback!

It is OK by me. It's never easy to find good names, but I think we're close. If you make the changes, then we can think a little bit more about the naming of the umbrella group in the meantime. It might be easier to think of something better when I see the code.

j1-lee:
So, I have updated the code. In contrast to what I have said, I did not include `newenvironment` (i.e., `texCmdNewenv` and `texNewenvArgName`) in the new umbrella groups; this is because I found out that `texNewenvArgName` originally is linked to `texEnvArgName`, not `texCmd`.

If I included `texNewenvArgName` in `texArgNew` and link the latter to `texCmd`, this would set the default color of the new environment names as the same as commands. I don't believe this would be a natural default behavior...

I apologize for confusion. Any comments are welcome!

lervag:
Thanks, I like this and thing it is good enough for merge already. Now I'll play with colors.. :)

PR #1876, opened 2020-12-01
Fix syntax match pattern for starred variants of \cite

j1-lee:
This PR attempts to fix syntax match pattern for starred variants of `\cite`.

Previously, the syntax match pattern could not properly handle commands such as `\cite*{bibtexkey}`; it only matches `\cite` (without `*`) only, resulting in the subsequent arg `bibtexkey` matched to the generic `texGroup` which is subject to the spell check.

Before: (the red text with the wriggly underline is marked as "bad" by spell)
before

After:
after


lervag:
Thanks!

Issue #1880, opened 2020-12-04
Updated Check for Math Zone


Konfekt:
For reference: since the new syntax highlighting, a Vimscript function `InMathZone()` that checks whether the cursor is in a math zone now has to take account of the new highlighting groups `texMathRegion`s (in addition to the built-in ones `texMathZone`s), that is,

```vim
function! InMathZone()
  return match(map(synstack(line('.'), max([col('.') - 1, 1])),
        \ 'synIDattr(v:val, ''name'')'),
        \ '^\%(texMathZone\%([A-L]S\?\|[V-Z]\)\|texMathRegion\%(Ensured\|Env\|EnvStarred\|X\|XX\)\)$')
        \ >= 0
endfunction
```

lervag:
This function is not necessary, because it is already available from vimtex directly: `vimtex#syntax#in_mathzone`. I know it is not documented and so it is not surprising that you (and generally people) don't know of it, but since you are posting this issue I thought it was OK to mention it :)

lervag:
Btw: With Vimtex v2, there is no such thing as "built-in `texMathZone`s.

Konfekt:
Yes, sorry for the confusion, built-in referred to built into Vim, in `$VIMRUNTIME`. The use case in mind was a function to define abbreviations that only expand in Math zones, independent of the used TeX plug-in, such as 

```vim
inoreabbrev <expr> ==     InMathZone() ? '\equiv' : '=='
```

lervag:
Btw, your function could be greatly simplified:

```vim
function! InMathZone()
  return match(map(synstack(line('.'), max([col('.') - 1, 1])),
        \ 'synIDattr(v:val, ''name'')'),
        \ '^texMath\%(Zone\|Region\)')
        \ >= 0
endfunction
```

I don't see any region of matching the final part, as I think the prefix `texMathZone` and `texMathRegion` should match all the relevant groups.

Konfekt:
Yes, the definition is pedantic in limiting exactly to the existing syntax groups as it is unlikely that a future syntax group matching the prefix will not be a math zone.

Konfekt:
If returning to the naming of `Zone` instead of `Region`, then for all practical ends the regular expression ` '^texMathZone` would suffice. It would also correspond to the naming of `vimtex#syntax#in_mathzone`.

PR #1884, opened 2020-12-06
syntax#in_mathzone: take account of text zones inside math zones

Konfekt:
See https://github.com/lervag/vimtex/issues/1880


lervag:
Thanks, this was clever!

PR #1890, opened 2020-12-09
use correct conceal symbols for \phi, \varphi

BertrandSim:


lervag:
No, I think you are wrong. See e.g. [this stackoverflow thread](https://tex.stackexchange.com/questions/304534/what-does-the-var-prefix-stand-for-in-varphi-and-varepsilon-etc), where it seems clear that we should use φ for `\varphi` and ϕ for `\phi`.

BertrandSim:
> we should use φ for `\varphi` and ϕ for `\phi`.

Yes, I fully agree on this... And I believe this is what I proposed. 

Perhaps the symbols appear differently for me?
This is how the PR looks like on my end. Does it appear the same for you?

lervag:
And this is how it looks on my end:

2020-12-10_09:42_screenshot

lervag:
The unicode characters we use right now are:

* `\varphi`: [03c6](https://www.fileformat.info/info/unicode/char/03c6/index.htm)
* `\phi`: [03d5](https://www.fileformat.info/info/unicode/char/03d5/index.htm)

lervag:
Perhaps you are using a font that has switched these characters?

BertrandSim:
Thank you for your comments. Indeed, I am using Consolas in vim, which swaps the characters. github also uses Consolas on my end, thus swapping the characters as well. It seems that other fonts also suffer from the same issue, (Calibri, Consolas, Lucida Sans Unicode, Palatino Linotype, and Symbola, as mentioned [here](https://meta.stackexchange.com/questions/313993/unicode-phi-characters-inconsistent-between-body-and-preview-and-across-platform)). 

In any case, thank you for taking the time to respond to me. I guess I should look for a new font then. Cheers

lervag:
No problem, and thanks for the reference to the relevant discussion/thread. An interesting difference between font, in the very least!

PR #1893, opened 2020-12-12
fix rm shell command

JingMatrix:
Original command misses dot between root file name and extension and doesn't
consider name escape in shell.


lervag:
Thanks! I've merged and made a minor modification.

Issue #325, opened 2016-01-12
Correct way to input quotes?


anjiro:
Normally in LaTeX I write curly quotes with

    ``hello''

But using vimtex, `` ` `` is mapped to `g:vimtex_imaps_leader`, which means to type two literal backticks I have to type `` ` `` four times instead of two. Putting in quotes is a pretty common activity, so I assume I'm missing something obvious here...

lervag:
You can change `g:vimtex_imaps_leader` to something else, e.g. with the following line in your `.vimrc` file:

```vim
let g:vimtex_imaps_leader = ';'
```

Personally, I like the idea of using an automatic snippet for these curly quotes, e.g. (example with both neosnippet and UltiSnips):

```vim
" With neosnippet
inoremap <silent><expr> " neosnippet#anonymous("``${1}''${0}")

" With UltiSnips
inoremap <silent> " "<c-r>=UltiSnips#Anon("``{$1}''$0", '"', '', 'i')<cr>
```

Rmano:
I have a normal snippet --- like 
```
snippet " "Latex quote" i
\`\`$1''
endsnippet
```

which works quite well (it's `"` + `TAB`, still fast)

PR #1907, opened 2021-01-01
Add imapping for ``

LinuxMercedes:
This mapping allows users to type ``` `` ``` and get ``` `` ```, rather than having to type ``` ```` ```.

The mapping is tightly focused, so hopefully it will not affect people who have customized `imaps_list` and `imap_leader`.

In addition, the default `imaps_list` location moved; this updates the docs with that move.

Fixes https://github.com/lervag/vimtex/issues/325.

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -142,7 +142,7 @@ function! vimtex#options#init() abort " {{{1
 
   call s:init_option('vimtex_imaps_enabled', 1)
   call s:init_option('vimtex_imaps_disabled', [])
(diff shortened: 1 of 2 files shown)


lervag:
Good idea, but would it not be better if this depended on the value of `g:vimtex_imaps_leader`?

LinuxMercedes:
That would also work! Shall I switch to that?

lervag:
I've made the necessary modifications myself. Thanks for the PR!

lervag:
Ok, merged now.

PR #1908, opened 2021-01-03
fix typo calling 'ᵨ' delta instead of rho

ces42:
`autoload/vimtex/syntax/core.vim` has a typo where the character `ᵨ` is used to conceal a subscript delta instead of a subscript rho. This should fix it.


lervag:
Thanks!

PR #1925, opened 2021-01-17
Update vimtex-faq-zathura-osx

danielfalbo:
For a fresh zathura installation, we need to tap [zegervdv/zathura](https://github.com/zegervdv/homebrew-zathura) first.

P.s.: maybe `vimtex-faq-zathura-macos` would be a better name for that documentation section since `OS X` is now officially called `macOS`?


lervag:
Thanks! I've updated the name accordingly.

PR #1928, opened 2021-01-22
VimtexDocPackage: add support for environments

apfelkuchen6:
Previously calling VimtexDocPackage with the cursor over `\begin{foo}`
showed documentation related to \begin. Showing documentation
related to foo is probably always more useful.

As the environments are stored in the same completion file as the
commands (in the format ^\begin{foo}$), we can mostly wrap the code for
commands to get nice documentation for environments.

### DIFF
--- a/autoload/vimtex/doc.vim
+++ b/autoload/vimtex/doc.vim
@@ -62,9 +62,24 @@ function! s:packages_get_from_cursor() abort " {{{1
     return s:packages_from_usepackage(l:cmd)
   elseif l:cmd.name ==# '\documentclass'
     return s:packages_from_documentclass(l:cmd)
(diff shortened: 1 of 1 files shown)


lervag:
> Previously calling VimtexDocPackage with the cursor over `\begin{foo}`
> showed documentation related to \begin. Showing documentation
> related to foo is probably always more useful.

Agreed!

> As the environments are stored in the same completion file as the
> commands (in the format ^\begin{foo}$), we can mostly wrap the code for
> commands to get nice documentation for environments.

Seems sensible. I'll review the code now.

lervag:
Thanks! I made some minor changes, but essentially your code was good and this definitely adds value to Vimtex. Appreciated!

PR #1941, opened 2021-02-01
Expand filenames, for example "~"

patricknraanes:
Example use case:

I keep the pdfs in a folder that is synced across different machines. So my HOME changes. So I would like to not hardcode it in the file links.


lervag:
Thanks. I moved the `map()` to where the list of pdfs is created, but I agree this makes sense.

PR #1974, opened 2021-03-03
pushd/popd root when parsing bib for context menu

patricknraanes:
Finding the bib files for the context menu for cite items doesn't work unless we're in the right dir. To fix this, I made these changes, inspired by `completer_bib.gather_candidates`

### DIFF
--- a/autoload/vimtex/context/cite.vim
+++ b/autoload/vimtex/context/cite.vim
@@ -39,10 +39,13 @@ endfunction
 
 " }}}1
 function! s:handler.get_actions() abort dict " {{{1
+  " Work in root, as in completer_bib.gather_candidates
+  call vimtex#paths#pushd(b:vimtex.root)
   let l:entries = []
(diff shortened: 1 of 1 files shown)


autoload/vimtex/context/cite.vim
```diff
@@ -39,10 +39,13 @@ endfunction
 
 " }}}1
 function! s:handler.get_actions() abort dict " {{{1
+  " Work in root, as in completer_bib.gather_candidates
```
lervag, line 42: For most people "completer_bib.gather_candidates" is not fully understandable, so I propose a different wording here, for instance "Ensure we're at the root directory when locating bib files".

lervag:
Thanks, this is indeed useful. If there's to be a comment, I'd like for it to be slightly more understandable; except for that, I think this is a simple and useful improvement!

lervag:
Thanks!

PR #1977, opened 2021-03-04
Match left/right pairs as delimiters

psvenk:
Currently, some delimiter pairs are categorized as `texMathSymbol` while others are categorized as `texMathDelim`; this pull request recategorizes all delimiter pairs to be `texMathDelim`.

Screenshots from `test/test-syntax/test-core.tex` (with `:colo default`):

| Before | After |
| --- | --- |
| image | image |


lervag:
I think you are right about this. Thanks for noticing and taking action!

PR #1979, opened 2021-03-05
Remove texZone link for Asymptote syntax

psvenk:
Because `texAsymptoteZone` includes the `asy` syntax file, linking it to `texZone` makes otherwise unassigned parts of the code (e.g. variable names) to be in the `texZone` color instead of the default color; this pull request removes this link so that Asymptote code is highlight as if it were in a standalone `.asy` file.

Screenshots from `test/test-syntax/test-asymptote.tex` (with `:colo default`):

| Before | After |
| --- | --- |
| image | image |


lervag:
You are right about the problem, but not about the fix: The reason for adding the highlight definition is for the cases where the asymptote syntax files are not available. Thus, the highlight should be applied if the `l:opts` is empty.

I'll fix it and acknowledge this PR for making me aware of the problem.

lervag:
And, to be clear: thanks! I very much appreciate when people submit issues and simultaneously suggest a fix with a PR!

PR #1980, opened 2021-03-06
redraw in view function after synctex

JingMatrix:
I need add redraw to make it work on my computer.
```
Linux Matrix.xyz 5.8.0-45-generic #51-Ubuntu SMP Fri Feb 19 13:24:51 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -82,6 +82,7 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
 
   execute 'normal!' a:line . 'G'
   normal! zMzvzz
+  redraw
 
   " Attempt to focus Vim
   if executable('pstree') && executable('xdotool')


lervag:
> I need add redraw to make it work on my computer.

Ok, but why? As you might understand, VimTeX is quite popular, and I can't remember having seen any issues that your PR addresses/fixes. I'm not saying I won't consider this, but I need more details on what this is supposed to fix/improve.

JingMatrix:
Sorry for saying too little about what happened to me, since I am used to directly change source code.

I am using `vimtex` (`vim` 8.2) with `zathura` and `tectonic`.

When I CTRL + click on `zathura` preview I expect synctex brings me to corresponding position in Latex code.

What actually happens is that I can tell the cursor position does move, while my screen stays on previous state. Because if I press `zz` or redraw using CTRL + L, I will go to the desired line.

After add `redraw`, this problem goes away.

lervag:
Ok, thanks! I also use Zathura and I don't have this particular issue, but I've looked at it and I can't see how this could be bad.

PR #1986, opened 2021-03-09
continuous mode: avoid fsync in writing compilation output

cridemichel:
in continuous mode (at least in mac osx with vim or macvim or in lubuntu with vim) when a recompilation of latex sources is triggered (e.g. writing the latex file) vim _stutters_ several times (i.e. 3-4 times there are some noticeable lags) and this is rather annoying since I use to save rather frequently my work. The culprit of this odd behavior is the following write inside the function `s:callback_continuous_output` in `autoload/vimtex/compiler/latexmk.vim`:

```vim
call writefile([a:msg], b:vimtex.compiler.output, 'a')
```
since a `fsync()` is invoked after it has been called. This problem can be fixed by adding the "S" flag which prevent a call to `fsync()` after writing, i.e. by replacing the above line with:

```vim
call writefile([a:msg], b:vimtex.compiler.output, 'aS')
```

I also noticed a similar call to `writefile` function in `s:compiler_jobs.exec()`. Maybe is it safer to add the `S` flag there too?

### DIFF
--- a/autoload/vimtex/compiler/arara.vim
+++ b/autoload/vimtex/compiler/arara.vim
@@ -209,7 +209,7 @@ endfunction
 " }}}1
 function! s:callback_nvim_output(id, data, event) abort dict " {{{1
   if !empty(a:data)
-    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'a')
+    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'aS')
   endif
 endfunction
 
--- a/autoload/vimtex/compiler/arara.vim
+++ b/autoload/vimtex/compiler/arara.vim
@@ -209,7 +209,7 @@ endfunction
 " }}}1
 function! s:callback_nvim_output(id, data, event) abort dict " {{{1
   if !empty(a:data)
-    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'aS')
+    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'a')
   endif
 endfunction
 
--- a/autoload/vimtex/compiler/generic.vim
+++ b/autoload/vimtex/compiler/generic.vim
@@ -238,7 +238,7 @@ endfunction
 " }}}1
 function! s:callback_nvim_output(id, data, event) abort dict " {{{1
   if !empty(a:data)
-    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'aS')
+    call writefile(filter(a:data, '!empty(v:val)'), self.output, 'a')
   endif
 endfunction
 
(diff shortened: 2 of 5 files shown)


lervag:
Thanks, this is quite interesting. I'm not so acquinted with `fsync()` (or C code and Linux kernel or whatever). What's the _drawback_ of adding the `S` flag?

lervag:
In any case, I see from `:help 'fsync'` that the defaults are different between Vim and neovim. So, the question is if the `S` flag will lead to unexpected behaviour (e.g. delayed callbacks?).

cridemichel:
Hi, 
I forgot to report that I also gave neovim a try and indeed neovim seems to not have this issue, probably because it does not issue the fsync(). About possible drawbacks according to vim manual:
```
	When {flags} contains "s" then fsync() is called after writing
	the file.  This flushes the file to disk, if possible.  This
	takes more time but avoids losing the file if the system
	crashes.
```
hence without the call to ```fsync()``` one can loose compilation output after a system crash, but
after a system crash I would not be concerned about this much...

cridemichel:
> But I think there may be more places where it should be applied.

I also applied the same patch in ```s:compiler_jobs.exec()``` and everything is ok so far.
I did not check all places where ```writefile``` is called but maybe it is wise to add the "S" flag everywhere to have a real (or more) asyncronous behavior of vimtex...

lervag:
I'm not sure if everywhere is necessary. But at the similar places of callbacks, and in the corresponding `arara.vim`, `latexrun.vim` (if applicable; I don't remember). Feel free to look into it, and if not, I'll also take a look myself when I get the time.

cridemichel:
I have just added the "S" flag to ```writefile``` calls pertaining the other compilers (arara, latexrun, etc.) and I also added the "S" flag to the following call:
```vim
try
    call writefile([], self.output, 'S')
  endtry
```
in  s:compiler_nvim.exec(). 
According to neovim manual:
```text
When {flags} contains "S" fsync() call is not used, with "s"
		it is used, 'fsync' option applies by default. No fsync()
		means that writefile() will finish faster, but writes may be
		left in OS buffers and not yet written to disk. Such changes
		will disappear if system crashes before OS does writing.
```
hence I think that it as safe in neovim as it is in vim.

cridemichel:
I have just reverted the extended changes since in some cases are not relevant (the other compilers do not have continous mode) and in once case (the initial writefile) it is not safe to make it asyncrounous

lervag:
Thanks! I've merged this now, and I'll try to keep this in mind if some similar issue should pop up for a different backend.

cridemichel:
ok, thank you very much!

lervag:
You're welcome, and thanks for contributing!

tir. 9. mar. 2021, 18:27 skrev Cristiano De Michele <
notifications@github.com>:

PR #1990, opened 2021-03-11
Handle % in URL arguments properly

psvenk:
Percent signs are fairly common in URLs to encode special characters such as spaces (`%20`), so they should not be recognized as a TeX comment marker within URLs provided to `\url` and `\href`.


lervag:
Thanks!

PR #1991, opened 2021-03-11
Highlight \fbox argument as text mode

psvenk:
`\fbox`, which is useful for enclosing an important result in a box in math mode, takes a text mode argument; change the syntax highlighting to reflect this fact.


lervag:
Thanks!

PR #1996, opened 2021-03-21
fix: fzf toc display and jump error when spaces in file path

WeixuanZ:
The file path and heading extracted by splitting at spaces would be wrong if the path contains spaces itself. Using \u2007 as a hack to preserve formatting (junegunn/fzf#585)

### DIFF
--- a/autoload/vimtex/fzf.vim
+++ b/autoload/vimtex/fzf.vim
@@ -21,10 +21,12 @@ function! vimtex#fzf#run(...) abort " {{{1
   " Note: The '--with-nth 3..' option hides the first two words from the fzf
   "       window. These words are the file name and line number and are used by
   "       the sink.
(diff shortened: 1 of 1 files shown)


lervag:
> The file path and heading extracted by splitting at spaces would be wrong if the path contains spaces itself. Using \u2007 as a hack to preserve formatting

Seems like a simple and useful hack. Thanks!

Do you have or can you show a simple test case for this?

lervag:
That is: Can you show a simple/minimal example that currently fails that this PR will fix?

WeixuanZ:
Sure, this is when `tests/test-toc/main.tex` is renamed to `test main.tex` (i.e. including a space) and using the current code:

old

Selecting `main.tex Preamble`, would open `test/test-toc/test` as a new file. 

With this PR:

new

lervag:
Thanks!

Issue #1965, opened 2021-02-20
[conceal] \cite, \citep, \citet to key list or just a simple marker


whisperity:
This is a suggestion request to extend Vimtex's conceal feature with handling `\cite`, `\citep`, and `\citet`. If possible, it should be concealed out with either a `(...)` (where `...` is the bib key for the cited work), or with `[X]` where `X` is a literal `X` to show there is a citation there, instead of the lengthy sequence a bib key could take.

Unfortunately, I'm not sure if conceal can handle "parsing" such elaborate sequences.

lervag:
I don't think Vim can do this type of "advanced" conceals, that is, we can only do relatively simple replacements.

To be able to answer/consider this request more clearly, perhaps you could show some examples of specifically how you would like this to work? I think the following shows approximately what I think can be achieved:

```tex
% Raw tex
See e.g. \cite{myref} for more info.
\citet{myref} show that ...

% Concealed
See e.g. [myref] for more info.
[myref] show that ...
```

A "perfect" solution would of course parse the reference key, but I think that is impossible (or at least too hard for me to consider). Further, a perfect solution would differentiate `\citet` and `\citep` and similar, but I would propose to simplify this to concealing boundaries, i.e. `\cite…{ -> [` and `} -> ]`. But is this sufficient for this to be an interesting feature?

whisperity:
No need for it to be perfect. And yes, I know the conceal is iffy... e.g. for C++ I couldn't even manage to get it to conceal `else if` and `else` and `if` differently, even... :frowning:

Parsing the reference key is definitely impossible in the general case, and not worth the effort.
The purpose of *conceal* is to remove the bloat. The `\cite(t|p)?` here is bloat, similarly how `\emph` is bloat if your terminal supports italic letters in the first place.

Is it hard to "throw away" the insides of the concealed boundaries?

lervag:
> Parsing the reference key is definitely impossible in the general case, and not worth the effort.

Good to hear we agree on that.

> The purpose of _conceal_ is to remove the bloat. The `\cite(t|p)?` here is bloat, similarly how `\emph` is bloat if your terminal supports italic letters in the first place.

Well, if we concealed `\cite{xxx}` so it looks like `[xxx]`, then it could be hard to realize that this is really a citation and not just the actual text `[xxx]`. But I guess if you were to use this feature it would probably be what you want.

> Is it hard to "throw away" the insides of the concealed boundaries?

[...]

PR #2002, opened 2021-03-22
Feature: Conceal support for \cite commands

lervag:
See #1965 for the current discussion.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -2025,6 +2025,13 @@ OPTIONS                                                        *vimtex-options*
 
   Default value: 1.
 
+*g:vimtex_syntax_conceal_default*
+  Specify the default flag for the conceal feature. This is used e.g. in
+  |vimtex_syntax_conceal| and for package specific options in
+  |vimtex_syntax_packages|.
+
+  Default value: 1
+
 *g:vimtex_syntax_conceal*
   A dictionary for specifying which core conceal features to activate. This
   mostly implies concealing particular elements with a replacement unicode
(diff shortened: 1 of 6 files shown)


whisperity:
I think it is great! :trophy: 

Screenshot_20210325_100729

(Please ignore the ugly white-ish background around concealed symbols, it's something that is broken in my (N)Vim and is irrespective of the plugin itself. Even C++ conceals are broken. :frowning: )

lervag:
> I think it is great!

Thanks!

One comment to your text: You should be aware of `\citet{...}` - don't write the author name + "et al" manually. :)

> (Please ignore the ugly white-ish background around concealed symbols, it's something that is broken in my (N)Vim and is irrespective of the plugin itself. Even C++ conceals are broken. frowning )

The background is controlled by the `Conceal` highlight group. You should read `:help hl-Conceal`.

Open question: What should the default configuration be? I think we should keep it enabled by default, but with the type set to brackets. Agreed?

whisperity:
> One comment to your text: You should be aware of `\citet{...}` - don't write the author name + "et al" manually. :)

This is an older and already published paper, which uses `IEEEtran` template. `\citet` is a `natbib` command which isn't available there, I think.
It is also usual with numbered citations that you do not keep the author and the cite number together, i.e. things like

> Pearsson and Fitch discussed that horses are brown [8].


> > (Please ignore the ugly white-ish background around concealed symbols, it's something that is broken in my (N)Vim and is irrespective of the plugin itself. Even C++ conceals are broken. frowning )
> 
> The background is controlled by the `Conceal` highlight group. You should read `:help hl-Conceal`.

Yeah, figured that, and this only happens with one particular colour scheme I am using, which is older than the *conceal* feature itself. There definitely isn't **any** conceal setting in the theme file, so now I'm trying to hunt down which other Vim plugin I use set the style... It's set to some odd value: `ctermfg=7 ctermbg=242`... And not by the theme file. Must be a link somewhere.

>
> Open question: What should the default configuration be? I think we should keep it enabled by default, but with the type set to brackets. Agreed?

Yeah, enabled, and maybe instead of the emoji the bracketed one is better.

lervag:
> > The background is controlled by the `Conceal` highlight group. You should read `:help hl-Conceal`.
> 
> Yeah, figured that, and this only happens with one particular colour scheme I am using, which is older than the _conceal_ feature itself. There definitely isn't **any** conceal setting in the theme file, so now I'm trying to hunt down which other Vim plugin I use set the style... It's set to some odd value: `ctermfg=7 ctermbg=242`... And not by the theme file. Must be a link somewhere.

You should read [this gist by romainl](https://gist.github.com/romainl/379904f91fa40533175dfaec4c833f2f); essentially:

```vim
augroup MyColors
  autocmd!
  autocmd ColorScheme * highlight Conceal ...
augroup END
```

doc/vimtex.txt
```diff
@@ -2070,18 +2077,39 @@ OPTIONS                                                        *vimtex-options*
...
 
-  Default value: 1
+    type~
+      Specify the type of concealment. There are two options, and the
+      difference is best explained by example:
+
+        Value       LaTeX     Concealed
+        -----       -----     ---------
+        'icon'      `\cite{x}`  `📖`
+        'brackets'  `\cite{x}`  `[x]`
```
whisperity, line 2101: Maybe it would be worth our while to replace `x` with `xyz` and show that the "brackets" concealment still shows the **full key** that's in there (as opposed to replacing `\cite{arbitrary_long_identifier}` with just `[x]`.

Melkster:
Looks great! Maybe it's too late now since you just merged this PR, but the only feedback I have is that perhaps `'brackets'` should be the default value since it's less intrusive. Personally though I'll probably be using `'icon'` :)

lervag:
`'brackets'` _is_ the default value :)

lervag:
> ... Maybe it's too late now ...

And it's never too late :)

postylem:
Great new feature! But I think there may be causing an issue (or one that is just coincidentally revealed).

When there is a concealed citation inside of another command like `\something{...\cite{xxx}}` the syntax highlighting marks the closing brace as an error.

Screen Shot 2021-03-26 at 02 28 50

When the cite concealing is turned off, with`let g:vimtex_syntax_conceal_default=0` (or, equivalently, setting the `cites` key to 0, in `g:vimtex_syntax_conceal_default`), it doesn't show this highlighting error:

Screen Shot 2021-03-26 at 02 29 58

Interestingly, the issue doesn't arise with the icon style concealment.

Screen Shot 2021-03-26 at 02 42 51

These screenshots are on terminal Vim 8.2.2576 on macOS 11.2.3.  I also found the same behaviour with neovim (0.5.0).

lervag:
> Great new feature! But I think it may be causing an issue (or there is one that is just coincidentally revealed here).

Sorry about that, I think it is fixed now! :)

Melkster:
> `'brackets'` _is_ the default value :)

Oh you're right, my bad!

PR #2004, opened 2021-03-23
Add minor documentation fixes

perebohigas:
Cool Vim plugin for a nice LaTeX integration!! 
If you want them, here are a couple of small fixes for the `DOCUMENTATION.md`.
Thanks for your work!!

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -297,7 +297,7 @@ supported and enabled, to automatically parse log files and display potential
 errors after compilation.
 
 The files `vimtex/autoload/vimtex/qf/*.vim` define different types of log
-parsers. E.g., `qf/bibtex.vim` is used to parse `.blg` files for bibtex related
+parsers. E.g., `qf/bibtex.vim` is used to parse `.blg` files for BibTeX related
 warnings and errors, and `qf/latexlog.vim` parses `.log` files for LaTeX
 warnings and errors. `qf/pulp.vim` defines an alternative log parser that can
 be used instead of `latexlog.vim`. See also `:help g:vimtex_quickfix_method`.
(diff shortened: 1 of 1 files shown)


DOCUMENTATION.md
```diff
@@ -94,16 +94,16 @@ for loading all of the VimTeX functionalities, except:
 * syntax highlighting is loaded from `syntax/tex.vim`
 * indentation is loaded from `indent/tex.vim`
 
-The main init function calls `vimtex#mymodule#init_buffer()` for each
+The main `init` function calls `vimtex#mymodule#init_buffer()` for each
```
lervag, line 97: Thanks, but I think "init" was used as text, not as a function name. Perhaps it is better to use "initialisation"? This change would be confusing, because there is nothing that is simply called `init`.

DOCUMENTATION.md
```diff
@@ -370,7 +370,7 @@ vimtex-denite`.
 
 # test
 This directory is used to, you guessed it, define tests for the VimTeX code.
-The tests are built on top of a Makefile based workflow. The top level Makefile
+The tests are built on top of a Makefile based workflow. The top level 'Makefile'
```
lervag, line 373: Perhaps better: "... based workflow. `test/Makefile`"?

lervag (changes requested):
Thanks! I appreciate suggestions like this. I had a couple of minor comments, though.

lervag:
> Cool Vim plugin for a nice LaTeX integration!! 
> If you want them, here are a couple of small fixes for the `DOCUMENTATION.md`.
> Thanks for your work!!

Thanks for the kind words, and thanks for the fixes. Improving docs is also an important part of a project, and so I appreciate it!

lervag:
Thanks!

PR #2012, opened 2021-03-29
update example using VimtexClean upon quit

postylem:
The example is using some outdated function (I think).
I wanted to get the behaviour described, and I got it using `VimtexClean` instead of `call vimtex#compiler#clean(0)` which didn't do anything (as I don't think that function exists?)

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -3362,8 +3362,8 @@ Examples: >
   " Compile on initialization, cleanup on quit
   augroup vimtex_event_1
     au!
-    au User VimtexEventQuit     call vimtex#compiler#clean(0)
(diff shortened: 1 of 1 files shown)


lervag:
> The example is using some outdated function (I think).
> I wanted to get the behaviour described, and I got it using `VimtexClean` instead of `call vimtex#compiler#clean(0)` which didn't do anything (as I don't think that function exists?)

https://github.com/lervag/vimtex/blob/6c5e13317d4abb2baa9ba0544fcf59b2b12f5659/autoload/vimtex/compiler.vim#L258-L277

The functions are not outdated and the both exist. The example is valid. However, I do agree it is probably better to refer to the already documented ex commands.

PR #2014, opened 2021-04-03
Set foldmethod=diff when doing diff.

jaekyeom:
Thanks for this great project! I find vimtex very helpful for my work.
Currently, when `g:vimtex_fold_manual=1`, vimtex enforces folding based on the content of the tex file even in the diff mode.
This PR changes it by setting `foldmethod=diff` during diff-ing, which I think is a more plausible and convenient behavior.
Please consider merging it.


lervag:
I agree, this does look sensible.

lervag:
Thanks!

Issue #1978, opened 2021-03-04
Conceal additional commands like \section{} and \texttt{}


Melkster:
VimTeX conceals some commands like for instance `\textbf{}`, `\textit{}` which I love.

Is there any plan to conceal more commands? The ones that I personally miss the most are for `\section{}`/`\subsection{}`/`\subsubsection{}` and `\texttt`.

For `\texttt{}` I think that the surrounding command should be concealed and the text inside it should get a syntax highlighting.

As for `\...section`, in order to differentiate between the different levels it could for instance get concealed into the markdown syntax like this:

**Without conceal:**

```tex
\section{First Section}
Foo bar

\subsection{Second Section}
Foo bar

\subsubsection{Third Section}
Foo bar
```

**With conceal:**

```markdown
# First Section
Foo bar

## Second Section
Foo bar

### Third Section
Foo bar
```

Would this be possible to implement?

lervag:
> VimTeX conceals some commands like for instance `\textbf{}`, `\textit{}` which I love.

:)

(I personally don't like it, but I understand it is a popular feature!)

> Is there any plan to conceal more commands? The ones that I personally miss the most are for `\section{}`/`\subsection{}`/`\subsubsection{}` and `\texttt`.

No plan except that I'll be glad to improve as suggestions are provided, such as this.

> For `\texttt{}` I think that the surrounding command should be concealed and the text inside it should get a syntax highlighting.

Ok, so: `\texttt{asdf} -> asdf` where `asdf` is given a specific highlighting; which kind, if so? This command is already matched as `texCmdStyle`, so I would propose that the contents would be `texStyleArgTT`.

> As for `\...section`, in order to differentiate between the different levels it could for instance get concealed into the markdown syntax like this:
> ...
> Would this be possible to implement?

Actually, no: syntax conceals only allow a single substitution character of the concealed text. In this particular case, we can conceal the `\*section{...}`, but we can't use more than a single character. So, we could do something like this:

```tex
% before
\section{sec}
\subsection{subsec}

% after
¤sec
#subsec
```

The initial character can be chosen quite freely, but we can't add a space separator or use multiple characters.

Thus, I'm not sure if concealing the `\section` family of commands is a very good idea?

lervag:
By the way, conceal of `\textttt` is also covered by #1968, so let's discuss that part there and let this discussion be about the `\section` stuff.

Melkster:
> which kind, if so?

[...]

PR #2022, opened 2021-04-06
Add conceal support for `\(sub)*sections`

lervag:
See #1978 for context.

Some things to discuss before merging:

* This direction of conceals is full of edge cases. For instance, how should we treat `\chapter`? I would perhaps claim we should only allow this particular type of concealment to the specific family `\(sub)*section`.

* I suspect this conceal feature will be much more opinionated. Perhaps it should be off by default? If so, then the `g:vimtex_syntax_conceal_default` becomes "obsolete", but it might have been a bad way to customize things anyways. Perhaps it is better to show the defaults explicitly, allow some defaults to be off, and instead add `g:vimtex_syntax_conceal_disable` to make it easy to disable all conceal features for whomever wants that.

* And, of course: is the current implementation satisfactory?


Melkster:
Looks good! Your conceal implementation looks way cleaner than my proof of concept hehe.

> how should we treat \chapter?

I personally don't use `\chapter{}` so perhaps I shouldn't be the one answering this, but perhaps it simply shouldn't be concealed like you suggest, or it could be concealed like `\section{}` but with another symbol than `#`, for instance `§`. I.e. `\chapter{Test}` would be concealed into `§ Test`.

> Perhaps it should be off by default?

I agree.

> If so, then the g:vimtex_syntax_conceal_default becomes "obsolete"

Yeah I think `g:vimtex_syntax_conceal` is a perfectly fine way to control the conceals. You could write out its default values in the docs, for instance:

```vim
g:vimtex_syntax_conceal = {
  \ 'styles': 1,
  \ 'accents': 1,
  \ 'greek': 1,
  \ 'math_fracs': 1,
  \ 'math_symbols': 1,
  \ 'math_super_sub': 1,
  \ 'cites': 1,
  \ 'fancy': 1,
  \ 'sections': 0,
  \ 'math_delimiters': 1,
  \ 'math_bounds': 1
  \ }
```

Adding `g:vimtex_syntax_conceal_disable` makes sense to me too.

lervag:
Thanks for your thoughts. I've pushed corresponding updates now. I propose to merge this, but I would be happy to hear a second opinion. Not sure if there may be "lurkers" reading this that would be interested in commenting? :)

Schmiddiii:
Hi, random "lurker" here.
This seems pretty good, but I still have some ideas for minor improvements.

First consider `\section*{Title}`. This is currently concealed as `# * Title`.
Would it be possible to leave out the `*` as I think this information is not neccessary. 
In my experience, either all sections are `section` or `section*`, not a mix.
This of course also applies for all `(sub)*section`s.

My next request is a conceal for `label`, as many `(sub)*section`s have a label beneath.
This is for example the case in the snippets for Ultisnips [here](https://github.com/honza/vim-snippets/blob/master/UltiSnips/tex.snippets#L131).
This could either be integrated into the `(sub)*section` conceal (if possible), e.g. `# Title (label)`, or be a new conceal, e.g.

```
# Title
↪ label
```

My last request would be the `(sub)*paragraph`.
The current conceal of `\paragraph{Text}` is `\paragraph Text` (just left out the `{}`).
This probably might be improved, but I have no idea how to conceal it.
To my information, makrdown does not have any character for a `paragraph`.
Also consider `paragraph*`. This is currently not concealed at all.
This is inconsistent and should probably be changed.

lervag:
> Hi, random "lurker" here.

Great, happy to "meat" you :)

> First consider `\section*{Title}`. ...
> In my experience, either all sections are `section` or `section*`, not a mix.

I've mixed things several times. Typically with things like adding a `\section*{Acknowledgement}`. However, I also tend to agree that the `# *` does not look so good, so I can agree on simply hiding the star as well.

> My next request is a conceal for `label`, as many `(sub)*section`s have a label beneath.
> This is for example the case in the snippets for Ultisnips [here](https://github.com/honza/vim-snippets/blob/master/UltiSnips/tex.snippets#L131).
> This could either be integrated into the `(sub)*section` conceal (if possible), e.g. `# Title (label)`, or be a new conceal, e.g.

Unfortunately, this may be more tricky: 1) I don't think we can conceal multiple lines, and 2) conceals generally only allow to conceal a single "matched part/region" with a single character. More complex conceals are all "hacks" that combine syntax rules in more or less twisted ways. :\

> My last request would be the `(sub)*paragraph`.
> The current conceal of `\paragraph{Text}` is `\paragraph Text` (just left out the `{}`).

Ah, that was actually an unintended side effect; a bug. Thanks for reporting. I'll revert this immediately.

lervag:
I'm sorry about the delay here; I've merged this now.

PR #2035, opened 2021-04-19
Fixes #1658 again

### LINKED ISSUE #1658: \cite completion and .bib file globbing
**Is your feature request related to a problem? Please describe it.**

The `\cite` completion list is empty for me with the following MWE:
```tex
\documentclass{article}
\usepackage{biblatex}
\addbibresource{*.bib}
\begin{document}
Test
\end{document}
```
`ref.bib`
```bibtex
@article{Agostini2019b,
    author        = "Agostini, M. and others",
    archivePrefix = "arXiv",
    collaboration = "Gerda",
    doi           = "10.1007/JHEP03(2020)139",
    eprint        = "1909.02522",
    journal       = "JHEP",
    pages         = "139",
    primaryClass  = "nucl-ex",
    title         = "{Modeling of GERDA Phase II data}",
    volume        = "03",
    year          = "2020"
}
```
And it works as expected if I `\addbibresource{ref.bib}`. Can you reproduce?

Would it be possible to add support for wildcards in the `\addbibresource` command? Or is there already some trick I can use to make this work?

TpmKranz:
The fix in 2efe8068d1c45c143a542e7fcee932fa9db14509 isn't used if BibLaTeX is detected, even though globbing is only possible with BibLaTeX's biber.

Since b7fbff179005845b03b713b7a46fa28784afdf42, that code path is only used if there aren't better methods of detecting bib files. That was okay because the only other method introduced was to parse BibTeX's blg file and BibTeX can't handle globbing anyway. But in 2536f043a78be504778733f22d654fa2efd0153e, support for biber's bcf file was added without also adding globbing support.

Apparently, globbing can be disabled per resource, so filenames that are candidates for globbing must be retained in the file list. This should only have an effect with curly braces, though. Globbing candidates that aren't actual filenames will be filtered out with `s:validate()`, as before.

See also: #1658


autoload/vimtex/bib.vim
```diff
@@ -11,6 +11,10 @@ function! vimtex#bib#files() abort " {{{1
       let l:bibs = map(
             \ filter(readfile(l:file), "v:val =~# 'bcf:datasource'"),
             \ {_, x -> matchstr(x, '<[^>]*>\zs[^<]*')})
+      let l:bibs_glob = filter(copy(l:bibs), {_, x -> x =~ '[*?{]' })
+      for l:f in l:bibs_glob
```
lervag, line 15: Since `l:bibs_glob` is only used here, I would prefer to avoid the variable:

```vim
       for l:f in filter(copy(l:bibs), {_, x -> x =~ '[*?{]' })
```

lervag (changes requested):
Thanks. This does seem pretty useful. It would be nice to have a simple example to test this. Perhaps we could even make a test?

autoload/vimtex/bib.vim
```diff
@@ -75,11 +79,10 @@ function! s:files_manual() abort " {{{1
         let l:entry = substitute(l:entry, '\\jobname', b:vimtex.name, 'g')
 
         for l:f in split(l:entry, ',')
-          if stridx(l:f, '*') >= 0
+          if l:f =~ '[*?{]'
             let l:files += glob(l:f, 0, 1)
-          else
-            let l:files += [fnamemodify(l:f, ':r')]
           endif
+          let l:files += [fnamemodify(l:f, ':r')]
```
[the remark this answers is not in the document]
TpmKranz: Because a file named literally `{1,2}.bib` (using `\addbibresource[glob=false]{{1,2}.bib}`) would not be found without that change. `glob('{1,2}.bib')` can only find `1.bib` and `2.bib`. This doesn't happend with `*` and `?` because these literal characters are described by their wildcard counterparts, so it would be possible to distinguish between `=~# '[*?]'` and `=~# '[{]'`, but that complicates the code unnecessarily because one could also have a `\addbibresource[glob=false]{*{1,2}.bib}`.
These are of course weird edge cases (who names their files that way!?), but why not just catch them?

autoload/vimtex/bib.vim
TpmKranz: Okay, that was an attempt at self-documenting code, but the `glob()` in the next line should be documentation enough. :)

autoload/vimtex/bib.vim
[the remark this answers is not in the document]
lervag: > These are of course weird edge cases (who names their files that way!?), but why not just catch them?

I guess my reaction originates from thinking "who names their files that way!?". Note that Vim's `glob()` function does not support the brace expansions (see `:help wildcards`). But I can't see any particular harm in this either, so let's run with it!

autoload/vimtex/bib.vim
lervag: Thanks. Just to add my 2c: I can understand the motivation of self-documenting code, but IMHO, this specific pattern is not more readable. E.g., it makes me wonder if the variable is used more places, and it adds another line of code/another statement. Still, I can see that, in some cases, it helps to add the variable to explain a sub statement that is hard to parse.

test/test-completion-bibtex/test_globbed.vim
```diff
@@ -4,15 +4,39 @@ filetype plugin on
 
 nnoremap q :qall!<cr>
 
+let s:build_dir = expand('<sfile>:r') . '.dir'
 let g:vimtex_cache_root = '.'
 let g:vimtex_cache_persistent = 0
+let g:vimtex_log_verbose = 0
+let g:vimtex_compiler_method = 'latexmk'
```
lervag, line 11: This is already the default value, so no need to specify the option.

lervag (commented):
From #2037:

> The test used to always take the second path because vimtex doesn't know about the biblatex package being loaded before a compile.

Ah, yes: there are now three branches in the `#files()` function: 1) the "biblatex + bcf" branch, 2) the "blg" branch, and 3) the manual branch. And the old code only checks the third.

However, I propose, instead of doing the compilation every time, we simply add a bcf file to the test assets. Then we can simplify everything quite a lot. I.e., we can have `test_globbed.bcf` with only a single line:

```xml
<bcf:datasource type="file" datatype="bibtex" glob="false">test_gl*.bib</bcf:datasource>
```

Of course, this will force the first branch to execute, and then the third one is not tested. So, to test both the old and new branches, it might be necessary, or at least simpler, to make two test files.

The following seems to work quite well:

```vim
set nocompatible
let &rtp = '../..,' . &rtp
filetype plugin on

nnoremap q :qall!<cr>

let g:vimtex_cache_root = '.'
let g:vimtex_cache_persistent = 0

silent edit test_globbed_manual.tex

if empty($INMAKE) | finish | endif

let s:candidates = vimtex#test#completion('\cite{', '')
call vimtex#test#assert_equal(1, len(s:candidates))

bwipeout!
silent edit test_globbed_bcf.tex

let s:candidates = vimtex#test#completion('\cite{', '')
call vimtex#test#assert_equal(2, len(s:candidates))

quit!
```

```tex
% test_globbed_manual.tex
\documentclass{article}
\addbibresource{test_gl*manual.bib}
\begin{document}
Test
\end{document}
```

```tex
% test_globbed_bcf.tex
\documentclass{article}
\usepackage{biblatex}
\addbibresource{test_gl*bcf.bib}
\begin{document}
Test
\end{document}
```

And where the `*manual.bib` has one key and `*bcf.bib` has two keys. What do you think?

autoload/vimtex/bib.vim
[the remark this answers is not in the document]
TpmKranz: I know that the listing in `:help wildcards` implies that braces won't work but they do (it's easy to test). Apparently, (n)vim's glob calls [the glob function](https://man7.org/linux/man-pages/man3/globfree.3.html) with the `GLOB_BRACE` flag. 🤷🏻‍♂️

TpmKranz:
This now relies on vimtex not throwing out duplicate bibliography keys. If this ever changes, one would need to replace the symlinks with proper bib files. Copies of the original with slightly different citation keys should be enough. Of course that would open the possibility for your idea of different numbers of keys per file, maybe powers of two so one can easily see from the assertion failure message which files weren't found.

lervag:
Thanks! FYI, this does not pass tests for me. In particular, the manual route does not pass, and it seems to be something with the caches. Perhaps there's a bug in the cache mechanism, but I'm curious, why is not my much simpler version of the tests sufficient here?

TpmKranz:
> FYI, this does not pass tests for me. In particular, the manual route does not pass, and it seems to be something with the caches.
This seems to be a flakey test. Before I pushed, the test was failing for me on top of v2.4 and passing on top of this here branch, as it should. On repeat tries I can also observe failures, but not conistently. That is very strange, but I also don't know anything about vimtex's cache architecture to know if that could be the reason. I've tried inserting `VimtexReload`s and `bwipeout!`s everywhere, but couldn't get it to pass consistently.

> why is not my much simpler version of the tests sufficient here?
You mean the two tex files with their own bibs? I wanted to reduce complexity while also testing more glob cases. But if my approach confuses the cache mechanism, then I should just implement your proposal to get the test to work reliably.

lervag:
I'm not fully sure if the problem is caused by the caches, actually, but it seems like a probable cause. Still, when I looked into it, I could not find a consistent way to reproduce or avoid the issue.

> You mean the two tex files with their own bibs? I wanted to reduce complexity while also testing more glob cases. But if my approach confuses the cache mechanism, then I should just implement your proposal to get the test to work reliably.

Yes, exactly. I understand the idea of reducing complexity and allowing more tests in a more automatic fashion. However, please alos note that the way you implement the tests is much harder to "reason" about. My suggestion may be more explicit and verbose, but it is also mostly trivial to reason about and understand. In my opinion, this is a good thing. It should also be quite easy to add more similar tests based on the same simple mechanism.

PS! This is getting into a long thread on testing technicalities where your suggested patch is clearly an improvement. I don't want to "scare" you from contributing by being pedantic. So, I just want to be clear that I already much appreciate your contribution and I won't mind to just write the test myself and merge this.

TpmKranz:
I see you've credited me in the release notes for 2.5, yet I can't see any changes to `autoload/vimtex/bib.vim`.

I've reworked the test cases to be clearer and fixed another problem in the process. Hopefully, we can get this problem fixed now. :)

TpmKranz:
Sorry for the repeat force-pushes, but I wanted to have all the tests for completions that should have worked in one commit.
After looking at what `biber` actually does (call perl's `glob` function), I've noticed another missing [meta character](https://perldoc.perl.org/File::Glob#META-CHARACTERS) I hadn't checked.
I don't think backslash and tilde are relevant, since vim's file handling functions should interpret them the same way that perl's `glob` would. If not (you'll know a lot more about how vim handles these than I could), please tell me.

lervag:
> I see you've credited me in the release notes for 2.5, yet I can't see any changes to `autoload/vimtex/bib.vim`.

Huh, strange. That probably a mistake, perhaps because I noticed #2037 that was closed. Sorry about that; I hope you don't mind that I edit the release notes. You will then be properly credited in the next release notes.

> I've reworked the test cases to be clearer and fixed another problem in the process. Hopefully, we can get this problem fixed now. :)

Great! I think the tests look good, except for some reason, they are failing. Do you immediately notice why?

> Sorry for the repeat force-pushes, but I wanted to have all the tests for completions that should have worked in one commit.

No problem; I prefer logical commits anyway :)

Feel free to also rebase your PR on top of the most recent master. Of course, I'll make sure to do it anyway before I merge, but in the small chance of a merge conflict it may help to have your eyes on the conflict matter.

> After looking at what `biber` actually does (call perl's `glob` function), I've noticed another missing [meta character](https://perldoc.perl.org/File::Glob#META-CHARACTERS) I hadn't checked.
> I don't think backslash and tilde are relevant, since vim's file handling functions should interpret them the same way that perl's `glob` would. If not (you'll know a lot more about how vim handles these than I could), please tell me.

Yes, I think they work similarly.

lervag:
Tests fails with following message:

```
❯ make
Assertion failed!
Vim(let):E220: Missing }.
make: *** [Makefile:15: test_globbed_braces] Error 1
```

lervag:
It seems the problem is here:

https://github.com/TpmKranz/vimtex/blob/0967d97e72f6fcfed7f85847373cc93f5b692611/autoload/vimtex/bib.vim#L95-L100

When I print the value of `l:f` that fails, it is "test_globbed_{1", which seems to make sense. The bracket is not matched, so the glob expression is not valid. Perhaps the regex must be more advanced? `=~# '[*?]\|{.*}\|\[.*\]'?

TpmKranz:
That must be vim-specific; I have to admit that I've been testing with neovim only, which doesn't throw an error for invalid globs. I'm not a fan of pre-checking potential globs for validity as that's already being done in `glob`, the regex would be quite opaque (PCRE maybe `^((\{(?1)\}|[^{}])*)$`? No idea about a vim regex.) and it also shouldn't matter because we are only adding to a list of potential files and an invalid glob can't mean anything other than the literal filename (which is already in our list of candidates). So could we maybe just `silent!` the line with a comment to explain why?

lervag:
> That must be vim-specific;

Nope. It fails also on neovim (see the Github actions result [here](https://github.com/lervag/vimtex/pull/2035/checks?check_run_id=3036025870)).

> I'm not a fan of pre-checking potential globs for validity as that's already being done in `glob`, the regex would be quite opaque (PCRE maybe `^((\{(?1)\}|[^{}])*)$`? No idea about a vim regex.) and it also shouldn't matter because we are only adding to a list of potential files and an invalid glob can't mean anything other than the literal filename (which is already in our list of candidates). So could we maybe just `silent!` the line with a comment to explain why?

Perhaps; if that would work. I don't mind checking more candidates, but we need to be sure the output is validated properly.

lervag:
Which version of neovim are you on?

TpmKranz:
> Which version of neovim are you on?

```
NVIM v0.4.4
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=1 -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/builddir/build/BUILD/neovim-0.4.4/x86_64-redhat-linux-gnu/config -I/builddir/build/BUILD/neovim-0.4.4/src -I/usr/include -I/usr/include/lua-5.1 -I/builddir/build/BUILD/neovim-0.4.4/x86_64-redhat-linux-gnu/src/nvim/auto -I/builddir/build/BUILD/neovim-0.4.4/x86_64-redhat-linux-gnu/include
Compiled by mockbuild@koji

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

clason:
Can you try the latest release (v0.5.0)?

TpmKranz:
> Can you try the latest release (v0.5.0)?

Fedora 34 doesn't yet carry that version and I'm not inclined to compile it myself, especially when the version used in the failed Github Action is the same as mine and it's therefore questionable whether the version itself is the problem.

For some reason, my neovim is unable to complete another test, even with `MYVIM="nvim --headless -u NONE"`:
```
make -C test-indentation
16c16
<       & 2 \\
---
>     & 2 \\
18c18
<       & 4 \\
---
>     & 4 \\
make[1]: *** [Makefile:15: test_ampersands_reference] Error 1
make: *** [Makefile:18: test-indentation] Error 2
```

Dunno why that is, maybe I've just got an entirely broken installation. Maybe I should just do my tests in a container?

lervag:
> For some reason, my neovim is unable to complete another test, even with `MYVIM="nvim --headless -u NONE"`:

You should not need to set the `MYVIM`; the defaul is `nvim --headless`. No need to add `-u NONE`, I think.

The tests pass on master both on my systems and on Github Actions, so this does seems somewhat strange. :\

lervag:
I've think I've found a reasonable solution. Let me know what you think (see my extra commit).

TpmKranz:
Yeah, that'll do. I didn't know that vim could do error catching. A bit disappointing that my PR couldn't be merged but very glad that the problem's been fixed to your satisfaction now!
Thanks for all the work you put into this project and your patience with my PR; I very much appreciate it.

TpmKranz:
Oh sorry, I'm just now seeing that my commits have, in fact, been merged. No reason for me to be disappointed, then. :)

lervag:
> Yeah, that'll do. I didn't know that vim could do error catching.

Yes, as I often say, Vimscript is not as bad as many people think. But it is peculiar at times. ;)

> A bit disappointing that my PR couldn't be merged but very glad that the problem's been fixed to your satisfaction now!
> Thanks for all the work you put into this project and your patience with my PR; I very much appreciate it.

> Oh sorry, I'm just now seeing that my commits have, in fact, been merged. No reason for me to be disappointed, then. :)

Yes, sorry if that was unclear. I rebased your commits on top of master, then added a minor commit where I made some changes that I believe were minor improvements. I try to keep the contributions from others clear and visible; I very much appreciate when others want to help and contribute!

And thanks for the kind words!

lervag:
By the way; the tests still fail. They pass on my system, so it must be a version or system related issue. Do you have any clue what might be wrong?

https://github.com/lervag/vimtex/runs/3042079001?check_suite_focus=true

PR #2037, opened 2021-04-21
Fix test_globbed.vim to detect shortcomings of the implementation

TpmKranz:
This fixes the test to take both code paths where globbing is needed (cf. #2035).

The test used to always take the second path because vimtex doesn't know about the biblatex package being loaded before a compile.

This doesn't need to be merged because it will also be part of #2035, I just want to illustrate why that bug wasn't uncovered by the existing test and that it indeed fails if done properly.

I would also like to discuss a way to test more than one .tex file within the same vimscript, but am not experienced enough in vimscript to do that with the current asynchronous implementation. The following happens when I want to edit another file and compile it:
```
Error detected while processing function TestCompletion[12]..vimtex#compiler#compile:
line    1:
E121: Undefined variable: b:vimtex
```
Maybe it would be better to do it synchronously, but I also haven't found out how to wait for an event.

### DIFF
--- a/test/test-completion-bibtex/test_globbed.vim
+++ b/test/test-completion-bibtex/test_globbed.vim
@@ -4,15 +4,39 @@ filetype plugin on
 
 nnoremap q :qall!<cr>
 
+let s:build_dir = expand('<sfile>:r') . '.dir'
 let g:vimtex_cache_root = '.'
 let g:vimtex_cache_persistent = 0
+let g:vimtex_log_verbose = 0
+let g:vimtex_compiler_method = 'latexmk'
+let g:vimtex_compiler_latexmk = {
+      \ 'callback' : 0,
+      \ 'continuous' : 0,
+      \ 'build_dir' : s:build_dir,
+      \}
+
+
+function! Clean()
+  if isdirectory(s:build_dir)
+    call delete(s:build_dir, 'rf')
+  endif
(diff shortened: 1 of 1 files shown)


lervag:
> This fixes the test to take both code paths where globbing is needed (cf. #2035).
> 
> The test used to always take the second path because vimtex doesn't know about the biblatex package being loaded before a compile.
> 
> This doesn't need to be merged because it will also be part of #2035, I just want to illustrate why that bug wasn't uncovered by the existing test and that it indeed fails if done properly.

Thanks! I'll close this PR, then, but we can continue the following discussion. However, I'll prefer to comment and discuss the test code specifically in #2035.

> I would also like to discuss a way to test more than one .tex file within the same vimscript, but am not experienced enough in vimscript to do that with the current asynchronous implementation. The following happens when I want to edit another file and compile it:
> 
> ```
> Error detected while processing function TestCompletion[12]..vimtex#compiler#compile:
> line    1:
> E121: Undefined variable: b:vimtex
> ```
> 
> Maybe it would be better to do it synchronously, but I also haven't found out how to wait for an event.

I am not quite sure what you are really trying here. Regarding asynchronicity, the vimscript based tests can only run synchronous. That is, vimscript is itself limited in this regard. We can run `make` in multiple threads, and I partly do that already.

Issue #2065, opened 2021-06-07
fold text in bib files?


yongrenjie:
**Is your feature request related to a problem? Please describe it.**

Setting `foldmethod=syntax` in bib files (provided by vim8.2's built in syntax file) gives nice folding, but the default fold text is a bit dull. For example (stealing a paper from my supervisor :-))

```
@article{Foroozandeh2014ACIE,
    doi = {10.1002/anie.201404111},
    author = {Foroozandeh, Mohammadali and Adams, Ralph W. and Meharry, Nicola J. and Jeannerat, Damien and Nilsson, Mathias and Morris, Gareth A.},
    title = {Ultrahigh-Resolution NMR Spectroscopy},
    journaltitle = {Angew.\ Chem.\ Int.\ Ed.},
    year = {2014},
    volume = {53},
    number = {27},
    pages = {6990--6992},
}
```

produces the following foldtext

```
+-- 10 lines: @article{Foroozandeh2014ACIE,-----------------------------------------------
```

IMO, it's not useful to know how many lines an entry is; also, the remainder of the space could be more usefully utilised to show some information about the bib entry.

**Describe the solution you'd like**

For most bib entries I think the most useful piece of extra information would be the title (?) or for `@set{...}`, it would be useful to see which entries are contained within the set.

I actually have an implementation of the foldtext function here: https://github.com/yongrenjie/abbotsbury.vim/blob/master/ftplugin/bib.vim which I use personally. For the same bib entry above it produces the following foldtext:

```
+-- article{Foroozandeh2014ACIE} -- Ultrahigh-Resolution NMR Spectroscopy ----------------
```

**Describe alternatives you've considered**

I mainly opened this issue as I was curious as to whether this would be something you might be interested in including in vimtex. I'm happy to modify the output slightly if desired!

The main alternative would be sticking with the default, I guess. There might be an argument that long fold texts lead to visual clutter.

lervag:
> I mainly opened this issue as I was curious as to whether this would be something you might be interested in including in vimtex. I'm happy to modify the output slightly if desired!

I don't quite see why not to add this. I think improving the fold text for bib files should generally be a good thing for most people, and it seems relevant to VimTeX. If it would be opinionated, then we could easily make it optional.

Do you want to open a PR for this? I could add this based on your implementation myself, but if you feel adventurous and want to keep your name on the contribution, then I wouldn't mind helping you with a PR.

yongrenjie:
[...]

PR #2066, opened 2021-06-09
[WIP] add folding for bib files

yongrenjie:
Since opening the issue (#2065) I've tweaked the output a bit, here's a sample (first one unfolded for comparison):

Screenshot 2021-06-09 at 9 05 06 PM

### DIFF
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -100,8 +100,11 @@ function! s:init_buffer() abort " {{{1
           \]
 
     if g:vimtex_fold_bib_enabled
-      let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
-      autocmd BufWrite * let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
+      let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
+      augroup vimtex_buffers
+        autocmd BufWrite <buffer>
+              \ let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
+      augroup END
       setlocal foldmethod=syntax
       setlocal foldtext=vimtex#fold#bib#text()
     endif
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -105,7 +105,8 @@ function! s:init_buffer() abort " {{{1
         autocmd BufWrite <buffer>
               \ let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
       augroup END
-      setlocal foldmethod=syntax
+      setlocal foldmethod=expr
+      setlocal foldexpr=vimtex#fold#bib#level(v:lnum)
       setlocal foldtext=vimtex#fold#bib#text()
     endif
   endif
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -98,6 +98,11 @@ function! s:init_buffer() abort " {{{1
           \ 'fold', 'matchparen', 'format', 'doc', 'imaps', 'cmd', 'delim',
           \ 'env', 'motion', 'complete',
           \]
+
+    if g:vimtex_fold_bib_enabled
+      setlocal foldmethod=syntax
+      setlocal foldtext=vimtex#fold#bib#text()
+    endif
   endif
 
   " Initialize buffer settings for sub modules
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -100,14 +100,7 @@ function! s:init_buffer() abort " {{{1
           \]
 
     if g:vimtex_fold_bib_enabled
-      let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
-      augroup vimtex_buffers
-        autocmd BufWrite <buffer>
-              \ let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
-      augroup END
-      setlocal foldmethod=expr
-      setlocal foldexpr=vimtex#fold#bib#level(v:lnum)
-      setlocal foldtext=vimtex#fold#bib#text()
+      call vimtex#fold#bib#init()
     endif
   endif
 
--- a/autoload/vimtex.vim
+++ b/autoload/vimtex.vim
@@ -100,6 +100,8 @@ function! s:init_buffer() abort " {{{1
           \]
 
     if g:vimtex_fold_bib_enabled
+      let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
+      autocmd BufWrite * let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
       setlocal foldmethod=syntax
       setlocal foldtext=vimtex#fold#bib#text()
     endif
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -19,47 +19,42 @@ endfunction
 
 function! vimtex#fold#bib#level(lnum) abort " {{{1
   " Handle blank lines
-  if getline(a:lnum) =~ '\v^\s*$'
-    if a:lnum == 1
+  if empty(trim(getline(a:lnum)))
+    let l:prev_level = vimtex#fold#bib#level(a:lnum - 1)
+
+    if l:prev_level == '<1' || l:prev_level == 0
       return 0
-    else
-      let l:prev_foldlevel = vimtex#fold#bib#level(a:lnum - 1)
-      if l:prev_foldlevel == '0'
+    elseif l:prev_level == '>1'
+      " Check if previous line is a standalone entry
+      let l:prev_line = getline(a:lnum - 1)
+      if trim(l:prev_line)[0] == '@' 
+            \ && s:count(l:prev_line, '{') == s:count(l:prev_line, '}')
         return 0
-      elseif l:prev_foldlevel == '>1'
-        let l:prev_line = getline(a:lnum - 1)
-        if s:count(l:prev_line, '{') == s:count(l:prev_line, '}')
-          return 0
-        else
-          return '='
-        endif
-      else
-        return '='
       endif
     endif
+
+    return 1
   endif
 
   " Search for the beginning of the entry
-  let l:firstline = a:lnum
-  while l:firstline >= 1
-    if getline(l:firstline) =~ '\v^\s*\@'
-      break
-    endif
-    let l:firstline -= 1
-  endwhile
+  let l:curpos = getcurpos()
+  call cursor(a:lnum, 0)
+  let l:firstline = search('\v^\s*\@', 'bcnW')
+  call setpos('.', l:curpos)
 
   if l:firstline == a:lnum
     return '>1'
-  elseif l:firstline == 0   " beginning of entry wasn't found
+  elseif l:firstline == 0
+    " Beginning of entry wasn't found
     return 0
   endif
 
   " Check if braces are closed by the current line
-  let l:text = join(map(range(l:firstline, a:lnum), 'getline(v:val)'))
+  let l:text = join(getline(l:firstline, a:lnum))
   if s:count(l:text, '{') == s:count(l:text, '}')
     return '<1'
   else
-    return '1'
+    return 1
   endif
 
   return 0
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -1,0 +1,67 @@
+" VimTeX - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+"
+
+function! vimtex#fold#bib#text() abort " {{{1
+  " Note that the actual folding is handled using foldmethod=syntax; this
+  " merely generates the fold text.
+  let l:entry_type = ''   " e.g. article, book, thesis, ...
+  let l:key = ''          " the cite key
+  let l:description = ''  " title of the work, or entryset for sets
+
+  " Search the first line of the fold for the entry type and the key.
+  let l:entry_type_key_match = matchlist(getline(v:foldstart),
+        \ '\v\@(\S+)\s*\{\s*%((\S+)\s*,)?')
+  if !empty(l:entry_type_key_match)
+    let l:entry_type = l:entry_type_key_match[1]
+    let l:key = l:entry_type_key_match[2]
+    " If the key wasn't found, then check the second line of the fold to see
+    " if it's there.
+    if empty(l:key)
+      let l:nextline_match = matchlist(getline(v:foldstart + 1),
+            \ '\v^\s*(\S+)\s*,')
+      if !empty(l:nextline_match)
+        let l:key = l:nextline_match[1]
+      endif
+    endif
+  endif
+
+  " Search inside the fold for the title of the work (or the entryset, if it's
+  " a set)
+  let l:description_pattern = l:entry_type == 'set' ? 
+        \ '\v^\s*entryset\s*\=\s*(\{.+\}|\".+\")\s*,?' :
+        \ '\v^\s*title\s*\=\s*(\{.+\}|\".+\")\s*,?'
+  let l:lnum = v:foldstart
+  while l:lnum <= v:foldend
+    let l:description_match = matchlist(getline(l:lnum), l:description_pattern)
+    if l:description_match != []
+      " Remove surrounding braces or quotes
+      let l:description = l:description_match[1][1:-2]
+      break
+    else
+      let l:lnum += 1
+    endif
+  endwhile
+
+  " Construct the fold text
+  if !empty(l:entry_type) && !empty(l:key)
+    let l:foldtext = '@' . l:entry_type . '{' . l:key . '}'
+    " Pad until the next 'tab' of 4 spaces, before adding in the title. 32
+    " just seems like a sensible lower bound to accommodate cite keys of
+    " typical lengths.
+    let l:width = strdisplaywidth(l:foldtext)
+    let l:desired_width = max([32, l:width + 4 - (l:width % 4)])
+    let l:foldtext .= repeat(' ', l:desired_width - l:width)
+    if !empty(l:description)
+      let l:foldtext .= l:description
+    endif
+    return l:foldtext
+  else
+    " The entry type or key wasn't found; return the default fold text.
+    return foldtext()
+  endif
+endfunction
+
+" }}}1
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -52,7 +52,7 @@ function! vimtex#fold#bib#text() abort " {{{1
     " just seems like a sensible lower bound to accommodate cite keys of
     " typical lengths.
     let l:width = strdisplaywidth(l:foldtext)
-    let l:desired_width = max([32, l:width + 4 - (l:width % 4)])
+    let l:desired_width = max([b:vimtex_fold_bib_maxwidth, l:width]) + 2
     let l:foldtext .= repeat(' ', l:desired_width - l:width)
     if !empty(l:description)
       let l:foldtext .= l:description
@@ -62,6 +62,43 @@ function! vimtex#fold#bib#text() abort " {{{1
     " The entry type or key wasn't found; return the default fold text.
     return foldtext()
   endif
+endfunction
+
+" }}}1
+
+function! vimtex#fold#bib#get_max_key_width() " {{{1
+  let start_time = reltime()
+  let l:lnum = 1
+  let l:max_width = 0
+  while l:lnum <= line('$')
+    if trim(getline(l:lnum))[0] == '@'
+      let l:entry_type = ''
+      let l:key = ''
+      let l:entry_type_key_match = matchlist(getline(l:lnum),
+            \ '\v\@(\S+)\s*\{\s*%((\S+)\s*,)?')
+      if !empty(l:entry_type_key_match)
+        let l:entry_type = l:entry_type_key_match[1]
+        let l:key = l:entry_type_key_match[2]
+        " If the key wasn't found, then check the second line of the fold to see
+        " if it's there.
+        if !empty(l:entry_type) && empty(l:key)
+          let l:nextline_match = matchlist(getline(l:lnum + 1),
+                \ '\v^\s*(\S+)\s*,')
+          if !empty(l:nextline_match)
+            let l:key = l:nextline_match[1]
+          endif
+        endif
+      endif
+      if !empty(l:entry_type) && !empty(l:key)
+        " the 3 extra characters includes the @ and the { }.
+        let l:width = 3 + strdisplaywidth(l:entry_type) + strdisplaywidth(l:key)
+        let l:max_width = max([l:max_width, l:width])
+      endif
+    endif
+    let l:lnum += 1
+  endwhile
+  echo "elapsed time:" reltimestr(reltime(start_time))
+  return l:max_width
 endfunction
 
 " }}}1
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -4,6 +4,56 @@
 " Email:      karl.yngve@gmail.com
 "
 
+function! vimtex#fold#bib#level(lnum) abort " {{{1
+  " Handle blank lines
+  if getline(a:lnum) =~ '\v^\s*$'
+    if a:lnum == 1
+      return 0
+    else
+      let l:prev_foldlevel = vimtex#fold#bib#level(a:lnum - 1)
+      if l:prev_foldlevel == '0'
+        return 0
+      elseif l:prev_foldlevel == '>1'
+        let l:prev_line = getline(a:lnum - 1)
+        if s:count(l:prev_line, '{') == s:count(l:prev_line, '}')
+          return 0
+        else
+          return '='
+        endif
+      else
+        return '='
+      endif
+    endif
+  endif
+
+  " Search for the beginning of the entry
+  let l:firstline = a:lnum
+  while l:firstline >= 1
+    if getline(l:firstline) =~ '\v^\s*\@'
+      break
+    endif
+    let l:firstline -= 1
+  endwhile
+
+  if l:firstline == a:lnum
+    return '>1'
+  elseif l:firstline == 0   " beginning of entry wasn't found
+    return 0
+  endif
+
+  " Check if braces are closed by the current line
+  let l:text = join(map(range(l:firstline, a:lnum), 'getline(v:val)'))
+  if s:count(l:text, '{') == s:count(l:text, '}')
+    return '<1'
+  else
+    return '1'
+  endif
+
+  return 0
+endfunction
+
+" }}}1
+
 function! vimtex#fold#bib#text() abort " {{{1
   let l:bib_entries = vimtex#parser#bib#parse_cheap(v:foldstart, v:foldend, {})
   if len(l:bib_entries) != 1
@@ -45,6 +95,19 @@ function! vimtex#fold#bib#get_max_key_width() " {{{1
         \ + strdisplaywidth(e.type)
         \ + strdisplaywidth(e.key)})
   return max(l:entries)
+endfunction
+
+" }}}1
+
+function! s:count(container, item) abort " {{{1
+  " Necessary because in old Vim versions, count() does not work for strings
+  try
+    let l:count = count(a:container, a:item)
+  catch /E712/
+    let l:count = count(split(a:container, '\zs'), a:item)
+  endtry
+
+  return l:count
 endfunction
 
 " }}}1
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -14,8 +14,11 @@ function! vimtex#fold#bib#text() abort " {{{1
     if !empty(l:entry.type) && !empty(l:entry.key)
       let l:foldtext = '@' . l:entry.type . '{' . l:entry.key . '}'
       let l:width = strdisplaywidth(l:foldtext)
-      " TODO: implement g:vimtex_fold_bib_max_key_width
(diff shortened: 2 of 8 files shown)


lervag:
Great, thanks! I'll review as soon as I get the chance. One thing I notice immediately: Should we shorten the first part of the foldtext if it becomes too long, e.g. in the lines `SchulzeSunninghausen2014*` in your screenshot?

yongrenjie:
That's something I was not entirely sure about, too. On one hand it _would_ looks nicer if it's trimmed. On the other hand I think the cite key is the most important bit of info. In this case, if we trim the length to 32 or fewer chars, then the user can't differentiate the keys `...2014JACS` and `...2017JMR`, and I think that's why I implemented it the way I did. But I don't think I'm hugely opinionated either way.

I experimented a bit with calculating the length of the longest key when opening the buffer, and setting it to a `b:vimtex_fold_bib_maxwidth` variable. Basically, it uses the same code as in the foldtext function to parse the bib file for cite keys. For an 837-line bib file this takes about 22 ms (on my computer, at least), and it scales linearly with the length of the file.

yongrenjie:
I pushed a second commit which does this, it's just a draft to see what you think about the behaviour, which I can clean up later!

Here's a short video I recorded. The buffer variable is recalculated when opening and also when saving, hence the 'jumps' when I save:

https://user-images.githubusercontent.com/22414895/121611169-940c6700-ca4f-11eb-8fc8-dd824de7ce4a.mp4

lervag:
> That's something I was not entirely sure about, too. On one hand it _would_ looks nicer if it's trimmed. On the other hand I think the cite key is the most important bit of info.

Yes, agreed. Good point.

> I experimented a bit with calculating the length of the longest key when opening the buffer, and setting it to a `b:vimtex_fold_bib_maxwidth` variable. Basically, it uses the same code as in the foldtext function to parse the bib file for cite keys.

Yes, that sounds good. I think having an option as well, e.g. `g:vimtex_fold_bib_max_key_width`, may be useful. To allow to restrict it, in case someone tends to have a few very long keys.

> For an 837-line bib file this takes about 22 ms (on my computer, at least), and it scales linearly with the length of the file.

So, in the insane case you have ~10000 lines, it would still only be about ~250 ms, which should be acceptable for most people. Let's say they have a slow computer, then it could be as bad as 1 second. Still acceptable if this is only when the file is initially opened.

autoload/vimtex.vim
```diff
@@ -89,6 +89,13 @@ function! s:init_buffer() abort " {{{1
           \ 'fold', 'matchparen', 'format', 'doc', 'imaps', 'cmd', 'delim',
           \ 'env', 'motion', 'complete',
           \]
+
+    if g:vimtex_fold_bib_enabled
+      let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
```
lervag, line 94: I prefer to keep spaces on either side of `=`.

autoload/vimtex.vim
```diff
@@ -89,6 +89,13 @@ function! s:init_buffer() abort " {{{1
...
+      autocmd BufWrite * let b:vimtex_fold_bib_maxwidth=vimtex#fold#bib#get_max_key_width()
```
lervag, line 95: This autocmd must be changed. I propose something like this:

```vim
augroup vimtex_buffers
  autocmd BufWrite <buffer>
        \ let b:vimtex_fold_bib_maxwidth = vimtex#fold#bib#get_max_key_width()
augroup END
```

There are two main points: i) autocmds should be included in a group, and ii) the autocmd should only be applied to the current buffer.

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,104 @@
+" VimTeX - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+
```
lervag, line 5: This is probably just silly, but I believe almost every single file has a `"` at the beginning of line 5 as well. And I like things to be consistent... :)

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,104 @@
...
+
+function! vimtex#fold#bib#text() abort " {{{1
+  " Note that the actual folding is handled using foldmethod=syntax; this
+  " merely generates the fold text.
```
lervag, line 9: I don't think this comment is necessary. If people don't know what a foldtext function is when reading this, then I think they're in deep water already...

I do like having comments to help make the code more understandable, but I also prefer to have as few comments as possible. (But not fewer.)

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,104 @@
...
+  let l:entry_type = ''   " e.g. article, book, thesis, ...
+  let l:key = ''          " the cite key
+  let l:description = ''  " title of the work, or entryset for sets
+
+  " Search the first line of the fold for the entry type and the key.
```
lervag, line 14: I'm curious: I've actually implemented a bib file parser, see here:

https://github.com/lervag/vimtex/blob/82fbb79f7bd1ba54ad61da73f83d71a9aa23febd/autoload/vimtex/parser/bib.vim#L169

Now, the following code is probably doing something of the same. Perhaps it would be possible to somehow make some minor modifications of the existing parser and then rely on it instead? (E.g., allow to pass it a list of lines instead of a filename.) This is not important, and I can consider to look into it myself. An immediate concern would be if it adds a noticable overhead (full parsing is more expensive than the dedicated parsing you do below).

lervag (changes requested):
I did a review of the code as well, and I am very glad to see that this is high quality content! Well done!

One "major" discussion point: should we somehow merge some of the bulk code here with the code under `autoload/vimtex/parser/bib.vim`? That could help make it useful for other related features in the future. Note, I won't block merging because of this, I can also simply consider to do that myself at some time. But hearing other peoples opinions are always interesting.

autoload/vimtex.vim
yongrenjie: OK! I was not all that sure whether it was a good idea to update on saving. The delay is probably not noticeable so I like having it. Do you think we should keep it (and by that I mean modify it as per your suggestion?)

autoload/vimtex/fold/bib.vim
yongrenjie: I thought I put it in by accident with `o` 😅

autoload/vimtex/fold/bib.vim
yongrenjie: I took a quick look at it.

```vim
function! vimtex#fold#bib#get_max_key_width_full_parse() " {{{1
  function! s:get_key_width(idx, val)
    return 3 + strdisplaywidth(a:val['type']) + strdisplaywidth(a:val['key'])
  endfunction
  let start_time = reltime()
  let l:file = expand('%:p')
  let l:data = vimtex#parser#bib#parse(l:file, {'backend': 'vim'})
  call map(l:data, function('s:get_key_width'))
  echomsg "elapsed time:" reltimestr(reltime(start_time))
  return max(l:data)
endfunction
```

This returns the same result but takes about 130 ms for the same bib file (837 lines) on the same computer. So not actually too bad but might be noticeable on larger files.

yongrenjie:
I must admit I've not looked at the bib parsing code much, and I'm going to be a little busy these few days, but will come back to it (and make the changes above) when I can!

As for moving the code inside `autoload/vimtex/parser/bib.vim`... I'm not sure about the extent to which it can be fully "integrated".

If we modify `function! s:parse_with_vim()`, then but the function signature would have to change in a couple of ways:

 1. to either perform full parsing (the existing function), or cheap parsing (which does only entry type and key). Chances are we have to keep it separate for performance reasons - we don't want to do full parsing for the folds.
 2. to accept either filename or a list of lines

That would introduce some kind of asymmetry between the bibtex/bibparse and vim parsers, which seems to be a bit unsatisfying to me.

The alternative would be to keep the parsing function separate, lets say write a `s:cheap_parse_with_vim()`, and to make use of the underlying parser 'primitives' e.g. `s:get_value()`. I don't think I can look at this properly in the next few days. But my impression was that they are all interlinked (since they pass dictionaries to each other), and trying to use just one of them and not the rest might be a bit tricky.

All in all I'm more inclined to keep it separate, mainly because I don't see the best way to merge the code, but I would defer to you on that 🙂

One benefit of utilising the existing parser (if it's feasible, and not as hard as I've claimed) might be to deal with edge cases. I tried to deal with a reasonable number of cases, but I know the foldtext function will break with pathological input like `title = "something" # {another thing},` (the regex will only capture `something`).

On the topic of edge cases: is it worth adding some tests for the above as well?

autoload/vimtex.vim
lervag: > Do you think we should keep it (and by that I mean modify it as per your suggestion?)

Good question. I don't see any issues with keeping it if the cost is so low. But I would prefer to avoid options, and sometimes simple is simply better.

I would suggest we keep the update on saving under the hypothesis that "it will just work". If there should come counter evidence, then I would not really mind either fixing a bug or removing this part and simplifying.

autoload/vimtex/fold/bib.vim
lervag: No problem. :)

autoload/vimtex/fold/bib.vim
lervag: Does it scale similarly? The complexity might give > O(n) cost. If that's the case, then I certainly would not want to do this.

lervag:
> I must admit I've not looked at the bib parsing code much, and I'm going to be a little busy these few days, but will come back to it (and make the changes above) when I can!

No problem, and no rush!

> As for moving the code inside `autoload/vimtex/parser/bib.vim`... I'm not sure about the extent to which it can be fully "integrated".
> ‥‥

Let's first assume we don't need to care about the performance. In this case, I think the only real change would be to allow a list of lines as an alternative input. It would bring an asymmetry, unless we also update the other alternatives to work with this kind of input (e.g. we could save the lines to a file first).

> The alternative would be to keep the parsing function separate, lets say write a `s:cheap_parse_with_vim()`, and to make use of the underlying parser 'primitives' e.g. `s:get_value()`. I don't think I can look at this properly in the next few days. But my impression was that they are all interlinked (since they pass dictionaries to each other), and trying to use just one of them and not the rest might be a bit tricky.

If performance is important (and it may very well be!), then I think the idea would be to implement the parser as a full API function with signature `vimtex#parser#bib#cheap_parse(lines)`. I think this might be a good idea as a fallback solution, because it would make it more likely that the code could be reused and improved in future work.

I do agree that adding such an extra function does seem slightly "off", but I think this is one of those cases where some comments may help our future selves (and others) understand the reasons we decided this (or that) way.

> On the topic of edge cases: is it worth adding some tests for the above as well?

Yes, definitely! I think it could be quite simple, actually:

1. Create a new test folder `test-folding-bib`.

2. Either rely on the existing `common/huge.bib` or similar, or create a new `test.bib` inside the test folder.

3. Create a `test.vim` file that opens a `bib` file. Ensure all folds are closed. Then use some `assert_equal`s based on results of `foldtextresult(lnum)`.

I think the `test-utils` may be a good example of how I write the Makefiles and test scripts, if you want to give it a shot.

lervag:
By the way; I realize there's a choice here that is undecided and sort of hard to decide. My suggestion would be to go for the cheap parser, but to move the main parsing code to e.g. `vimtex#parser#bib#cheap_parse(lines)`. I think it would be good, and not too hard, to allow it to work well both on any number of lines that contain one or more entries. It should return a list of parsed results. This would allow to use the same function both in the foldtext function and in the `get_max` function.

yongrenjie:
Ok, finally free from real life work for a bit 😅

I agree with your assessment re. implementing `vimtex#parser#bib#cheap_parse`; I think for me performance would be the main reason. I'll work on that main code first, then can write some tests in the coming days, since I probably need a little bit of time to figure out how the tests are laid out.

autoload/vimtex/fold/bib.vim
yongrenjie: It seems to scale as O(n): 122 ms, 253 ms, 470 ms, 938 ms for four files which are successively doubled in size.

yongrenjie: The refactored code that I'll push later is much faster (even faster than my initial PR, I don't know what I did correctly this time!!): 8 ms, 15 ms, 28 ms, 56 ms for the same four files as above, which is ca. 15x faster.

yongrenjie:
I think the actual code is done. Here is how it looks with the default `g:vimtex_fold_bib_max_key_width=0`

Screenshot 2021-06-21 at 8 10 02 PM

and with `g:vimtex_fold_bib_max_key_width=20`

Screenshot 2021-06-21 at 8 10 10 PM

I added some docs as well, and will look into tests. But one more thing regarding edge cases: I'm pretty certain that the most brittle part of the current setup is actually `foldmethod=syntax`. For example, the following snippet

```bib
@string{ foo = "Mrs. Foo" }

@article{Kupce2017ACIE,
    doi = {10.1002/anie.201705506},
    author = {Kup{\v{c}}e, {\=E}riks and Claridge, Tim D.\ W.},
    journal = {Angew.\ Chem.\ Int.\ Ed.},
    title = {{NOAH}: {NMR} Supersequences for Small Molecule Analysis and Structure Elucidation},
    year = {2017},
    volume = {56},
    issue = {39},
    pages = {11779--11783},
}
```

produces a single fold for all 12 lines, instead of (probably) the more natural result of `@string` line being left on its own and the `@article` being folded. So it seems like it might be sensible to write a `foldexpr` as well, and I'd wager that some of the existing code could be used for that as well... I'll see what I can come up with.

lervag:
> I think the actual code is done. Here is how it looks ...

Great, this looks good! One possibly good idea could be to parse the titlestring and remove the brackets, as that would allow to show more of the title text. But that's clearly not very important, of course.

> I added some docs as well, and will look into tests.

Great!

> But one more thing regarding edge cases: I'm pretty certain that the most brittle part of the current setup is actually `foldmethod=syntax`. For example, ...

Ah, yes. I've generally found `fdm=syntax` to be quite brittle, and I agree it could be simple to write a `foldexpr` for bib files. So that might be a good improvement here! (However, for a robust fold expression, you would probably need to match/count the braces. People who are "messy" and careless may put their braces at unexpected places. :p --- I do this in the TeX fold code, and I can probably help if you want.)

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,54 @@
...
+" }}}1
+
+function! vimtex#fold#bib#get_max_key_width() " {{{1
+  if g:vimtex_fold_bib_max_key_width > 0
+    return g:vimtex_fold_bib_max_key_width
+  endif
+
+  let l:entries = vimtex#parser#bib#cheap_parse(1, line('$'),
+        \ {'get_description': v:false})
+  call map(l:entries, 's:get_single_key_width(v:val)')
```
lervag, line 43: I think the code could be more condense and still readable, e.g.:

```vim
  " Extra 3 is for the @ symbol plus curly braces
  call map(l:entries, {_, e -> 3
        \ + strdisplaywidth(x.entry_type)
        \ + strdisplaywidth(x.key)})
```

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,54 @@
...
+"
+
+function! vimtex#fold#bib#text() abort " {{{1
+  let l:bib_entries = vimtex#parser#bib#cheap_parse(v:foldstart, v:foldend, {})
+  if len(l:bib_entries) != 1
+    return foldtext()
+  else
+    let l:entry = l:bib_entries[0]
+
+    if !empty(l:entry.entry_type) && !empty(l:entry.key)
```
lervag, line 14: Could we call it simply "type" instead of "entry_type"? That would also better align with the similar name from the "full" parser.

autoload/vimtex/parser/bib.vim
```diff
@@ -22,6 +22,65 @@ endfunction
 
 " }}}1
 
+function! vimtex#parser#bib#cheap_parse(start_line, end_line, opts) abort " {{{1
```
lervag, line 25: I tend to favor names where the "category" comes first, e.g. `parse_cheap`, because it makes it easier to notice the difference between similarly named functions. Feel free to disagree :p

lervag (changes requested):
I understand you still want to make some more changes. Still, here are some minor comments to the current version. In general, I'm very happy to see that you are suggesting very high quality work (in my opinion!). Thanks!

autoload/vimtex/fold/bib.vim
yongrenjie: Aha, I tried with a `lambda` just now, but didn't realise that the value is the second argument and not the first. Still learning...

yongrenjie:
@lervag Sorry about all the force pushes/rebasing! I think I'm done for now, though.

I wrote a foldexpr in `vimtex#fold#bib#level(lnum)` which seems to be fairly robust towards badly positioned curly braces (the test contains a pretty good demonstration IMO). It's slow, no surprises I guess!

| File | Number of lines | foldtext (ms) | foldexpr (ms) |
| -- | -- | -- | -- |
| hsqc1.bib | 845 | 10 | 86 | 
| hsqc2.bib | 1690 | 20 | 161 |
| hsqc4.bib | 3380 | 38 | 300 |
| hsqc8.bib | 6760 | 70 | 592 |
| hsqc16.bib | 13520 | 122 | 1165 |

Here `foldtext` refers to the fime needed to evaluate `vimtex#fold#bib#get_max_key_width()`, and `foldexpr` refers to the time needed to evaluate `vimtex#fold#bib#level(lnum)` on every line of the file. (`hsqc1.bib` is this file [here](https://github.com/yongrenjie/hsqc-paper/blob/master/final_submission/source_files/hsqc.bib); the rest are just hsqc1.bib concatenated with itself as many times as necessary.)

With the current code in the PR, `hsqc8.bib` takes 855 ms to startup (this includes various other plugins which I didn't bother disabling), whereas if I revert to `foldmethod=syntax` (but keeping the foldtext functions) this drops to 446 ms. So I wonder whether one option to 'fall back' to `foldmethod=syntax`, but keep the foldtext functionality, might be useful. Or maybe I'm just worrying too much about people who keep crazy 10k-line bib files. :-)

There are also some tests now, and I changed whatever was necessary in the docs (which turned out to be very little).

Let me know what you think / if you want me to change anything, whenever you are free, of course!

lervag:
Thanks! And sorry for the delay, I'll try and get time to review and possibly merge tomorrow or the day after!

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,116 @@
...
+  while l:firstline >= 1
+    if getline(l:firstline) =~ '\v^\s*\@'
+      break
+    endif
+    let l:firstline -= 1
+  endwhile
+
+  if l:firstline == a:lnum
+    return '>1'
+  elseif l:firstline == 0   " beginning of entry wasn't found
```
lervag, line 40: I prefer to keep comments on separate lines; in this case move to next line, indented.

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,116 @@
...
+      return l:foldtext
+    else
+      return foldtext()
+    endif
+  endif
+endfunction
+
+" }}}1
+
+function! vimtex#fold#bib#get_max_key_width() " {{{1
```
lervag, line 86: If we move the init code to `vimtex#fold#bib#init`, then we can degrade this function to `s:get_max_key_width()` and make it private.

lervag (commented):
> @lervag Sorry about all the force pushes/rebasing! I think I'm done for now, though.

No need to apologize; I really do prefer the force pushing (although I don't enforce it). :)

> I wrote a foldexpr in `vimtex#fold#bib#level(lnum)` which seems to be fairly robust towards badly positioned curly braces (the test contains a pretty good demonstration IMO). It's slow, no surprises I guess!

Yes, it's a known "issue" with `foldexpr` and the reason why plugins like Fastfold exists. Thanks for looking into the timings. I could make the same tests on my machine, but your bib file seems to be unavailable (perhaps private?).

> With the current code in the PR, `hsqc8.bib` takes 855 ms to startup (this includes various other plugins which I didn't bother disabling), whereas if I revert to `foldmethod=syntax` (but keeping the foldtext functions) this drops to 446 ms. So I wonder whether one option to 'fall back' to `foldmethod=syntax`, but keep the foldtext functionality, might be useful. Or maybe I'm just worrying too much about people who keep crazy 10k-line bib files. :-)

I won't mind bothering users a little bit with pushing this and then waiting for the reactions. I'll have a deeper look into the foldexpr function after merging to see if I can spot some optimizations (I did not see anything obvious, except perhaps a minor improvement with the `foldlevel()` thing.

> There are also some tests now, and I changed whatever was necessary in the docs (which turned out to be very little).

Great!

> Let me know what you think / if you want me to change anything, whenever you are free, of course!

I had a few minor comments/suggestions. After that, I think I'll do a merge. Again, I appreciate the high quality of your contribution!

autoload/vimtex/fold/bib.vim
```diff
@@ -0,0 +1,116 @@
...
+" Email:      karl.yngve@gmail.com
+"
+
+function! vimtex#fold#bib#level(lnum) abort " {{{1
+  " Handle blank lines
+  if getline(a:lnum) =~ '\v^\s*$'
+    if a:lnum == 1
+      return 0
+    else
+      let l:prev_foldlevel = vimtex#fold#bib#level(a:lnum - 1)
```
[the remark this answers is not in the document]
yongrenjie: Hmm, I tried this before, and ran into the problem that `foldlevel()` only returns a number. i.e. it throws away the extra information that a foldexpr can provide. In this case I check the return value against `'>1'`, whereas `foldlevel(a:lnum - 1)` would only return `1`.

It turns out that changing this only fails one of the tests, specifically the case where [there are blank lines in the middle of a bib entry](https://github.com/yongrenjie/vimtex/blob/e8cfacd75bd86ff90579ab636732a7f3228b78d1/test/test-folding-bib/test.bib#L76-L87). Which is a bit of an unusual practice to say the least. *But*, I also checked the performance and it seems to make basically no change, which is a shame. I also thought it could shave off a fair bit of the time.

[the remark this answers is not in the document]
yongrenjie: Actually, the `'>1'` case can only happen if the previous line was the beginning of an entry, so it suffices to check that `trim(getline(a:lnum - 1))[0] == '@'`. This gives something like a 2% speed boost, but makes the code a bit more straightforward anyway, so I'll include that in the next commit.

yongrenjie:
> I could make the same tests on my machine, but your bib file seems to be unavailable (perhaps private?).

Whoops! Yes it was private. Fixed now. https://github.com/yongrenjie/hsqc-paper/blob/master/main_final/hsqc.bib

I'll refactor the rest tomorrow probably. Had another look at the foldexpr and managed to squeeze out a 5–10% speedup (so now `hsqc8.bib` is around 540–550 ms, from the previous 592 ms), but I don't think I've any more ideas.

[I'm sure there is some way to improve it, though; the alternative is claiming that I got it right on the first go. :)]

yongrenjie:
@lervag OK, I'm finally done! I couldn't get around recursively calling `vimtex#fold#bib#level()`, unfortunately: when I tried to do so I introduced bugs which my tests didn't catch... But the last commit cuts around 20% of the time calculating foldexprs, which I'm pretty happy with.

Basically the last two commits contain the only 'new' changes since your last round of comments. I also added more tests, which got squashed into the earlier tests commit.

lervag:
Thanks! It looks good!

I've looked at the foldlevel code: You need the recursion because you want to handle the case of single line entries. I'm curious if it would not be a better idea to handle single line entries by _not_ giving them the result `>1` in the first place. I.e., perhaps we could count the braces when we find `@...` and return 0 if the element is self contained on a single line? If this was done, then I believe you could avoid the recursion. Not sure if it is worth the effort, though; I'll still merge this now, but feel free to investigate if you think my idea may have merit.

lervag:
I've merged now. I also made some minor changes on top. Nothing important. To test my idea, I believe the only necessary change is to add to the `if l:firstline == a:lnum` conditional - perform the count here as well (save count results) and return 0 if the braces are balanced, e.g.:

```vim
  " Check if we're at fold start
  let l:text = join(getline(l:firstline, a:lnum))
  let l:braces_balanced = s:count(l:text, '{') == s:count(l:text, '}')
  if l:firstline == a:lnum
    return l:braces_balanced ? 0 : '>1'
  endif

  " Beginning of entry wasn't found
  if l:firstline == 0 | return 0 | endif

  " Check if braces are closed by the current line
  return l:braces_balanced ? '<1' : 1
```

Ref:

https://github.com/lervag/vimtex/blob/1782e479a0c6a716af3a92d5ed85b65ecfad3dc0/autoload/vimtex/fold/bib.vim#L48-L57

yongrenjie:
Thanks! I have a vague recollection that I tried this, and it failed the tests. IIRC the problematic bib entry is

https://github.com/lervag/vimtex/blob/1782e479a0c6a716af3a92d5ed85b65ecfad3dc0/test/test-folding-bib/test.bib#L62-L73

Admittedly, it is probably a pretty rare edge case, so if there is a performance boost it might be worthwhile on balance to change it.

lervag:
I believe I've managed to simplify and improve a little bit. I've pushed an update now; could you test the timings again on your end and see if it has improved?

Could you share the process you use for doing the timings?

lervag:
I've looked at the numbers on my side, and I don't really see any significant speedup on my latest changes. But at least the code is somewhat shorter.

For more detailed profiling, one can use this:

```vim
" test.vim
set nocompatible
let &rtp = '../../..,' . &rtp
filetype plugin on

let g:vimtex_fold_bib_enabled = 1

function! TestProfile() abort
  call vimtex#profile#start()
  normal! zM
  call vimtex#profile#stop()
endfunction

autocmd BufEnter * call TestProfile()
```

The `vimtex#profile#stop` is useful because it interpolates the script local names to make them readable. To use it, run `nvim -u test.vim hsqc16.bib`; then afterwards inspect `prof.log`.

It seems the most expensive lines are these:

```prof
13520              0.113667   let l:firstline = search('\v^\s*\@', 'bcnW')
13520              0.031338   call setpos('.', l:curpos)

                              " Check if we're at fold start
13520              0.072083   let l:text = join(getline(l:firstline, a:lnum))
13520   0.206131   0.085421   let l:count_open = s:count(l:text, '{')
13520   0.198364   0.086337   let l:braces_balanced = l:count_open == s:count(l:text, '}')
```

(From the current version of the code, so not quite what you proposed, but it is generally the same code.)

yongrenjie:
Oh, my 'profiling' is very crude, it's just a `foldlevel()` calculation for the entire file:

https://github.com/yongrenjie/dotfiles/blob/bdc71c43374edc8f171a58e6e0354f811ccf51d2/.vim/ftplugin/bib.vim#L4-L11

By appropriately inserting `return 0` I could also crudely profile how long each block of code took. I think I reached a similar conclusion as above, that the pattern searching is the most time-consuming part, but that's probably not a huge surprise. What I didn't expect was that the whole `getpos()`->`search()`->`setpos()` scheme to find `l:firstline` was actually a lot faster than using a `while` loop to look for it.

The timings are pretty much the same as before, which is a good thing, because I think you made it easier to read. When I wrote this I think I tried to shortcircuit as much as possible, but chances are in 'typical' cases there's not all that much difference.

lervag:
> Oh, my 'profiling' is very crude, it's just a foldlevel() calculation for the entire file:

Thanks! Simple solutions are good, especially when they're sufficient :)

And I agree, the idea of using `search()` was neat! I was not so surprised that it was faster, but I am not sure if I would have realized it myself.

Let's consider this matter resolved for now, then. Again, thanks for your contribution!

paniash:
Hi! I believe the recent commit (related to this PR), now folds bib files by default. Is there any way to avoid this? (I personally prefer no folds in bib files).

clason:
Oh, yeah, folding should remain opt-in ;)

autoload/vimtex/options.vim
```diff
@@ -61,6 +61,8 @@ function! vimtex#options#init() abort " {{{1
     let g:vimtex_fold_enabled = 0
   else
     call s:init_option('vimtex_fold_enabled', 0)
+    call s:init_option('vimtex_fold_bib_enabled', 1)
```
clason, line 64: That should be `0`...

doc/vimtex.txt
```diff
@@ -3998,12 +4017,17 @@ In order to get slightly cleaner fold text, I recommend setting the global
 
 Note: Remember to include the whitespace after backslash!
 
+In addition, VimTeX also provides basic folding in .bib files, which is
+enabled by default. This can be turned on/off via |g:vimtex_fold_bib_enabled|.
```
clason, line 4021: And that should not be the case :)

(I think the assumption was that `fold_bib` is enabled by default _if_ folding is enabled, but that is not how it's implemented -- it is now indiscriminately enabled even if global folding is disabled.)

yongrenjie:
Actually we discussed the default at #2065; but personally I'm not hugely fussed whichever way you guys want to do it.

@paniash `let g:vimtex_fold_bib_enabled = 0` will turn it off.

clason:
I assumed that this was conditional on folding enabled -- if I had realized that this was a new, independent, folding option to be turned on by default, I would have spoken up, _loudly_ ;)

paniash:
> I assumed that this was conditional on folding enabled -- if I had realized that this was a new, independent, folding option to be turned on by default, I would have spoken up, _loudly_ ;)

I agree with @clason. Such options should be opt-in.

yongrenjie:
OK, you can use the option first, and I'll put something together.

clason:
To be clear, I am completely fine with it being enabled by default (or even automatically; I don't care :P) _if_ general folding is enabled. But if folding is disabled, it _must_ be disabled too.

lervag:
Thanks, @paniash; I agree that it was wrong to make the new bib folds enabled by default. It is changed now through the PR by @yongrenjie in #2089.

Thanks for pitching in, @clason. I'm curious, what do you mean with "general folding is enabled"? Do you mean e.g. if `g:vimtex_fold_enabled` is 1? Or do you mean something _more_ general?

clason:
Yes, `g:vimtex_fold_enabled` is exactly what I meant -- I don't think there's a general "folding" option in vim? (Again, I wouldn't know :P)

PR #2067, opened 2021-06-10
Redraw before coming back from quickfix window

andryandrew:
Another try at fixing #1595.
Related discussions are also in a previous PR (#1860) with a different proposed fix which had too many side effects.


lervag:
Thanks! I do not immediately see any harm in this. Can anyone confirm if this also fixes the original issue as intended?

lervag:
I tested the original issue description by @Zeta611 from #1595, and it seems your fix solves the issue. And since I can't see any bad side effects, I'll merge immediately.

PR #2072, opened 2021-06-17
doc: Add nvim-compe to the autocompletion section

savq:
Following [this reddit thread](https://www.reddit.com/r/neovim/comments/o1gcwt/is_it_possible_to_add_the_vimtex_completion_in/), this PR adds some information on how to use VimTeX with nvim-compe.

I didn't find info on whether compe can use regex patterns for filetypes 😕 but I followed u/mybumsonfire advice on adding the filetype explicitly.

I added a code sample in Lua, though another one in Vimscript can also be added.

I'm glad VimTeX is so well documented 👍

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -3695,11 +3695,11 @@ Note: If you want to change the default value of `TEXMFHOME` in your shell
 AUTOCOMPLETE                                             *vimtex-complete-auto*
 
 Vim does not provide automatic completion by itself, but there exist at least
-several good plugins that provide this: |coc-nvim|, |deoplete|, |neocomplete|, |ncm2|,
-|nvim-completion-manager| and |youcompleteme|.  Moreover, there is |VimCompletesMe|
-that overrides <tab> to trigger different built-in completions, such as the
-omni-completion by VimTeX, depending on the context. See below for
-descriptions on how to setup these with VimTeX.
+several good plugins that provide this: |coc-nvim|, |deoplete|, |neocomplete|,
+|ncm2|, |nvim-completion-manager|, |youcompleteme| and |nvim-compe|.
+Moreover, there is |VimCompletesMe| that overrides <tab> to trigger different
+built-in completions, such as the omni-completion by VimTeX, depending on the
+context. See below for descriptions on how to setup these with VimTeX.
 
 coc.nvim~
                                                      *vimtex-complete-coc.nvim*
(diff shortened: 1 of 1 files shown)


jdhao:
I am writing just to confirm that @savq has provided the correct config (which the original reddit post is not).  The value for the key `filetypes` should be a list, not string. My working config for nvim-come (in lua):

```
require'compe'.setup {
  enabled = true;
  autocomplete = true;
  debug = false;
  min_length = 1;
  preselect = 'always';
  throttle_time = 80;
  source_timeout = 200;
  incomplete_delay = 400;
  max_abbr_width = 100;
  max_kind_width = 100;
  max_menu_width = 100;
  documentation = true;

  source = {
    omni = {filetypes = {'tex'}};
    path = true;
    buffer = true;
    spell = {filetypes = {'markdown', 'tex'}};
    emoji = true;
    nvim_lsp = true;
    nvim_lua = true;
    ultisnips = true;
    calc = false;
    vsnip = false;
  };
}
```

lervag:
> Following [this reddit thread](https://www.reddit.com/r/neovim/comments/o1gcwt/is_it_possible_to_add_the_vimtex_completion_in/), this PR adds some information on how to use VimTeX with nvim-compe.

Thank you!

> I'm glad VimTeX is so well documented +1

Yes, I think it's important to aim for good documentation, especially considering the size of the plugin. And help from people like you is therefore very useful!

> I am writing just to confirm that @savq has provided the correct config (which the original reddit post is not).

Thanks! Also thanks for the references to relevant issues!

----

One open question remains: Is there something in nvim-compe that is equivalent to the `input_patterns` variable in deoplete/neocomplete, or the `complete_pattern` in ncm2?

savq:
> One open question remains: Is there something in nvim-compe that is equivalent to the input_patterns variable in deoplete/neocomplete, or the complete_pattern in ncm2?

@lervag, compe seems to be able to use regex patterns for the buffer source, but not for omni-completion. The default pattern matches alphanumeric words, including kebab-case: `\h\w*\%(-\w*\)*`. [The docs](https://github.com/hrsh7th/nvim-compe/blob/d0934e7b335283879fd3e4ca68adaee6a92688b0/doc/compe.txt#L243) mention it's an experimental feature.

lervag:
Ok, thanks. If something like that should be implemented, then feel free to notify me and/or suggest an update to the configuration suggestions in the docs.

PR #2076, opened 2021-06-24
Append success/failure commands of latexmk instead of overwriting.

### LINKED ISSUE #1237: latexmk callback commands (e.g. success_cmd) not respected
I have set `success_cmd` in my global `~/.latexmkrc`:

```perl
$success_cmd = 'my shell command';
```

This works as expected when running latexmk outside of vimtex.  However, vimtex overrides `success_cmd`, and my custom command never runs.

The `*_cmd` variables are discussed in the vimtex docs and there is even an example using xdotool.  There is a note that says:

> If you define these options similar to the above xdotool trick and still want to enable the vimtex callbacks, then one must include a semicolon at the end of the cmd strings so that vimtex may append safely to the options.

I have tried this, i.e. in `~/.latexmkrc`:

```perl
$success_cmd = 'my shell command;';
```

But it still does not work.  Indeed, I can see that in `vimtex/autoload/vimtex/compiler/latexmk.vim` there is no logic for appending to the command strings.

Any thoughts?  Thanks for building vimtex!

Nicholas42:
Append success and failure callbacks to `success_cmd` and `failure_cmd` instead of overwriting them. This allows to use them in a latexmkrc-file, where right now they are overwritten and hence not executed. This fixes #1237. 

I introduce a function `wrap_append_cmd_option` that only differs from `wrap_option` by 

1.  using `.=` instead of `=` which is appending in perl
2. adding a `;` before the appended value. This separates shell commands. It is no problem if the variable was empty beforehand.

I am unsure if this works correctly on windows since I never did scripting there and do not know if `;` has the same effect. One could easily use the old variant for windows if this does not work. 

The change only influences `success_cmd` and `failure_cmd` since I did not change `wrap_option` here. In my opinion overwriting is the right thing to do at the other places it is used.

### DIFF
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -22,6 +22,14 @@ endfunction
 
 "}}}1
 
+function! vimtex#compiler#latexmk#wrap_append_cmd_option(name, value) abort " {{{1
+  return has('win32')
+        \ ? ' -e "$' . a:name . ' .= '';' . a:value . '''"'
+        \ : ' -e ''$' . a:name . ' .= ";' . a:value . '"'''
+endfunction
+
+"}}}1
+
 function! vimtex#compiler#latexmk#get_rc_opt(root, opt, type, default) abort " {{{1
   "
   " Parse option from .latexmkrc.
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! I don't see any reason not to merge this. And I appreciate your contribution!

lervag:
See #2078: I've had to revert this, because it breaks things when the `;` becomes the leading part of a command.

PR #2079, opened 2021-06-26
Fixed appending to success_cmd

Nicholas42:
DO NOT MERGE before testing this on windows. I cannot.

We are very officially in escaping-hell (not to be confused with escaping hell, we do not do that here). Generally, we want to create the following string:

` -e '$success_cmd = ($success_cmd ? $success_cmd . ";" : "")  . "echo something"'`

with `&` instead of `;` on windows. In perl, this checks if `$success_cmd` is non-empty, if so, it adds a `;` after it. To this we append the wanted command. This will break if someone writes `$success_cmd = ' '` in their latexmkrc, but I really do not care.

On linux, we want to use single-quoted strings in the command line, since otherwise, the shell will substitute variables and what not. We would like to also use single quoted strings in the perl expression, since the same applies there, but we cannot escape single quotes in a single-quoted shell string. We could use `'"'"'` to end the single-quoted string, insert a double-quoted single quote and start the single-quoted string again, but this expands to `''"''"''` in vim and that's just unreadable. Hence, we use double quotes in perl and this should be no problem in our use cases.

On windows, there are only double-quoted strings in the cmd.exe, so we use them. Then, we can just use single quotes for perl strings. The command separator is `&`. I am unsure if this has to be specially escaped even if it is quoted, cmd.exe seems to be weird about this. Someone should test this before merging.

### DIFF
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -626,9 +626,11 @@ endfunction
 " Utility functions
 "
 function! s:wrap_option_appendcmd(name, value) abort " {{{1
+  " Do not use with $ in value. On linux, we use double quoted perl strings
+  " that interpolate.
   return has('win32')
-        \ ? ' -e "$' . a:name . ' .= '';' . a:value . '''"'
-        \ : ' -e ''$' . a:name . ' .= ";' . a:value . '"'''
+        \ ? ' -e "$' . a:name . ' = ($' . a:name . ' ? $' . a:name . ' . '' & '' : '''') . ''' . a:value . '''"'
+        \ : ' -e ''$' . a:name . ' = ($' . a:name . ' ? $' . a:name . ' . " ; " : "") . "' . a:value . '"'''
 endfunction
 
 "}}}1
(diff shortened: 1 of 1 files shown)


yongrenjie:
This works fine on my Mac, and I also gave it a quick spin on gvim 8.2 on Windows (with and without `$success_msg = 'echo hello'`) and it seems fine :-)

Nicholas42:
Nice! I tested it the same way on linux.

lervag:
> We are very officially in escaping-hell (not to be confused with escaping hell, we do not do that here).

Yes, these things are really not fun to work with!

> ... This will break if someone writes `$success_cmd = ' '` in their latexmkrc, but I really do not care.

Yes, that seems safe to ignore.

> On linux, we want to use single-quoted strings in the command line, since otherwise, the shell will substitute variables and what not. We would like to also use single quoted strings in the perl expression, since the same applies there, but we cannot escape single quotes in a single-quoted shell string. We could use `'"'"'` to end the single-quoted string, insert a double-quoted single quote and start the single-quoted string again, but this expands to `''"''"''` in vim and that's just unreadable. Hence, we use double quotes in perl and this should be no problem in our use cases.

It seems you understand these things well. Good!

> DO NOT MERGE before testing this on windows. I cannot.

> On windows, there are only double-quoted strings in the cmd.exe, so we use them. Then, we can just use single quotes for perl strings. The command separator is `&`. I am unsure if this has to be specially escaped even if it is quoted, cmd.exe seems to be weird about this. Someone should test this before merging.

I also don't use Windows much, so thanks @yongrenjie for confirming that this works!

Issue #2084, opened 2021-06-29
Latexmk: Need to specify exactly one filename for preview-continuous mode but 6 were specified


Th3Tr00p3r:
Hello,
I have this weird issue that has just popped out seemingly out of nowhere - I'm saying this since vimtex has worked for me in the past few days, compiling without issues using `\ll`. Now for some reason, when I do `\ll` a buffer opens showing the following:

```
latexmk: security risk: running with elevated privileges
Rc files read:
  NONE
Latexmk: This is Latexmk, John Collins, 29 May 2021, version: 4.74b.

Latexmk: Need to specify exactly one filename for preview-continuous mode
    but 6 were specified
Use
   latexmk -help
to get usage information
''' is not recognized as an internal or external command,
operable program or batch file.
''' is not recognized as an internal or external command,
operable program or batch file.
```

`:VimtexInfo` yields:

```
System info
  OS: Microsoft Windows 10 Home (10.0.19043 N/A Build 19043)
  Vim version: NVIM v0.4.4
  Has clientserver: true
  Servername: \\.\pipe\nvim-19252-0

VimTeX project: report
  base: report.tex
  root: path
  tex: path\report.tex
  out: 
  log: 
  aux: 
  fls: 
  main parser: current file verified
  compiler: latexmk
    configuration: 
      continuous: 1
      callback: 1
      latexmk options:
        -verbose
        -file-line-error
        -synctex=1
        -interaction=nonstopmode
      latexmk engine: -pdf
    job: 15
    pid: 0
    cmd: set max_print_line=2000 & latexmk -verbose -file-line-error -synctex=1 -interaction=nonstopmode -pdf -pvc -e "$pdf_previewer = 'SumatraPDF -reuse-instance'" -e "$success_cmd = ($success_cmd ? $success_cmd . ' & ' : '') . 'echo vimtex_compiler_callback_success'" -e "$failure_cmd = ($failure_cmd ? $failure_cmd . ' & ' : '') . 'echo vimtex_compiler_callback_failure'" "report.tex"
  viewer: General
  qf: LaTeX logfile
    addqflist: 345
    fix_paths: 346
    set_errorformat: 344
  document class: article
  packages:
    amsmath
    amssymb
    appendix
    babel
    bm
    caption
    color
    csquotes
    culmus
    epsfig
    epstopdf
    float
    gensymb
    graphicx
    hyperref
    ifpdf
    ifthen
    inputenc
    latexsym
    ragged2e
    sidecap
[...]

PR #2088, opened 2021-07-04
Fix problems with success_cmd on windows nvim.

Nicholas42:
Applied changes as suggested by @yongrenjie in #2084

### DIFF
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -629,9 +629,10 @@ function! s:wrap_option_appendcmd(name, value) abort " {{{1
   " Do not use with $ in value. On linux, we use double quoted perl strings
   " that interpolate.
   let l:win_cmd_sep = has('nvim') ? '^&' : '&'
+  let l:common = a:name . ' = ($' . a:name . ' ? $' . a:name
   return has('win32')
-        \ ? ' -e "$' . a:name . ' = ($' . a:name . ' ? $' . a:name . ' . '' ' . l:win_cmd_sep . ' '' : '''') . ''' . a:value . '''"'
-        \ : ' -e ''$' . a:name . ' = ($' . a:name . ' ? $' . a:name . ' . " ; " : "") . "' . a:value . '"'''
+        \ ? ' -e "$' . l:common . ' . '' ' . l:win_cmd_sep . ' '' : '''') . ''' . a:value . '''"'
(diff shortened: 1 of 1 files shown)


yongrenjie:
I think, just to be safe, I should check what happens if `&shell` is different from the default, since that may affect how the command string is quoted... I'll do that when I can...

Nicholas42:
Yes, that seems like a good idea. One could also think about a parameter en/disabling this since it seems to be very fragile and not *that* important (though, personally I like it very much).

yongrenjie:
Yeah, sadly, adding

```vim
set shell=powershell
set shellcmdflag=-command
```

to `init.vim` gives this error

```
At line:1 char:13
+ "cmd /s /c "set max_print_line=2000 & latexmk -verbose -file-line-err ...
+             ~~~
Unexpected token 'set' in expression or statement.
At line:1 char:37
+ "cmd /s /c "set max_print_line=2000 & latexmk -verbose -file-line-err ...
+                                     ~
The ampersand (&) character is not allowed. The & operator is reserved for future use; wrap an ampersand in double 
quotation marks ("&") to pass it as part of a string.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnexpectedToken
 ```

But this ampersand is already part of the existing code, and wasn't introduced by this PR. So it's probably not our/your problem to fix. :-)

yongrenjie:
And I just noticed `:h vimtex-faq-windows` says to not change `&shell` anyway, so I guess that's not really a problem.

lervag:
Thanks for this, and for the discussion. As I'm not a Windows user, I really do depend on the contribution of others to help make VimTeX work well on Windows as well - it is much appreciated!

lervag:
Please note I also made a simple adjustment; see here:

https://github.com/lervag/vimtex/blob/c1e02461f5ccb15892bf8ce21d41f7fcac33d0db/autoload/vimtex/compiler/latexmk.vim#L628-L638

I believe this should still work and I think it is slightly more readable. I also believe the comment makes more sense now. Can you confirm?

lervag:
Please note I also made a simple adjustment; see here:

https://github.com/lervag/vimtex/blob/c1e02461f5ccb15892bf8ce21d41f7fcac33d0db/autoload/vimtex/compiler/latexmk.vim#L628-L638

I believe this should still work and I think it is slightly more readable. I also believe the comment makes more sense now. Can you confirm?

PR #2089, opened 2021-07-06
hotfix to disable bib folding by default

yongrenjie:
This seems like the easiest way to let the user control tex and bib folding separately (as opposed to making it conditional on `g:vimtex_fold_enabled`, for example).

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -62,7 +62,7 @@ function! vimtex#options#init() abort " {{{1
     let g:vimtex_fold_bib_enabled = 0
   else
     call s:init_option('vimtex_fold_enabled', 0)
-    call s:init_option('vimtex_fold_bib_enabled', 1)
+    call s:init_option('vimtex_fold_bib_enabled', 0)
   endif
   call s:init_option('vimtex_fold_bib_max_key_width', 0)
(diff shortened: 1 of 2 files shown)


clason:
But I _want_ it conditional on `g:vimtex_fold_enabled` (which I have off)!

yongrenjie:
OK, but what happens if someone wants folding in bib files but not tex? Then they would have to `let g:vimtex_fold_enabled=1` and then manually turn off all the tex folding; unless I'm missing something, of course.

clason:
Yes, if that is to be expected, this indeed is the only option that makes sense.

clason:
(I'm not using folds at all, so I have a hard time imagining all the ways that people might wish to ;))

lervag:
I believe the following makes sense:

```vim
call s:init_option('vimtex_fold_enabled', 0)
call s:init_option('vimtex_fold_bib_enabled', g:vimtex_fold_enabled)
```

The main reason is that I expect most people who want folds will want them in both tex and bib. For the few who only wants tex folds _or_ bib folds, it is still easy to adjust the settings accordingly.

I'll merge and make this change immediately.

lervag:
Merged now. Thanks for the PR, @yongrenjie, and for the comments, @clason!

PR #2090, opened 2021-07-07
Do not parse .blg files if bibunits is used.

manu-mannattil:
The [bibunits](https://www.ctan.org/pkg/bibunits) package provides a way to use multiple bibliographies in one document.

A .bib file can be included using bibunits via the commands `\putbib[...]` or `\defaultbibliography{...}`.  Vimtex can be told about these .bib files by changing the `g:vimtex_bibliography_commands` variable.  

However, bibunits generates separate .blg files (sequentially named bu1.blg, bu2.blg, etc.) for each bibliography and because of this, the main .blg file (if it gets generated) does not contain any information about the .bib files that are actually used.  But it is the main .blg file that Vimtex parses and since there is nothing in it, Vimtex cannot find the right .bib files.  This in turn prevents omni completion of the bibliographic entries.

Hence, if bibunits is used, one should resort to manual parsing of the document using the `s:files_manual()` function to find the right .bib files.

### DIFF
--- a/autoload/vimtex/bib.vim
+++ b/autoload/vimtex/bib.vim
@@ -16,7 +16,7 @@ function! vimtex#bib#files() abort " {{{1
   endif
 
   let l:file = b:vimtex.ext('blg')
-  if filereadable(l:file)
+  if filereadable(l:file) && !has_key(b:vimtex.packages, 'bibunits')
     let l:bibs = map(
           \ filter(readfile(l:file), 'v:val =~# ''^Database file #\d'''),
           \ {_, x -> matchstr(x, '#\d\+: \zs.*\ze\.bib$')})


lervag:
Thanks for the detailed explanation and the suggested fix!

I think it would also be useful to add a single test for this functionality. I can do it if you want, but then it would be instructive if you could provide a set of minimal example files (e.g. `mwe.tex`, `mwe1.bib`, `mwe2.bib`, ...).

If you want to attempt to implement a test, then look at for instance [test/test-completion-bibtex/test_custom_bibs.vim](https://github.com/lervag/vimtex/blob/master/test/test-completion-bibtex/test_custom_bibs.vim).

manu-mannattil:
Thanks for the feedback.  On further investigation, this is apparently a problem with the main TeX class (REVTeX) that I was using.  Other classes, e.g, the default article class, or amsart, doesn't have this problem.  (I think it might have something to do with how REVTeX tries to "repair" natbib, but I could be wrong.)  If you think that this is becoming too class/package specific, then feel free to close this pull request.  If you think that this fix is still useful (considering how REVTeX is a popular class), I'll include the following MWE and a test.

<details>
<summary>MWE for REVTeX + bibunits</summary>

```tex
\begin{filecontents*}{library.bib}
  @book{Knuth:1991,
    author = {Donald E. Knuth},
    title = {The \TeX{} Book},
    publisher = {Addison--Wesley, Reading, Massachusetts},
    year = 1991,
  }

  @book{Lamport:1994,
    author = {Leslie Lamport},
    title = {\LaTeX: A Document Preparation System},
    publisher = {Addison--Wesley, Reading, Massachusetts},
    edition = 2,
    year = 1991,
  }

  @book{Adobe:1985,
    author = {Adobe Systems Inc.},
    title = {Postscript Language Tutorial and Cookbook},
    publisher = {Addison--Wesley, Reading, Massachusetts},
    year = 1985,
  }
\end{filecontents*}

% This is a problem with both versions 4.1 and 4.2
\documentclass{revtex4-1}
% \documentclass{revtex4-2}

\usepackage{bibunits}
\defaultbibliographystyle{apsrev4-1}
\defaultbibliography{library}

\begin{document}

\begin{bibunit}
  References to the \TeX book \cite{Knuth:1991} and to Lamport’s \LaTeX\ book, which appears only in the references\nocite{Lamport:1994}.
  \putbib
\end{bibunit}

\begin{bibunit}
  A new bibunit with a different style and a cite of a Postscript tutorial \cite{Adobe:1985}.
  \putbib
\end{bibunit}

\end{document}
```
</details>

Instead of a package-specific fix, perhaps it's better to have an option that disables parsing of the .blg files, though I'm not sure if other users would find such an option useful.

lervag:
> Thanks for the feedback. On further investigation, this is apparently a problem with the main TeX class (REVTeX) that I was using. Other classes, e.g, the default article class, or amsart, doesn't have this problem. (I think it might have something to do with how REVTeX tries to "repair" natbib, but I could be wrong.) ...

Ah, ok. It seems to me that `revtex4` is to blame. I'm pushing a fix for this now. Can you test if it is sufficient?

manu-mannattil:
Yes, this does work.  Thanks!

lervag:
Great, happy to hear it!

PR #2105, opened 2021-07-13
Motion mappings doc update for g:vimtex_motion_enabled

krishnakumarg1984:
It looks like the documentation wording for `g:vimtex_motion_enabled` has not been updated to account for the motions added to vimtex since the original set.  This PR has a minor documentation update which includes the full range of motions supported currently. 

The updated list is slightly re-ordered to be consistent with the order presented in README.md and in the feature-overview section at the top of the help file.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -161,7 +161,7 @@ FEATURE OVERVIEW                                              *vimtex-features*
     - give enhanced |gf| command
 - Easy access to (online) documentation of packages
 - Word count (through `texcount`)
-- Motions
+- Motions                                                      *vimtex-motions*
   - Move between section boundaries with `[[`, `[]`, `][`, and `]]`
   - Move between environment boundaries with `[m`, `[M`, `]m`, and `]M`
   - Move between math environment boundaries with `[n`, `[N`, `]n`, and `]N`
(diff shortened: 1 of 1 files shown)


lervag:
> It looks like the documentation wording for `g:vimtex_motion_enabled` has not been updated to account for the motions added to vimtex since the original set. This PR has a minor documentation update which includes the full range of motions supported currently.

When I look at this now, I realize it may be silly to list the motions here. We already see how I forget to update it. Perhaps it is better to keep a single list and refer to it?

lervag:
For instance, I think it could be better to add a tag here:

https://github.com/lervag/vimtex/blob/c6929fda3253b3e7a476fb2f40571a01ffe963d7/doc/vimtex.txt#L164-L170

```
- Motions                                                   *vimtex-motions*
  - Move between section boundaries with `[[`, `[]`, `][`, and `]]`
  - Move between environment boundaries with `[m`, `[M`, `]m`, and `]M`
  - Move between math environment boundaries with `[n`, `[N`, `]n`, and `]N`
  - Move between frame environment boundaries with `[r`, `[R`, `]r`, and `]R`
  - Move between comment boundaries with `[*` and `]*`
  - Move between matching delimiters with `%`

...
...

*g:vimtex_motion_enabled*
  This option enables the motion mappings, see |vimtex-motions|. It also
  enables the highlighting of matching delimiters.

  Default value: 1
```

What do you think?

krishnakumarg1984:
Great. That's perfect I think. Subscribes to the DRY (Don't repeat yourself) principle. Would you like me to update the PR with the proposed change (I'll be happy to it tomorrow)?

lervag:
Yes, that would be great!

ons. 14. jul. 2021, 01:16 skrev Krishnakumar Gopalakrishnan, PhD <
***@***.***>:

krishnakumarg1984:
Done!

PR #2118, opened 2021-07-27
Allow optional group and more than one mandatory argument in nospell commands

tornaria:
When a command is in `g:vimtex_syntax_nospell_commands`, its argument is excluded from spell checking; after the first patch this also works for commands with an optional argument.

After the second patch, a command in `g:vimtex_syntax_nospell_commands` can have more than one mandatory argument; spell check is disabled for all of them.

I use this for custom commands with optional arguments; also for some standard commands like `\numberwithin` (2 mandatory arguments).

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -287,7 +287,7 @@ function! vimtex#syntax#core#init() abort " {{{1
     execute 'syntax match texCmdNoSpell nextgroup=texNoSpellOpt,texNoSpellArg skipwhite skipnl "\\' . l:macro . '"'
   endfor
   call vimtex#syntax#core#new_opt('texNoSpellOpt', {'next': 'texNoSpellArg'})
-  call vimtex#syntax#core#new_arg('texNoSpellArg', {'contains': 'TOP,@Spell'})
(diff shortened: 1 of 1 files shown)


autoload/vimtex/syntax/core.vim
```diff
@@ -758,6 +759,8 @@ function! vimtex#syntax#core#init_highlights() abort " {{{1
   highlight def link texNewthmArgName      texArg
   highlight def link texNewthmOptCounter   texOpt
   highlight def link texNewthmOptNumberby  texOpt
+  highlight def link texNoSpellOpt         texOpt
+  highlight def link texNoSpellArg         texArg
```
lervag, line 763: I'm not sure if we really want to add the default highlight for `texNoSpellArg` here. Users can add that if they want, but I thought it OK to just keep it transparent.

lervag (changes requested):
Great, thanks. I don't see any reason not to accept this, except my minor comment.

autoload/vimtex/syntax/core.vim
tornaria: You are right, done.

lervag:
Thanks!

PR #2119, opened 2021-07-27
Add a few commands for amstex

tornaria:
- `\DeclareMathOperator`: since the body is in math mode, do not spell check
- new commands: `\theoremstyle`, `\numberwithin` (2 args), `\subjclas` (1 opt + 1 arg): do not spell-check.


lervag:
Thanks! You even order the highlight groups consistently! :D

One minor thing, but important IMHO: Can you add some examples of the new commands to the test files, `test-amsthm.tex` and `test-amsmath.tex`? You might also want to add tests, but the most important thing is to make it easy to manually inspect with e.g. `vim -u test-amsthm.vim`.

PR #2121, opened 2021-07-28
use vimtex#util#trim() instead of trim() for Vim versions < 8.1

Konfekt:
The two occurrences of trim() in

  https://github.com/lervag/vimtex/blob/1782e479a0c6a716af3a92d5ed85b65ecfad3dc0/autoload/vimtex/parser/bib.vim#L34

and

  https://github.com/lervag/vimtex/blob/3807da1c530e46fb6a633a2268f6e435b16b657b/autoload/vimtex/fold/bib.vim#L24

are replaced by

  https://github.com/lervag/vimtex/blob/6228f72a8bcf25b31bf1a966de864349ed4ea61c/autoload/vimtex/util.vim#L192

for Vim versions < 8.1.

This creates dependencies outside of the bib file, but they have been there before anyway.

### DIFF
--- a/autoload/vimtex/fold/bib.vim
+++ b/autoload/vimtex/fold/bib.vim
@@ -21,7 +21,7 @@ endfunction
 
 function! vimtex#fold#bib#level(lnum) abort " {{{1
   " Handle blank lines
-  if empty(trim(getline(a:lnum)))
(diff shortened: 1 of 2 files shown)


lervag:
Thanks!

PR #2122, opened 2021-07-30
Allow braces in column specification for array env

psvenk:
The `array` environment supports (most of) the same column specifications as the `tabular` environment does; update the syntax highlighting to reflect this fact and avoid a `texGroupError` when using braces in the column specification.

Screenshots from a [file](https://gist.github.com/psvenk/0fd74b6e271df9916e21f763f4fc1c8c) that I threw together (with `:colo default`):

| Before | After |
| --- | --- |
| screenshot-2021-07-29-20-19-36 | screenshot-2021-07-29-20-24-18 |

I also updated `test/test-syntax/test-array.tex` and `test/test-syntax/test-array.vim` with a test case for this, along with an analogous test case for `tabular` to the `.vim` file because that was not already tested.

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -570,7 +570,9 @@ function! vimtex#syntax#core#init() abort " {{{1
   " Support for array environment
   syntax match texMathCmdEnv contained contains=texCmdMathEnv "\\begin{array}" nextgroup=texMathArrayArg skipwhite skipnl
   syntax match texMathCmdEnv contained contains=texCmdMathEnv "\\end{array}"
-  call vimtex#syntax#core#new_arg('texMathArrayArg', {'contains': ''})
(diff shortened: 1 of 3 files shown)


lervag:
Are you sure about this? I've never found any documentation that supports this, and your example does not compile on my end...

image

krypt-n:
The `m{10}` seems to require the `array` package. Replace it with e.g. l to make the example compile without it.

```
\documentclass{article}

\begin{document}
\[
    \begin{array}{*{4}{l}}
        5
    \end{array}
\]
\end{document}
```

psvenk:
Sorry for the confusion. As @krypt-n observed, the examples in the test file do indeed require the `array` package. I had tested primarily with [this file](https://gist.github.com/psvenk/0fd74b6e271df9916e21f763f4fc1c8c), which does compile on my end without any packages.

image

lervag:
Ok, thanks! I've merged and applied some minor adjustments to make the test file compileable.

PR #2125, opened 2021-08-04
prepend backslashes to `=word`s in 'indentkeys'

BertrandSim:
Currently, inserting tex keywords `item`, `else`, `fi`, and tex delimiters autoindents the current line. 
However, there are some words that are part of these keywords. (For example, 'fix', 'figure', are part of 'fi'.) 
In these cases, it would be better _not_ to autoindent the current line.

Adding a backslash to these keywords and tex delimiters will reduce the number of unexpected autoindents (autoindent only on `\item`, `\else`, `\fi`, etc.)

### DIFF
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -20,12 +20,12 @@ set cpoptions&vim
 
 setlocal autoindent
 setlocal indentexpr=VimtexIndentExpr()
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! That's simple and useful!

PR #2128, opened 2021-08-09
Vlty: guess language

Konfekt:
Vlty often fails because of nonexistent  language variant codes. Try a best guess before warning the user of such a code.


lervag:
Thanks! I made some adjustments, but this looks good!

PR #2135, opened 2021-08-17
Changes to SyncTeX backward search

### LINKED ISSUE #2127: Zathura backward search: bug with multi-file projects
#### Problem
The SyncTeX support for Zathura is excellent out of the box, but I noticed that the backward search command line is hard-coded to use `--remote-expr` to call `vimtex#view#reverse_goto`:
https://github.com/lervag/vimtex/blob/078292ed7efb95a5ff6c4cf21f4273ae599af2bd/autoload/vimtex/view/zathura.vim#L38-L41

This function works well in single-file projects, but I noticed a few problems when switching files in multi-file projects. Specifically, consider the case that a project has files `a.tex` and `b.tex`, my Vim currently has `a.tex` open, and I perform a reverse search from a place in the PDF that corresponds to `b.tex` line 50.
- If `a.tex` has unsaved changes, then VimTeX focuses `a.tex` line 50. This seems to be because the `silent edit` fails here:
https://github.com/lervag/vimtex/blob/078292ed7efb95a5ff6c4cf21f4273ae599af2bd/autoload/vimtex/view.vim#L74-L77
- If `a.tex` does not have unsaved changes, then `b.tex` takes over the buffer where I was editing `a.tex`. While this may be desirable behavior for other users, I would prefer that `b.tex` open in a new tab page.

#### Suggested solutions/workarounds
I suppose that the first bullet point is a bug and should be fixed globally, but that the second bullet point hinges on subjective preferences. So, I propose that the first part be fixed e.g. by splitting the window, which is what Vim's `--remote` argument does. To address the tab page issue, I think it would be a good idea to make the behavior configurable.

Alternatively, the use of `--remote-expr` (as opposed to `--remote` or `--remote-tab`) could be made configurable. The stopgap solution that I am using right now is changing the hard-coded `--remote-expr` to use `--remote-tab`, and this seems to work fine.

<details>
<summary>My stopgap solution</summary>

```diff
diff --git a/autoload/vimtex/view/zathura.vim b/autoload/vimtex/view/zathura.vim
index 95235783..ec17b141 100644
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -37,8 +37,7 @@ function! s:zathura.start(outfile) dict abort " {{{1
   if self.has_synctex
     let l:cmd .= ' -x "' . g:vimtex_compiler_progname
           \ . ' --servername ' . v:servername
-          \ . ' --remote-expr '
-          \ .     '\"vimtex#view#reverse_goto(%{line}, ''%{input}'')\""'
+          \ . ' --remote-tab +\%{line} \%{input}\"'
     if g:vimtex_view_forward_search_on_start
       let l:cmd .= ' --synctex-forward '
             \ .  line('.')
@@ -80,7 +79,7 @@ function! s:zathura.latexmk_append_argument() dict abort " {{{1
     if self.has_synctex
       let zathura .= ' -x \"' . g:vimtex_compiler_progname
           \ . ' --servername ' . v:servername
-          \ . ' --remote +\%{line} \%{input}\" \%S'
+          \ . ' --remote-tab +\%{line} \%{input}\" \%S'
     endif
 
     let cmd  = vimtex#compiler#latexmk#wrap_option('new_viewer_always', '0')
```

</details>

I suppose that one could also work around this by using the `general` viewer and configuring everything manually, but this has obvious downsides.

#### Miscellany
Additionally, I noticed that `--remote` and not `--remote-expr` is used later in the same file, in `s:zathura.latexmk_append_argument()`. Is there a reason for this? Apologies if this (or anything else in this issue) has already been addressed elsewhere.
https://github.com/lervag/vimtex/blob/078292ed7efb95a5ff6c4cf21f4273ae599af2bd/autoload/vimtex/view/zathura.vim#L79-L84

psvenk:
Fixes #2127. This pull requests addresses three closely related issues related to SyncTeX backward search in multi-file projects, the last of which is the one originally brought up in #2127:
- The `latexmk` command line, invoked upon `<LocalLeader>ll`, opens Zathura with `--remote` for backward search, while running `<LocalLeader>lv` opens Zathura with `--remote-expr` for backward search. This can lead to inconsistent behavior; 2139057fe4 fixes this by properly escaping all arguments in the `latexmk` command line in order to use `--remote-expr`.
- Backward search does not properly take into account the possibility that a file is open in another buffer, but not in the current tab page. This was fixed in 9af661097f.
- The function `vimtex#view#reverse_goto`, which is called upon backward search, is hard-coded to use `edit` to open a new file, usurping the current window; b003fee92d makes this configurable and uses `bufloaded` instead of `bufexists` to avoid falling back to usurpation when a buffer exists but is not visible (e.g., if it were opened in another tab page and then closed, we want to reopen it in a tab page — if this is the user's preference — and not always try to usurp the current window to load that buffer).

### DIFF
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -72,10 +72,10 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
   let l:file = resolve(a:filename)
 
   " Open file if necessary
-  if !bufexists(l:file)
+  if !bufloaded(l:file)
     if filereadable(l:file)
       try
-        execute 'edit' l:file
+        execute g:vimtex_view_reverse_search_edit_cmd l:file
       catch
         call vimtex#log#warning("Reverse goto failed")
         return
@@ -97,7 +97,7 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
     execute l:tabnr . 'tabnext'
     execute l:winnr . 'wincmd w'
   catch
-    execute 'edit' l:file
+    execute g:vimtex_view_reverse_search_edit_cmd l:file
   endtry
 
   execute 'normal!' a:line . 'G'
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -87,11 +87,18 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
   endif
 
   " Go to correct buffer and line
+
+  " Get buffer number
   let l:bufnr = bufnr(l:file)
-  let l:winnr = bufwinnr(l:file)
-  execute l:winnr >= 0
-        \ ? l:winnr . 'wincmd w'
-        \ : 'buffer ' . l:bufnr
+  " Get window and tab numbers
+  try
+    let [l:winid] = win_findbuf(l:bufnr)
+    let [l:tabnr, l:winnr] = win_id2tabwin(l:winid)
+    execute l:tabnr . 'tabnext'
+    execute l:winnr . 'wincmd w'
+  catch
+    execute 'edit' l:file
+  endtry
 
   execute 'normal!' a:line . 'G'
   redraw
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -80,7 +80,7 @@ function! s:zathura.latexmk_append_argument() dict abort " {{{1
     if self.has_synctex
       let zathura .= ' -x \"' . g:vimtex_compiler_progname
           \ . ' --servername ' . v:servername
-          \ . ' --remote +\%{line} \%{input}\" \%S'
+          \ . ' --remote-expr \"\\\"\"vimtex#view#reverse_goto(\%{line}, ''"''"''\%{input}''"''"'')\"\\\"\"\" \%S'
     endif
 
     let cmd  = vimtex#compiler#latexmk#wrap_option('new_viewer_always', '0')
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -378,6 +378,7 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_view_method', 'general')
   call s:init_option('vimtex_view_use_temp_files', 0)
   call s:init_option('vimtex_view_forward_search_on_start', 1)
+  call s:init_option('vimtex_view_reverse_search_edit_cmd', 'edit')
 
   " OS dependent defaults
   let l:os = vimtex#util#get_os()
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -2743,6 +2743,18 @@ OPTIONS                                                        *vimtex-options*
 
   Default value: 1
 
+*g:vimtex_view_reverse_search_edit_cmd*
+  When working in a multi-file project, initiating backward search
+  |vimtex-synctex-backward-search| may require opening a file that is not
+  currently open in a window. This option controls the command that is used to
+  open files as a result of a backward search.
+
+  Examples: `edit` (open buffer in current window); `tabedit` (open buffer in
+  new tab page, similar to the Vim command-line option `--remote-tab`);
+  `split` (split current window to open buffer, similar to `--remote`)
+
+  Default value: `edit`
+
 *g:vimtex_view_method*
   Set the viewer method.
 


lervag:
> Fixes #2127. This pull requests addresses three closely related issues related to SyncTeX backward search in multi-file projects, the last of which is the one originally brought up in #2127:

Thanks!

> The `latexmk` command line, invoked upon `<LocalLeader>ll`, opens Zathura with `--remote` for backward search, while running `<LocalLeader>lv` opens Zathura with `--remote-expr` for backward search. This can lead to inconsistent behavior; [2139057](https://github.com/lervag/vimtex/commit/2139057fe4a9c9a5495472c8e321f4ed0e66810d) fixes this by properly escaping all arguments in the `latexmk` command line in order to use `--remote-expr`.

Wow, impressive level of shell escape magic!

> Backward search does not properly take into account the possibility that a file is open in another buffer, but not in the current tab page. This was fixed in [9af6610](https://github.com/lervag/vimtex/commit/9af661097f15af588cd572f25397cd2f4c451dc8).

Great, thanks!

> The function `vimtex#view#reverse_goto`, which is called upon backward search, is hard-coded to use `edit` to open a new file, usurping the current window; [b003fee](https://github.com/lervag/vimtex/commit/b003fee92d5b4ea5179f5801a3f554506dccdb95) makes this configurable and uses `bufloaded` instead of `bufexists` to avoid falling back to usurpation when a buffer exists but is not visible (e.g., if it were opened in another tab page and then closed, we want to reopen it in a tab page — if this is the user's preference — and not always try to usurp the current window to load that buffer).

Sounds good!

autoload/vimtex/view.vim
```diff
@@ -87,11 +87,18 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
...
-        \ ? l:winnr . 'wincmd w'
-        \ : 'buffer ' . l:bufnr
+  " Get window and tab numbers
+  try
+    let [l:winid] = win_findbuf(l:bufnr)
+    let [l:tabnr, l:winnr] = win_id2tabwin(l:winid)
+    execute l:tabnr . 'tabnext'
+    execute l:winnr . 'wincmd w'
+  catch
+    execute g:vimtex_view_reverse_search_edit_cmd l:file
```
lervag, line 100: I don't get why we need the second `execute ...` (which is the same as line 78)? Is it because of the `if !bufloaded`?

autoload/vimtex/view.vim
```diff
@@ -87,11 +87,18 @@ function! vimtex#view#reverse_goto(line, filename) abort " {{{1
...
   " Go to correct buffer and line
+
+  " Get buffer number
   let l:bufnr = bufnr(l:file)
-  let l:winnr = bufwinnr(l:file)
-  execute l:winnr >= 0
-        \ ? l:winnr . 'wincmd w'
-        \ : 'buffer ' . l:bufnr
+  " Get window and tab numbers
+  try
```
lervag, line 94: When would the following code fail? During `tabnext`? Or `wincmd`?

autoload/vimtex/view/zathura.vim
```diff
@@ -80,7 +80,7 @@ function! s:zathura.latexmk_append_argument() dict abort " {{{1
     if self.has_synctex
       let zathura .= ' -x \"' . g:vimtex_compiler_progname
           \ . ' --servername ' . v:servername
-          \ . ' --remote +\%{line} \%{input}\" \%S'
+          \ . ' --remote-expr \"\\\"\"vimtex#view#reverse_goto(\%{line}, ''"''"''\%{input}''"''"'')\"\\\"\"\" \%S'
```
lervag, line 83: Hahahaha! This makes my day! :D

lervag:
I'll merge after feedback on my questions.

autoload/vimtex/view/zathura.vim
lervag: And the worst part of this is that it works! Great work!

autoload/vimtex/view.vim
psvenk: This would actually fail in line 95; if the buffer does not exist, then `win_findbuf` returns an empty list and the unpacking fails.

autoload/vimtex/view.vim
psvenk: Yes, I think so. I included this to replicate the behavior of the old line 94, which used `buffer` to load the buffer if it exists but is not loaded. With the change from `bufexists` to `bufloaded`, this would load the buffer if it is loaded but for whatever reason `win_findbuf` does not find it. From my limited knowledge, it seems that this should not happen, but I think I found a counterexample in my testing (though I don't remember what the situation was).

I changed `buffer` to the value of `g:vimtex_view_reverse_search_edit_cmd` to avoid usurping the current buffer in this case (if that is the user's preference). With default settings, `edit` is used instead of buffer and (as far as I know) the user experience is largely unchanged.

autoload/vimtex/view/zathura.vim
psvenk: Thank you! I had to do quite a bit of testing with latexmk on the command line. For the record, this is what I used:
```sh
latexmk -pvc -new-viewer -pdf lectures.tex -e '$pdf_previewer = "zathura -x \"/usr/bin/vim.gtk3 --servername VIM --remote-expr \"\\\"\"vimtex#view#reverse_goto(\%{line}, '"'"'\%{input}'"'"')\"\\\"\"\" lectures.pdf"'
```
The output contained the command line used by latexmk to start Zathura, which I could cross-check against the use of `--remote-expr` already in `zathura.vim`:
```
Running 'start zathura -x "/usr/bin/vim.gtk3 --servername VIM --remote-expr "\""vimtex#view#reverse_goto(%{line}, '%{input}')"\""" lectures.pdf'
```

lervag:
Thanks for the good work! I made some minor adjustments and merged this just now.

PR #2151, opened 2021-08-27
Patch 1

AlexFreik:
--remote option hasn’t been restored by Neovim, so current solution with backward search in Skim aren't working anymore.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -5331,7 +5331,7 @@ A: Set |g:vimtex_view_method| to `skim` (and read |vimtex_viewer_skim| if you
 
        Select the `Custom` preset from the drop-down list and set:
          Command: nvr~
-         Args: --servername `cat /tmp/curnvimserver.txt` +"%line" "%file"
+         Args: --servername `cat /tmp/curnvimserver.txt` +"%line" "%file"~
           
        Please ensure that you have the `nvr` (neovim-remote) utility and the
        `pynvim` Python module available, see |vimtex-faq-neovim|.
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -5307,9 +5307,32 @@ A: Set |g:vimtex_view_method| to `skim` (and read |vimtex_viewer_skim| if you
    options according to your desired version of Vim:
 
      * For `MacVim`, select the `MacVim` preset.
-     * For `neovim`, select the `Custom` preset from the drop-down list and set:
+     * For `neovim`.
+       You would need to install neovim-remote.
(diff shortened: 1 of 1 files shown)


AlexFreik:
> I don't think that's correct; I'm using neovim with Skim and the current instructions are working fine for me. I most certainly don't need a complicated autocommand!
> 
> (Of course you need to have `neovim-remote` installed, but that is explained in the very next paragraph (which you can still see in the diff view.)

Indeed it works) I hadn't noticed instructions to install neovim-remote and it wasn't working... Sorry for the misleading)

AlexFreik:
> I don't think that's correct; I'm using neovim with Skim and the current instructions are working fine for me. I most certainly don't need a complicated autocommand!
> 
> (Of course you need to have `neovim-remote` installed, but that is explained in the very next paragraph (which you can still see in the diff view.)

Sorry again) I rechecked and backward search is not working with the Args: --remote +"%line" "%file", but it works with Args: --servername `cat /tmp/curnvimserver.txt` +"%line" "%file"

I have neovim-remote and pynvim (when I run pip3 install neovim-remote the result is Requirement already satisfied: neovim-remote in /usr/local/lib/python3.9/site-packages (2.4.0)... and the same for pynvim)

AlexFreik:
> @AlexFreik Can you explain what the problem with the current instructions are? My impression has been that they should already work well. Perhaps it has to do with conflicting servernames? I believe `nvr` works well if you only use a single neovim instance, but if you use multiple at the same time, then you might need to specify the servername - and this may be the problem the updated guide is solving?

I have just one opened file $ nvim main.tex, so it is not the root of problem I think. For some reason current solution just silently not working. Should I open the issue instead?

clason:
I think we should first figure out why it is not working for you. I have been using exactly these argument (I just double-checked) for years and never had any issue.

AlexFreik:
> I think we should first figure out why it is not working for you. I have been using exactly these argument (I just double-checked) for years and never had any issue.

Do you have any ideas why it is not working? My neovim version: NVIM v0.6.0-dev+170-gdb1b0ee3b. I am on the MacBook Air 2019 year model with Big Sur OS version 11.4

clason:
None whatsoever. You haven't shared your VimTeX (and `latexmk`) configuration, though.

lervag:
> Should I open the issue instead?

Yes, that would be my suggestion. Notice the structure of a good issue report; the template should help.

AlexFreik:
> > Should I open the issue instead?
> 
> Yes, that would be my suggestion. Notice the structure of a good issue report; the template should help.

Submitted new Issue #2157

Issue #2190, opened 2021-10-02
Wrong language passed to grammar checkers 'textidot' and 'vlty'


cristobaltapia:
### Description

When trying to use the grammar checkers 'textidot' or 'vlty' an error is displayed, saying that the language is unknown. 

### Steps to reproduce

I have the following configuration defined:

```vim
let g:tex_flavor = 'latex'

let g:vimtex_grammar_vlty = {}
let g:vimtex_grammar_vlty.lt_command = 'languagetool'
let g:vimtex_grammar_vlty.server = 'my'
let g:vimtex_grammar_vlty.show_suggestions = 1
let g:vimtex_grammar_vlty.shell_options =
        \   ' --multi-language'
        \ . ' --packages "*"'
        \ . ' --define ~/vlty/defs.tex'
        \ . ' --replace ~/vlty/repls.txt'
        \ . ' --equation-punctuation display'
        \ . ' --single-letters "i.\,A.\|z.\,B.\|\|"'
        \ . " --lt-options '~--language en-US'"

let g:vimtex_grammar_textidote = {
            \ 'jar': '/opt/textidote/textidote.jar',
            \ 'args': '',
            \}

```

I define the spell language within the file with a modline:
```tex
\documentclass{minimal}
\begin{document}
Hello world!
\end{document}
% vi:spelllang=en_us
```

Then run e.g.:
```
compiler textidote | make
```

### Expected behavior

Grammar check

### Actual behavior

Errors occur.
Specifically,, textidote throws this error:

```
:!java -jar '/opt/textidote/textidote.jar' --no-color --output sing
leline --check enUS 'minimal.tex'  2>&1| tee /tmp/nvimIfjg0e/11         
TeXtidote v0.8.2 - A linter for LaTeX documents and others
(C) 2018-2021 Sylvain Hallé - All rights reserved

Unknown language: enUS
```

and vlty this:

```
Language 'en' not listed in output of languagetool --list!
Trying 'en' instead.                                                   
Error detected while processing /home/tapia/.vim/plugged/vimtex/compiler/vlty.vim
:
line   90:
:!python3 -m yalafi.shell --lt-command languagetool --server my --encoding utf-8
--language en --disable "WHITESPACE_RULE" --enable "" --disablecategories "" --en
ablecategories "" --documentclass "minimal" --packages ""  --multi-language --pac
kages "*" --define ~/vlty/defs.tex --replace ~/vlty/repls.txt --equation-punctuat
[...]

PR #2191, opened 2021-10-03
fix check whether valid language was passed to VLTY

Konfekt:
addresses https://github.com/lervag/vimtex/issues/2190


lervag:
Thanks!

Issue #1885, opened 2020-12-07
Reduce delay of completion 


tlannaax:
**Is your feature request related to a problem? Please describe it.**
When a completion invokes kpsewhich there is usually a noticeable delay for the first time.
Setting g:vimtex_include_search_enabled = 0 is only a compromise:
-  you lose access to, for example, manually installed packages and master bib database under your local texmf. 
-  when used together with autocompletion plugins, (I tried coc-vimtex,) kpsewhich is still triggered and causes delay.

As a workaround, I turned off vimtex completion completely and used texlab LSP to provide the completion feature.

**Describe the solution you'd like**
Maybe exploiting the async feature of vim8 and neovim to run kpsewhich asynchronously could reduce the delay.

lervag:
Can you be more specific about which type of completion where you experience the delay? Cite completion? If I am to work on this, it would be quite useful to have a good example where the delay is present. I.e., 

For me, the delay is really quite small (but I agree it is noticable). Perhaps 0.5 seconds. And just the first time. Note that this is for cite completion with a relatively large bib file.

I'm not quite sure how to _use_ the async features, to be honest. I would be happy to do this, but it would be easier if someone who are already familiar with the features could guide me.

tlannaax:
I am on windows, and vim calls cmd.exe to run kpsewhich. You know terminal on windows is considerably slower than that on linux or mac. For me, cite completion takes about 2 seconds. The worse thing is when I edit an old project and try to insert an environment. There are usually many packages included in the preamble. After \begin{ triggering autocompletion, it usually takes more than 5 seconds for kpsewhich to finish, meanwhile a bunch of cmd windows appear and closes.

lervag:
> I am on windows, and vim calls cmd.exe to run kpsewhich. You know terminal on windows is considerably slower than that on linux or mac.

I really didn't know, and I have to admit I don't really know how to improve that. Except, perhaps, adding the async idea you propose.

> For me, cite completion takes about 2 seconds. The worse thing is when I edit an old project and try to insert an environment. There are usually many packages included in the preamble. After \begin{ triggering autocompletion, it usually takes more than 5 seconds for kpsewhich to finish, meanwhile a bunch of cmd windows appear and closes.

I assume this only occurs the first time, right? There is caching now, so subsequent completion should be fast.

[...]

PR #2195, opened 2021-10-06
Create modern jobs api for simple processes

lervag:
NB! This PR is not finished yet and is not ready for testing.

I want to remove the old `process.vim` module and switch it with a new `jobs.vim` module that relies on the `jobstart` (nvim) and `job_start` (Vim) functions. I believe this may be useful in many cases, and perhaps give a better experience on Windows especially (cf. e.g. #1885).

### DIFF
--- a/autoload/vimtex/qf/pplatex.vim
+++ b/autoload/vimtex/qf/pplatex.vim
@@ -85,9 +85,10 @@ function! s:qf.addqflist(tex, log) abort dict " {{{1
   endif
 
   let l:tmp = fnamemodify(a:log, ':r') . '.pplatex'
-  silent call system(printf('pplatex -i "%s" >"%s"', a:log, l:tmp))
+
+  call system(printf('pplatex -i "%s" >"%s"', a:log, l:tmp))
   call vimtex#qf#u#caddfile(self, l:tmp)
-  silent call delete(l:tmp)
+  call delete(l:tmp)
 endfunction
 
 " }}}1
--- a/autoload/vimtex/qf/pulp.vim
+++ b/autoload/vimtex/qf/pulp.vim
@@ -49,12 +49,12 @@ function! s:qf.addqflist(tex, log) abort dict " {{{1
     throw 'VimTeX: No log file found'
   endif
 
-  let l:tmp = fnameescape(fnamemodify(a:log, ':r') . '.pulp')
+  let l:tmp = fnamemodify(a:log, ':r') . '.pulp'
   let l:log = fnameescape(a:log)
 
-  silent call system(printf('pulp %s >%s', l:log, l:tmp))
+  call system(printf('pulp %s >%s', l:log, fnameescape(l:tmp)))
   call vimtex#qf#u#caddfile(self, l:tmp)
-  silent call system('rm ' . l:tmp)
+  call delete(l:tmp)
 endfunction
 
 " }}}1
--- a/autoload/vimtex/view.vim
+++ b/autoload/vimtex/view.vim
@@ -133,12 +133,12 @@ function! s:focus_vim() abort " {{{1
   let l:pids = l:pids[: index(l:pids, string(l:current_pid))]
 
   for l:pid in reverse(l:pids)
-    let l:xwinids = filter(reverse(split(system(
-        \ 'xdotool search --onlyvisible --pid ' . l:pid))),
-        \ '!empty(v:val)')
+    let l:xwinids = filter(reverse(
+          \ systemlist('xdotool search --onlyvisible --pid ' . l:pid)),
+          \ '!empty(v:val)')
 
     if !empty(l:xwinids)
-      silent call system('xdotool windowactivate ' . l:xwinids[0] . ' &')
+      call system('xdotool windowactivate ' . l:xwinids[0] . ' &')
       call feedkeys("\<c-l>", 'tn')
       return l:xwinids[0]
       break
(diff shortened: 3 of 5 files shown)


habamax:
Sure, will check it, @lervag

habamax:
regarding #1885, I am experiencing the same delay for initial `<C-x><C-o>` (~30~40 seconds for my doc). Not sure if you were to fix it with this PR too.

Other than that it works the same for me. Will continue using it and let you know if there are issues.

habamax:
Another note, after I `VimtexClearCache All`, it takes < 2 secs to show the completion. 

Tried with another new document it is < 2 seconds. :)

Not sure why first time was so long.

habamax:
> That's good to hear, though! Are you on Vim or neovim?

Latest nightly vim

lervag:
How is this compared to the current master branch? Is the master branch still slow (as in O(10) seconds)?

lervag:
I'm tempted to merge this. Feel free to dissuade me, I'll wait until later today in any case.

lervag:
Thanks; I've merged this, then. :)

lervag:
I've also pushed quite a few updates that further improve on this. I've tested quite thoroughly on my side with neovim and Vim, and I believe things should still work well on MacOS and Windows. However, I am curious to hear feedback on this, especially from the Windows side (@habamax).

lervag:
Btw, some very crude benchmarking from my side:

* Neovim (`hyperfine -M 3 make`)

    commit   | Timing (s)  |  Note
    ------   | ----------  |  -----
    fb7c6d59 | 48.187 ± 0.221 | Latest version
    f40305e8 | 49.068 ± 0.079
    ef3b014d | 48.965 ± 0.031 | This PR merged
    ef3b014d | 49.217 ± 0.213
    aaecdbba | 48.282 ± 0.189 | Before this PR

* Vim (`MYVIM="vim -T dumb --not-a-term -n" hyperfine -M 3 make`)

    commit   | Timing (s)  | Note
    ------   | ----------  | -----
    fb7c6d59 | 51.026 ± 0.138 | Latest version
    f40305e8 | 52.302 ± 0.138
    ef3b014d | 52.239 ± 0.204 | This PR merged
    ef3b014d | 52.208 ± 0.153
    aaecdbba | 51.368 ± 0.097 | Before this PR

The commands are run inside `VIMTEX/test`; [hyperfine is here](https://github.com/sharkdp/hyperfine). Again, this is very crude and does not really say very much. Between aaecdbba and fb7c6d59 there are also added several new tests, for instance, so in a way one should expect the run time to increase. I believe these tests may (perhaps) be better than nothing for justifying a claim that the improvements _are_ improvements, especially considering that I think they strongly improve the robustness of the code.

bongbang:
Sorry for being late to the party. I got the email rather late to begin with and was not quite savvy enough with git and Github to download a branch for testing.  (I normally just let Vundle handle the updates.)

For future reference (so that I can do my part),  is the following roughly what I should do?

1. `cd` into `vimfiles/bundle`
2. `git clone https://github.com/lervag/vimtex/ --branch <branch> --single-branch vimtex-test`
3. Find a handy way to switch back and forth between two versions of VimTeX (maybe you can tell me). 

Thank you, Karl Lervåg , for ever diligent improvements.  So far, I haven't noticed much difference in the latest version except perhaps one bug that I've been meaning to report may have already be solved.

habamax:
@lervag well, now it takes more time with cleared cache (~ 30 secs) to show the completion for the same document :)

vimtex-jobs2


But when I save it under other path/name -- it is again "fast enough"


vimtex-jobs3

I wonder if it is possible to run background job once tex file is loaded to "construct" what is behind that cache.

lervag:
@bongbang

> Sorry for being late to the party.

No problem!

> For future reference (so that I can do my part), is the following roughly what I should do?

Yes, more or less. I would do it like this:

```sh
cd /path/to/vimtex
# e.g. ~/.vim/bundle/vimtex

# Pull latest updates, including all branches
git pull --all

# Checkout the test branch
git checkout <branch>

# Checkout back to master
git checkout master
```

This way, you don't need to work with two copies of the code. `git checkout` as such serves as a handy way to switch between two versions.

> Thank you, Karl Lervåg , for ever diligent improvements. So far, I haven't noticed much difference in the latest version except perhaps one bug that I've been meaning to report may have already be solved.

Great, I'm glad to hear this might have solved a bug, and I was not really expecting to hear of any noticable differences. Thank you for testing! And for the kind words :)

habamax:
> 
> > But when I save it under other path/name -- it is again "fast enough"
> 
> What do you mean here? "save under other path/name"?

`:w ~/docs/anotherpath/anothername.tex` then switch to it and do all the same

> 
> > I wonder if it is possible to run background job once tex file is loaded to "construct" what is behind that cache.
> 
> Perhaps. Could I propose that you again open a new issue for this? I'm sorry about the inconvenience, but I now know much more about what actually happens, and so we could continue the discussion and possibly find a way to fix it.

Sure will create it.


> One thing I'm thinking: Are you perhaps using WSL 2? Could it have something to do with the slowness of accessing files outside the linux partition?

Well, that is another issue I might create. For this particular case I have used `gVim` (and in general for tex I use it).

I also have WSL1 where completion works "instantly", but the issue is that compilation is not working there, vimtex can't find latemxk executable:

image

lervag:
> Sure will create it.

Great, let's continue this discussion there, then.

PR #2197, opened 2021-10-07
Generalize argument parsing for vimtex#cmd#get functions

lervag:
`vimtex#cmd#get*()` is a family of commands to parse a LaTeX command. The parser is generic and greedy, and it does not guarantee that it parses the "right amount". As it stands, the matching is not good enough, because it is restricted to the following pattern `\cmd[...]*{...}*`.

This PR changes the parsing to allow the following pattern: `\cmd\([...]\|{...}\|(...)\|<...>\)*`. This will for instance allow commands such as `\begin{textblock*}{3in}[0,0](3in,1in)` to be parsed as a single command with 4 different types of arguments. Also, it will allow to parse constructs such as `\test{arg}[opt]`.

Further, the current matching only "gobbles" the arguments when they follow directly. That is, there is no tolerance for white space or "commented new lines". That is, the following LaTeX snippets are equivalent, but only the first of them is currently properly parsed:

```tex
\begin{textblock*}{3in}[0,0](3in,1in)

\begin{textblock*}{3in}[0,0]%
    %
    %
    (3in,1in)%
```

It could be of interest to also relax the current "no whitespace" restriction.

Opinions? @clason @reportaman


lervag:
There seems to be little interest for this, and I'm now working on a different idea for parsing in general. Thus, I'll close this as unwanted.

PR #2209, opened 2021-10-16
fix chktex compiler setup and error parsing

Konfekt:
Previous set up used invalid command-line arguments. Now detects column number.


lervag:
Thanks; I've made some minor changes and merged.

lervag:
I agree with the skepticism of the value of `CompilerSet`, so I just ignored it and use the setlocal mechanism instead.

Issue #2217, opened 2021-10-23
Allow synctex backwards search when there are multiple vim sessions


user202729:
Currently to use synctex it's necessary to specify `--servername` correctly.

It would be better (and possible?) if the command is send to all vim instances, then each can handle it correspondingly.

(Would require `remote_startserver` for each vim instance that does not start, and some vim-side implemented function to iterate over the existing servers, then on each server check if there's any open buffer with the matching name and move the cursor on that buffer)

There are some workarounds such as in https://github.com/lervag/vimtex/issues/1576#issuecomment-586752584 but even that one does not support multiple instance of vim run at once (although this is not particularly frequent use-case)

user202729:
Okay I just pulled the latest version (actually I already read that trick, coming across the GitHub comment)

Remark: I think in Windows cmd can access the temp folder as simply `%TEMP%` or `%TMP%` (not tested)

----------

Yes that's possible, but still if some user "happen to" run multiple TeX files with opened PDF file simultaneously it's still not ideal.

For the script approach... it's still possible to use vim itself (instead of Python script or something) as the script interpreter, and store the script in the plugin's autoload or something. It only assumes vim is in path.

PR #2219, opened 2021-10-25
Implement better inverse search

lervag:
This PR is inspired by #2217. The idea is to implement the inverse search functionality within Vimscript for both Vim and neovim, so that the viewer configuration can be simplified. I.e., when this is finished, a viewer can run something like this:

```sh
# For Vim
vim -T dumb --not-a-term -n -c "call vimtex#view#inverse_search_comm(%l, '%f')"

# For neovim
nvim --headless -c "call vimtex#view#inverse_search_comm(%l, '%f')"
```

Then the function `inverse_search_comm` will pass this on to all available Vim/neovim instances. The idea is that each instance will inspect the received arguments and only perform the inverse search if relevant.

Pseudocode:

```
vimtex#view#inverse_search_comm:

for server in serverlist
  pass_to_server: vimtex#view#inverse_search(line, file)

quit
```

and

```
vimtex#view#inverse_search:

if (not in vimtex buffer) or (not in correct vimtex project): return

navigate to: line and file
```

Remaining tasks:
* [x] Implement `vimtex#view#inverse_search_comm`
* [x] Implement `vimtex#view#inverse_search`
* [x] Update documentation of inverse search settings
* [x] Thoroughly test the new feature with both neovim and Vim on all platforms
  * [x] Vim
    * [x] Linux
    * [x] MacOS
    * [x] Windows (if possible)
  * [x] neovim
    * [x] Linux
    * [x] MacOS
    * [x] Windows (if possible)


PanagiotisS:
## minimal init.vim

```vim
set encoding=utf-8
scriptencoding utf-8

" Dein {{{
let &runtimepath .= ',' . expand(stdpath('data') . '/dein/repos/github.com/Shougo/dein.vim/')
let s:path = expand(stdpath('data') . '/dein')

if dein#load_state(s:path)
    call dein#begin(s:path)

    call dein#add('Shougo/dein.vim')
    call dein#add('lervag/vimtex', {'rev': 'feat/better-inverse-search'})

    call dein#end()
    call dein#save_state()
endif
" }}}

let g:vimtex_view_general_viewer = 'C:\Users\ps\AppData\Local\SumatraPDF\SumatraPDF'
let g:vimtex_view_general_options = '-reuse-instance -forward-search @tex @line @pdf'
let g:vimtex_view_general_options_latexmk = '-reuse-instance'

"
set number
set ruler
set visualbell
set background=dark
```

# Setup
## OS
Windows 10

## pdfviewer
Sumatra v3.3.3
Settings: 
image

## Neovim
`NVIM v0.6.0-dev+3-gd791274a9`

```vim

health#vimtex#check
========================================================================
## VimTeX
  - OK: Vim version should have full support!
  - OK: General viewer should work properly!
  - OK: Compiler should work!
 ....
```

# Comments

## Issue 1

It seems that psutils does not work.

Calling from the terminal `nvim --headless -c "VimtexInverseSearch 8 'test.tex'"` it seems that the `l:socket_ids` variable is empty in function `s:inverse_search_cmd_nvim`. 

Testing psutils from a python interpeter

```python
>>> for proc in (p for p in psutil.process_iter(attrs=['name']) if 'nvim' in p.info['name']):
...  print(proc)
...
psutil.Process(pid=600, name='nvim.exe', status='running', started='01:35:12')
psutil.Process(pid=6296, name='nvim.exe', status='running', started='23:40:55')
psutil.Process(pid=8164, name='nvim.exe', status='running', started='22:31:17')
psutil.Process(pid=8168, name='nvim.exe', status='running', started='01:28:26')
psutil.Process(pid=10244, name='nvim-qt.exe', status='running', started='01:35:12')
```
We can see that on windows we get 'nvim.exe' and 'nvim-qt.exe' in case of GUI.

Howevre, `proc.connections` seems that it does not have 'unix'. 

```python
>>> proc.connections('unix')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\__init__.py", line 1162, in connections
    return self._proc.connections(kind)
  File "C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py", line 679, in wrapper
    return fun(self, *args, **kwargs)
  File "C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py", line 1005, in connections
    return net_connections(kind, _pid=self.pid)
  File "C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py", line 356, in net_connections
    raise ValueError("invalid %r kind argument; choose between %s"
ValueError: invalid 'unix' kind argument; choose between 'all', 'tcp', 'tcp4', 'udp', 'udp4', 'inet', 'inet4', 'inet6', 'tcp6', 'udp6'
```

Maybe is related to https://github.com/giampaolo/psutil/issues/1649 ?
Testing instead with `all` get get 0 connections. 

```python
In [14]: procs = [proc for proc in (p for p in psutil.process_iter(attrs=['name']) if 'nvim' in p.info['name'])]

In [15]: procs[0]
Out[15]: psutil.Process(pid=600, name='nvim.exe', status='running', started='01:35:12')

In [16]: procs[0].connections('all')
Out[16]: []
```

I am afraid that I am not sure how to proceed from here without using pynvim and especially on windows OS.

If I comment out the [relevant python part](https://github.com/lervag/vimtex/pull/2219/files#diff-d20929521b785dff96b2ca148df4e60a112e9daa746c81c403ab8f076d41044aR142-R150) and instead replace the socket_ids with the hard-coded result of `v:servername` of the neovim instance with 'test.tex' 

``` vim
let l:socket_ids = ['\\.\pipe\nvim-2064-0']
```

and subsequently run from the terminal 
``` terminal
nvim --headless -c "VimtexInverseSearch 8 'test.tex'"
````
it works as expected.


## Issue 2

However, when is run from the SumatraPDF it instead runs the command
```
 nvim --headless -c "VimtexInverseSearch 8 'C:\Users\ps\Documents\MyDocuments\test.tex'"
````

So is using the full path and I am afraid that it is not working.
I do not have the time right now to go through the code for this one. It is getting quite late here.
Please let me know if you need someone to do it and I can have a look.

##
I hope the above are useful.
Cheers

**Comment by BertrandSim** [comment]
Hi lervag, I am on Windows, vanilla vim, SumatraPDF. I have followed the instructions in your reddit post (mentioned above). 

However, After 
 1. changing the inverse search cmd within SumatraPDF, and 
 2. double clicking somewhere in the pdf file, 

A blank cmd window opens to vim, the cursor stays on the first line for a second, then moves to the last line of the cmd window, for a brief moment, then exits. Nothing else happened I'm afraid.. Existing vim windows did not change at all.

**Comment by lervag** [comment]
@PanagiotisS Thank you! I'm so glad to see your comment (even though it things are not working quite as well as I was hoping). I appreciate your attention to the details and that you went deep in trying to understand what is happening!

> It seems that psutils does not work.

Ah, yes; I was actually expecting this. The `psutil` snippet was inspired by `neovim-remote`, see [here](https://github.com/mhinz/neovim-remote/blob/master/nvr/nvr.py#L371-L382). My idea was that `nvr --serverlist` seemed to work, and the above linked code is the code used to get the list of servers. Do the `inet4` or `inet6` parts perhaps work on Windows?

> We can see that on windows we get 'nvim.exe' and 'nvim-qt.exe' in case of GUI.

Or does the code perhaps stop already here with `proc.info['name'] == 'nvim'`?

> I am afraid that I am not sure how to proceed from here without using pynvim and especially on windows OS.

The point of this stuff is to get a list of all servers. However, two alternatives could work here:

* Fallback to the env var `NVIM_LISTEN_ADDRESS`. If you open the viewer from within neovim, then perhaps it would also inherit this env var, as happens on linux/macos? (The code for this is already there, though, i.e. [here](https://github.com/lervag/vimtex/pull/2219/files#diff-d20929521b785dff96b2ca148df4e60a112e9daa746c81c403ab8f076d41044aR152))

* Perhaps we could figure out a _better_ way to get the list of servers? I don't know how, but perhaps it would be possible through Lua or something?

> If I comment out ... and subsequently run from the terminal ... it works as expected.

Nice, good to hear.

> However, when is run from the SumatraPDF it instead runs the command
> 
> ```
>  nvim --headless -c "VimtexInverseSearch 8 'C:\Users\ps\Documents\MyDocuments\test.tex'"
> ```
> 
> So is using the full path and I am afraid that it is not working.

Ok; is this path the correct path? If so, then I believe there may be some issue on the receiving part, i.e. somewhere here:

https://github.com/lervag/vimtex/blob/868d021239482d94e26751dfaf268bfc795d28c3/autoload/vimtex/view.vim#L62-L72

If you have the time/possibility, perhaps you could investigate like this:

* First, make sure you use the hard coded servername as you already tried and that you have the correct neovim instance with the LaTeX source available.
* At the top of the function `...#inverse_search()`, add a simple `unsilent echo 'DBG'` and then `return`, and check if the inverse search from SumatraPDF end up calling the function inside the correct neovim instance. If it does, the echo should be visible and we can go on. If not, then there is something strange going on.
* Now, inspect the input variables. E.g. `echo a:line` and `echo a:filename`. Go deeper with things like `echo resolve(a:filename)` or similar. Try to locate where the issue originates.

> I do not have the time right now to go through the code for this one. It is getting quite late here. Please let me know if you need someone to do it and I can have a look.

Again, thank you! I don't use Windows, so I definitely need users to help out here.

This PR is not "important", and there is no rush. I'll be patient.

**Comment by lervag** [comment]
> A blank cmd window opens to vim.exe, the cursor stays on the first line for a second, then moves to the last line of the cmd window, for a brief moment, then exits. Nothing else happened I'm afraid.. Existing vim windows did not change at all.

Thanks @BertrandSim, I much appreciate that you've taken the time to test this! Can you try the following variant:

```
cmd /c vim -T dumb --not-a-term -n -c "VimtexInverseSearch %l '%f'"
```

**Comment by PanagiotisS** [comment]
> Do the inet4 or inet6 parts perhaps work on Windows?

None of them work. 
In my understanding "all" should include all possibilities,

> [`"all"`: the sum of all the possible families and protocols](https://psutil.readthedocs.io/en/latest/index.html?highlight=connections#psutil.Process.connections)

```python
In [3]: procs = [proc for proc in (p for p in psutil.process_iter(attrs=['name']) if 'nvim' in p.info['name'])]
In [4]: procs
Out[4]: [psutil.Process(pid=6884, name='nvim.exe', status='running', started='00:55:30')]
In [7]: procs[0].connections('inet4')
Out[7]: []

In [8]: procs[0].connections('inet6')
Out[8]: []

In [9]: procs[0].connections('inet')
Out[9]: []

In [10]: procs[0].connections('tcp')
Out[10]: []

In [11]: procs[0].connections('tcp4')
Out[11]: []

In [13]: procs[0].connections('tcp6')
Out[13]: []

In [14]: procs[0].connections('udp')
Out[14]: []

In [15]: procs[0].connections('udp4')
Out[15]: []

In [16]: procs[0].connections('udp6')
Out[16]: []

In [17]: procs[0].connections('unix')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-3aaa9af8e6a5> in <module>
----> 1 procs[0].connections('unix')

C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\__init__.py in connections(self, kind)
   1160         +------------+----------------------------------------------------+
   1161         """
-> 1162         return self._proc.connections(kind)
   1163
   1164     # --- signals

C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py in wrapper(self, *args, **kwargs)
    677     def wrapper(self, *args, **kwargs):
    678         try:
--> 679             return fun(self, *args, **kwargs)
    680         except OSError as err:
    681             raise convert_oserror(err, pid=self.pid, name=self._name)

C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py in connections(self, kind)
   1003     @wrap_exceptions
   1004     def connections(self, kind='inet'):
-> 1005         return net_connections(kind, _pid=self.pid)
   1006
   1007     @wrap_exceptions

C:\tools\miniconda3\envs\ppt\lib\site-packages\psutil\_pswindows.py in net_connections(kind, _pid)
    354     """
    355     if kind not in conn_tmap:
--> 356         raise ValueError("invalid %r kind argument; choose between %s"
    357                          % (kind, ', '.join([repr(x) for x in conn_tmap])))
    358     families, types = conn_tmap[kind]

ValueError: invalid 'unix' kind argument; choose between 'all', 'tcp', 'tcp4', 'udp', 'udp4', 'inet', 'inet4', 'inet6', 'tcp6', 'udp6'

In [18]: procs[0].connections('all')
Out[18]: []
```

> Or does the code perhaps stop already here with proc.info['name'] == 'nvim'?

As a first step I changed the following line
https://github.com/lervag/vimtex/blob/868d021239482d94e26751dfaf268bfc795d28c3/autoload/vimtex/view.vim#L146-L147

As `p.info['name'] == 'nvim'` returns an empty array, I changed it into `'nvim' in p.info['name']`.
Maybe is very generic conditional but for my testing was sufficient.
(Btw, out of curiosity, I installed and tested Neovide, and it seems that the process name is `nvim.exe`, which is good I guess.)

Then, when trying to access `process('unix')` the code was crashing and therefore I commented out the the whole part and hard coded the `v:servername` I wanted to use. I suppose you can call it Issue 1a, and 1b the lack of connections.

> Fallback to the env var NVIM_LISTEN_ADDRESS. If you open the viewer from within neovim, then perhaps it would also inherit this env var, as happens on linux/macos? (The code for this is already there, though, i.e. here)

If open `nvim test.tex` and then from within nvim
If I write `:echo $NVIM_LISTEN_ADDRESS` I get the same as `v:servername` (as expected)
If I write `echo $NVIM_LISTEN_ADDRESS_VIMTEX` I get nothing. Tab completion is not working either so is not set.

I will look further into this.

Also, if is an environmental variable which we set when we open neo/vim, then that variable should not be accessible from a different terminal?

> The psutil snippet was inspired by neovim-remote, see here. 

Here is a PR which tries to patch the plugin for windows. 

- For nvim.exe

Similar change as the suggested

[`if proc.info['name'] in ['nvim', 'nvim.exe']:`](https://github.com/mhinz/neovim-remote/pull/155/files#diff-946319562aae58a650bd19f1a1fdb2f2fb525657e4724d975349f67bbb50c59cR370-R379)

- For psutil `connection('unix')`
Within a conditional to avoid errors

```python
                if os.name == 'posix':
                    for conn in proc.connections('unix'):
                        if conn.laddr:
                            addresses.insert(0, conn.laddr)
```
https://github.com/mhinz/neovim-remote/pull/155/files#diff-946319562aae58a650bd19f1a1fdb2f2fb525657e4724d975349f67bbb50c59cR376

- For the listening address

They set a default listening address and they mention that exact problem.

```python
    useraddr = options.servername or env.get('NVIM_LISTEN_ADDRESS')
    # Since before build 17063 windows doesn't support unix socket, we need another way
    defaddr = '127.0.0.1:6789' if os.name == 'nt' else '/tmp/nvimsocket'

    address = useraddr or defaddr
```
https://github.com/mhinz/neovim-remote/pull/155/files#diff-946319562aae58a650bd19f1a1fdb2f2fb525657e4724d975349f67bbb50c59cR411-R412

The project's README.md mentions [This requires
forking, so it won't work on Windows.](https://github.com/mhinz/neovim-remote/blob/master/README.md?plain=1#L42-L43). So does this mean the whole plugin does not work with Windows or the specified feature only? Looking at the PR it most likely means the latter.


> Perhaps we could figure out a better way to get the list of servers? I don't know how, but perhaps it would be possible through Lua or something?

I will also look further into psutil and the links you mentioned. I am not familiar with this but I have some free time these days.
Maybe I can try using a default address like the aforementioned PR. However, in this plugin's case it does not seem like a good idea as you can have multiple instances of vim open when writing latex. Maybe split main and chapters, or bibliography, or an old archive to compare with for examples etc etc. (Sure you can use splits, tabs, etc etc but we cannot tell others how should they do things). Maybe we can check if (the default) address is in use and increment by one each time.
I think we need a more elegant solution.


## 

> is this path the correct path?
When using just the name of the tex file it is indeed the correct path. Also, considering that the script works it makes sense. In any case I did use `echom` for `a:line` and `a:filename` and they had the correct values. 

When using the full path, I did not do any further testing.

I will look into this issue. It should be fairly easy to fix.

user202729:
@BertrandSim @lervag Regarding the popup window perhaps replacing `vim` with `start /b vim` (do not create a new window for the process) would work.

BertrandSim:
> Thanks @BertrandSim, I much appreciate that you've taken the time to test this! Can you try the following variant for the SumatraPDF setting:
> 
> ```
> cmd /c vim -T dumb --not-a-term -n -c "VimtexInverseSearch %l '%f'"
> ```
> 
> Does "vanilla vim" mean that you typically use gvim? Perhaps you could test with `gvim ...` instead of `vim ...` as well?

No problem @lervag, happy to help :) 

By "vanilla vim" I meant that I'm not on neovim :S  

Prepending `cmd /c` still gave the same results. A cmd window opens briefly to vim.exe, then exits.

When I changed the command in sumatra to 
```
vim -c "VimtexInverseSearch %l '%f'"
```
I see my vim instance starting for a brief second (the colorscheme is loaded), then exits.
The same happens when I replace 'vim' with 'gvim'.

Thus, I think vim is initializing correctly, with or without 'cmd /c'.

Now, while viewing the .tex file in gvim/vim, I ran `:VimtexInverseSearch <line> '<file>'`, with `<line>` being a small number, and `<file>` being the current file (without the .pdf extension). Strangely, the gvim window exits / vim session closes!

user202729:
> Strangely, the gvim window exits / vim session closes!

It's not strange, it's by design. (that command sends the command to other instances then exit vim. Read the documentation)

`start /b` should not popup the window, try it (although I think there may remain some psutil issue, perhaps it only affect neovim?)

BertrandSim:
> It's not strange, it's by design. (that command sends the command to other instances then exit vim. Read the documentation)

Sorry about that, I read the revised docs but that part was not obvious to me.

> `start /b` should not popup the window, try it (although I think there may remain some psutil issue, perhaps it only affect neovim?)

Oh yes, I should have mentioned that I have tried `start /b vim ...`previously. I got an error from SumatraPDF "Cannot start inverse search command. Please check the command line in the settings."

user202729:
Looking around a little more looks like that the only other solution is to use vbscript.

At least it's built-in to Windows... but it's not that clean code-wise.

PanagiotisS:
Regarding the path it was a quite simple indeed.

`call map(l:sources, {_, x -> b:vimtex.root . '/' . x})`
returns 
`echom string(l:sources)` -> `['C:\Users\ps\Documents\MyDocuments\cv/test.tex']`
while 
`echom l:file` ->  `C:\Users\ps\Documents\MyDocuments\cv\test.tex`
and therefore
`    if index(l:sources, l:file) < 0 | return -2 | endif`
returns `-1`

If we replace the map with, for example, `globpath` (is `globpath` the correct function?)

```
    " call map(l:sources, {_, x -> b:vimtex.root . '/' . x})
    call map(l:sources, {_, x -> globpath(b:vimtex.root, x)})
```
Then it works.

(In my experience the issue most of the times with Windows is that it is using backslash `\` for the paths and therefore it either needs to be escaped `\\` or there is somewhere a hard-coded `/`.)

With hard coded the servername I can now double click the pdf (with SumatraPDF) and I get to the line in the opened tex file.
It works with both nvim and nvim-qt.
(For the testing one can use for example `nvim .\test.tex --listen '\\.\pipe\nvim-2444-0'`)

This issue should affect vim as well so hopefully now vim should work without any further issues.

P.S. A terminal window shows up briefly everytime I double click a line as it was mentioned above for me as well. I did not consider this to be an issue so I did not mention it earlier.

andryandrew:
I tested it on MacOS, with Vim/MacVim/neovim + Skim.
1) "vanilla" Vim: AFAICT, `clientserver` requires X11 to work and won't work in terminal vim. As usual, community support for MacOS is scattered and difficult to obtain, so I might be wrong here.
2) MacVim: `clientserver` functionality is implemented in the "graphical" interface (launching with `mvim`) but does not work in the terminal (`mvim -v`). This is expected (`:h macvim-clientserver`). I followed the steps in `:h vimtex-clientserver` to configure my `.vimrc`.
- Using the suggested configuration for Skim (command: `mvim`, arguments: `-T dumb --not-a-term -n -c "VimtexInverseSearch %line '%file'"`) does not work: it just open a new empty MacVim window instead of initiating the backwards search
- This is easily fixed by adding `-v` to the arguments, i.e. `-v -T dumb --not-a-term -n -c "VimtexInverseSearch %line '%file'"`. Now, inverse search works when starting Skim from MacVim (with `<localleader>lv`). It also works with multiple files open, both as buffers in the same mvim instance and as separate instances. In the former case, inverse search is only initiated if the corresponding buffer has the focus, which (I guess?) is intended behavior... personally, I would like it to work even when the buffer is unfocused, if it is possible. But I can always start separate instances.
3) neovim: in this case it just works™ with no configuration needed on neovim side. Dropping the dependency on `nvr` is really nice. It seems also even snappier than before.
- In Skim, command: `nvim`, arguments: `--headless -c "VimtexInverseSearch %line '%file'"`. Everything works with multiple files when opened in the same neovim instance, with the same limitation as in MacVim (buffer has to be focused).
- However opening multiple neovim instances does not work for me. In this case, inverse search in the second file does not work, while the first one still does. I will try later to put some debug `echo`s to see if the event is received.

PanagiotisS:
Addressing psutils issue on Windows

Using `psutils.environ()` we can access the environmental variables of the said process and therefore the `proc.environ()['NVIM_LISTEN_ADDRESS_VIMTEX']`

```vim
140 function! s:inverse_search_cmd_nvim(line, filename) abort " {{{1
141   if empty($NVIM_LISTEN_ADDRESS_VIMTEX)
142       if has('win32')
143          py3 << EOF
144 import psutil
145
146 sockets = [
147     proc.environ()["NVIM_LISTEN_ADDRESS_VIMTEX"]
148     for proc in (
149         p for p in psutil.process_iter(attrs=["name"]) if "nvim" in p.info["name"]
150     )
151     if proc.environ().get("NVIM_LISTEN_ADDRESS_VIMTEX")
152 ]
153 EOF
154       else
155          py3 << EOF
156 import psutil
157
158 sockets = []
159 for proc in (p for p in psutil.process_iter(attrs=['name'])
160              if p.info['name'] == 'nvim'):
161     sockets += [c.laddr for c in proc.connections('unix') if c.laddr]
162 EOF
163     endif
164     let l:socket_ids = filter(py3eval('sockets'), 'v:val != v:servername')
165   else
166     let l:socket_ids = [$NVIM_LISTEN_ADDRESS_VIMTEX]
167   endif
```

I suppose this should work on linux as well so maybe we can keep only that?

In any case, with these two changes (https://github.com/lervag/vimtex/pull/2219#issuecomment-957537209) inverse search works without problems.

Tested with both nvim, nvim-qt.
Tested with having three neovim instances open, "test1.tex", "test2.tex", "view.vim" open and pdf opened with SumatraPDF "test1.pdf", "test2.pdf".

The only issue is that we need pynvim and psutil installed.

setting in init.vim
```vim
" Conda env with Python 3.9 and pynvim, and psutil installed - works
let g:python3_host_prog = 'C:\tools\miniconda3\envs\ppt\python.exe'

" Conda env with Python 3.9 and only psutil installed - does not work
let g:python3_host_prog = 'C:\tools\miniconda3\envs\test_psutil\python.exe'

" Python 3.9 without any packages installed - does not work
let g:python3_host_prog = 'C:\Python39\python.exe'
```

BertrandSim:
Here's an update for Windows, vim/gvim, SumatraPDF:

Following PanagiotisS' comment https://github.com/lervag/vimtex/pull/2219#issuecomment-957537209,
> If we replace the map with, for example, globpath (is globpath the correct function?)
>
>```
>   " call map(l:sources, {_, x -> b:vimtex.root . '/' . x})
>    call map(l:sources, {_, x -> globpath(b:vimtex.root, x)})
>```
>Then it works.

I modified the line as per above in vimtex/autoload/vimtex/view.vim . 
Initially, using `globpath()` did not work for me. However, when I used a hardcoded backslash, as in

```
  call map(l:sources, {_, x -> b:vimtex.root . '\' . x})
```

The inverse search works for me. 

It turns out that this is due to my `'shellslash'` setting, which I had _set_ (it is _not set_ by default for windows users). Once I had `set 'noshellslash'`, the inverse search works for both cases, `\` and `globpath()`.

(tested with a single gvim instance, and on a single tex file)

BertrandSim:
BTW, there was a discussion above about the pop up terminal window. I would like to clarify that that is a non-issue for me.

lervag:
Thank you all for the useful input! There was a lot of input here, and I'll try to respond to everything. If I missed something or someone needs additional clarification, please feel free to ask.

### Windows

@BertrandSim

Thanks for pitching in and for verifying that things work if we resolve the bug found by @PanagiotisS. And also for the clarification that the popup window is not a big issue, although I would be happy if we can resolve it.

@user202729

As you surmised, the `psutil` stuff is only relevant for neovim. Thanks for pitching in and helping here!

It would be nice if we can find a good command line for avoiding the terminal popup. It should be possible to test your various suggestions and tweak until it works if we just call the things from another terminal. So, perhaps one could try that and use different variants, e.g.:

```cmd
cmb /b start /b ...
start /b "vim ...."
```

> Actually there are some (vim-specific) solutions on vim fandom and other sites, using -s or something. I didn't test however.

I don't think the `-e -s` argument will work; I tried, but I did not find a way. But perhaps I'm just missing something silly...

@PanagiotisS

> > Do the inet4 or inet6 parts perhaps work on Windows?
> 
> None of them work. In my understanding "all" should include all possibilities,

Great, thanks for looking into it!

> As `p.info['name'] == 'nvim'` returns an empty array, I changed it into `'nvim' in p.info['name']`. Maybe is very generic conditional but for my testing was sufficient.

Good idea. The alternative `p.info['name'] in ('nvim', 'nvim.exe')` looks better, though.

> However, running the command from the terminal the `NVIM_LISTEN_ADDRESS_VIMTEX` is empty.
> 
> Shouldn't this be expected? As this is an environmental variable which we set when we open neo/vim, then that variable should not be accessible from a different terminal?

Yes; but there is a significant trick here: If you open SumatraPDF from neovim, then it should inherit the environment variables. At least, this is how things work on Linux. Similarly, if you open a terminal from within neovim, the terminal should have the `NVIM_LISTEN_ADDRESS` inherited, and possibly the `_VIMTEX` one as well if it was created by VimTeX.

> They set a default listening address and they mention that exact problem.

Yes, this reference seems to indicate that we can't properly resolve this right now. So, perhaps we could fall back to something inspired by @jdhao here. That is, instead of doing the python stuff, we can write the servernames to a file when VimTeX is initialized. We simply append to the end of the file. Then the `...#inverse_search_cmd` function reads the servernames from the file and cleans out the servernames from the file that don't exist. That should solve this part for neovim in most/all cases, I think. And I think it should be efficient and then we wouldn't really need the `NVIM_LISTEN_ADDRESS_VIMTEX` trick either.

> However, in this plugin's case it does not seem like a good idea as you can have multiple instances of vim open when writing latex. Maybe split main and chapters, or bibliography, or an old archive to compare with for examples etc etc. (Sure you can use splits, tabs, etc etc but we cannot tell others how should they do things). Maybe we can check if (the default) address is in use and increment by one each time. I think we need a more elegant solution.

Yes, you might be right. In this case, the env var option is not so bad, because it ensures that the viewer communicates with the neovim instance that opened the viewer. This works perfectly on Linux, and I would think it should work on Windows as well (but I may be wrong).

> Regarding the path it was a quite simple indeed.

Thank you! This does clarify things!

> If we replace the map with, for example, `globpath` (is `globpath` the correct function?)

Can you check if `resolve()` and or `simplify()` works here, i.e.

 ```
     call map(l:sources, {_, x -> resolve(b:vimtex.root . '/' . x)})
 ```

Or perhaps avoid shellslash; i.e. if we add `set noshellslash` just before? If that works, then we can temporarily disable shellslash within this function.

> With hard coded the servername I can now double click the pdf (with SumatraPDF) and I get to the line in the opened tex file. It works with both nvim and nvim-qt.

Progress!

> P.S. A terminal window shows up briefly everytime I double click a line as it was mentioned above for me as well. I did not consider this to be an issue so I did not mention it earlier.

If we can find an adjusted command line to prevent this terminal window from popping up, than that would be useful. Did you try the suggestions by @user202729?

> Using `psutils.environ()` we can access the environmental variables of the said process and therefore the `proc.environ()['NVIM_LISTEN_ADDRESS_VIMTEX']`

That's a neat idea! Similar to the one I presented above; I'm not sure which I like the best. The one you propose here should work on all platforms, I believe, so perhaps it is better. I'll test it and push it if it works, as it seems to resolve most of this thread so far. Then I'll consider the file based approach afterwards.

> The only issue is that we need pynvim and psutil installed.

`psutil` is a standard lib, so I think in reality the dependency here is python3 and pynvim. And those seem quite common and I don't expect it to be considered a problem.

### MacOS

@andryandrew

> I tested it on MacOS, with Vim/MacVim/neovim + Skim.

Nice! I believe @clason already tested this with neovim; nice to get feedback on Vim and MacVim!

> "vanilla" Vim: AFAICT, `clientserver` requires X11 to work and won't work in terminal vim. As usual, community support for MacOS is scattered and difficult to obtain, so I might be wrong here. If someone has a working `clientserver` configuration for Vim on MacOS, I am happy to test it.

Ok; TIL. I guess most people on MacOS uses MacVim or neovim, so I'll not care so much about this.

> MacVim: the server part of the `clientserver` functionality is implemented in the "graphical" interface (launching with `mvim`) but does not work in the terminal (`mvim -v`). This is expected (`:h macvim-clientserver`). I followed the steps in `:h vimtex-clientserver` to configure my `.vimrc`.

But is this not very much similar to standard Vim in terminal, which also does not start the server unless explicitly stated? That is, would `mvim --servername test` work in the terminal and start a server?

> Using the suggested configuration for Skim (command: `mvim`, arguments: `-T dumb --not-a-term -n -c "VimtexInverseSearch %line '%file'"`) does not work: it just open a new empty MacVim window instead of initiating the backwards search
>
> This is easily fixed by adding `-v` to the arguments, i.e. `-v -T dumb --not-a-term -n -c "VimtexInverseSearch %line '%file'"`. Now, inverse search works when starting Skim from MacVim (with `<localleader>lv`).

Great, I'll add that to the docs; thanks!

> It also works with multiple files open, both as buffers in the same mvim instance and as separate instances. In the former case, inverse search is only initiated if the corresponding buffer has the focus, which (I guess?) is intended behavior... personally, I would like it to work even when the buffer is unfocused, if it is possible. But one can always start separate instances.

This is something we can address and work more on later. But yes, it is by design. If you want to change this, could you open a feature request after this PR has been merged?

> neovim: in this case it just works™ with no configuration needed on neovim side. Dropping the dependency on `nvr` is really nice. It seems also even snappier than before.

Very glad to hear it!

> However opening multiple neovim instances does not work for me. In this case, inverse search in the second file does not work, while the first one still does. I will try later to put some debug `echo`s to see if the event is received.

This is also partly by design. You opened Skim from neovim instance 1 - then Skim inherits the `NVIM_LISTEN_ADDRESS_VIMTEX` environment variable and neovim uses it directly. Thus neovim instance 2 and so on will always be ignored. Again, I think this can be left for after PR merge, as I agree it may not be properly thought out (i.e., if you open neovim + LaTeX file, then Skim, then close neovim and open again, then the inverse search will no longer work because `NVIM_LISTEN_ADDRESS_VIMTEX` is defined but the target server is killed).

lervag:
Ok; with the updated version, I believe things should work on Windows with Vim and neovim, as well as on MacOS with macvim and neovim. Remaining points:

* [ ] Verify that I'm right and that it works.
* [ ] For neovim: Consider if we should manually write servernames to a file and bookkeep this stuff ourselves. It seems simple and will avoid the Python dependency; it should work on all OSes and I think it is a good idea. Perhaps it should come after this PR, though.
* [ ] Check if we can avoid the popup on Windows.

PanagiotisS:
## Regarding psutil

> Good idea. The alternative p.info['name'] in ('nvim', 'nvim.exe') looks better, though.

The disadvantage of this is that it does not include 'nvim-qt.exe', which we can add.
I checked Neovide and it's process name is 'nvim.exe'. 
However, they are plenty more GUIs out there and we either have to check them and keep an up to date list or wait for someone to complain that it is not working with X GUI.

On the other hand 'nvim' in 'name' is too broad, however if we are going to check afterwards if it has $NVIM_LISTEN_ADDRESS or $NVIM_LISTEN_ADDRESS_VIMTEX or something else then I think we are good.

> Yes; but there is a significant trick here: If you open SumatraPDF from neovim, then it should inherit the environment variables. At least, this is how things work on Linux. Similarly, if you open a terminal from within neovim, the terminal should have the NVIM_LISTEN_ADDRESS inherited, and possibly the _VIMTEX one as well if it was created by VimTeX.

In my opinion this is a big IF. For example I never do. Especially on Linux as pdf viewers tend to report a lot of warnings on the terminal and they annoy me a lot, so usually I open a new terminal, open the pdf file and close the said terminal. In any case I think the solution with psutil works fine so we do not have to worry about this.

> psutil is a standard lib, so I think in reality the dependency here is python3 and pynvim. And those seem quite common and I don't expect it to be considered a problem.

On my windows installation for both python 3.9 and 3.10, and fresh Conda installations 3.9 and 3.10 is not installed by default and must be installed. (Relevant random result form SO which shows more people having the same issue https://stackoverflow.com/questions/50316358/error-no-module-named-psutil). Here also are the instruction how to install it https://github.com/giampaolo/psutil/blob/master/INSTALL.rst . Also, the psutil docs are not part of the official docs (which is usually how I check if something is part of the standard).

I do not consider them a problem either, however it should be mentioned on the README/doc for anyone interested.
For example, I do not use any plugin with pynvim dependency so I would be wondering what is going on.

> Yes, this reference seems to indicate that we can't properly resolve this right now. So, perhaps we could fall back to something inspired by @jdhao here. That is, instead of doing the python stuff, we can write the servernames to a file when VimTeX is initialized. We simply append to the end of the file. Then the ...#inverse_search_cmd function reads the servernames from the file and cleans out the servernames from the file that don't exist. That should solve this part for neovim in most/all cases, I think. And I think it should be efficient and then we wouldn't really need the NVIM_LISTEN_ADDRESS_VIMTEX trick either.

That would remove the psutil and pynvim dependency which is nice but it would add reading and writing files which is I personally prefer even less. In any case, if you decide on this approach, neovim has `stdpath('cache')`  (see `:help stdpath`) which can be used as a default.

### Comments on the commits

```python
148 sockets = [
149     p
150     for p in psutil.process_iter(attrs=["name"])
151     if p.info["name"] in ("nvim", "nvim.exe")
152 ]
153 print(sockets)
```

This is what I get from the print

```
[
psutil.Process(pid=9412, name='nvim.exe', status='running', started='00:06:29'), 
psutil.Process(pid=9748, name='nvim.exe', status='running', started='00:01:31'), 
psutil.Process(pid=12420, name='nvim.exe', status='running', started='23:53:44'), 
psutil.Process(pid=12852, name='nvim.exe', status='running', started='23:53:24')
]
```
Which is not what we want. Also when vim tries to run filter on this, it fails.
We need the extra bit where we get the "$NVIM_LISTEN_ADDRESS_VIMTEX"

If I run the same in iPython

```python
In [3]: for p in psutil.process_iter(attrs=["name"]):
   ...:     if p.info["name"] in ("nvim", "nvim.exe"):
   ...:         print(p)
   ...:
psutil.Process(pid=9748, name='nvim.exe', status='running', started='00:01:31')
psutil.Process(pid=12420, name='nvim.exe', status='running', started='23:53:44')
psutil.Process(pid=12852, name='nvim.exe', status='running', started='23:53:24')

In [9]: sockets[0].environ()["NVIM_LISTEN_ADDRESS_VIMTEX"]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-9-c78047347fff> in <module>
----> 1 sockets[0].environ()["NVIM_LISTEN_ADDRESS_VIMTEX"]

KeyError: 'NVIM_LISTEN_ADDRESS_VIMTEX'

In [10]: sockets[1].environ()["NVIM_LISTEN_ADDRESS_VIMTEX"]
Out[10]: '\\\\.\\pipe\\nvim-12420-0'

```

In this way we can simply filter and keep only nvim instances which have "NVIM_LISTEN_ADDRESS_VIMTEX".
(Which also means that using `nvim in p.info["name"]` is fine as if we get any other process which is not neovim, will be filtered out here).

Rereading the code I wrote in the earlier post I messed it up before posting it and have 2 loops for no reason.
This is what we need

```python
In [19]: for p in psutil.process_iter(attrs=["name"]):
    ...:     if "nvim" in p.info["name"] and p.environ().get("NVIM_LISTEN_ADDRESS_VIMTEX"):
    ...:         print(p)
    ...:
psutil.Process(pid=12420, name='nvim.exe', status='running', started='23:53:44')
```

``` python
In [21]: [p.environ()["NVIM_LISTEN_ADDRESS_VIMTEX"] for p in psutil.process_iter(attrs=["name"]) if "
    ...: nvim" in p.info["name"] and p.environ().get("NVIM_LISTEN_ADDRESS_VIMTEX")]
Out[21]: ['\\\\.\\pipe\\nvim-12420-0']
```

So finally in vim

```vim
144     if vimtex#util#is_win()
145       py3 << EOF
146 import psutil
147
148 sockets = [
149     p.environ()["NVIM_LISTEN_ADDRESS_VIMTEX"]
150     for p in psutil.process_iter(attrs=["name"])
151     if "nvim" in p.info["name"]
152     and p.environ().get("NVIM_LISTEN_ADDRESS_VIMTEX")
153 ]
154
155 EOF
```

With this it works now.

Alternatively, we can check only for "NVIM_LISTEN_ADDRESS" and then compare the filename only as we already do.
So strictly speaking we do not need "NVIM_LISTEN_ADDRESS_VIMTEX" in this scenario.

Btw, NVIM_LISTEN_ADDRESS has a note (Deprecated). Does this mean they will remove it completely in the future?
Maybe using `v:servername` to assign `NVIM_LISTEN_ADDRESS_VIMTEX` is safer?

> $NVIM_LISTEN_ADDRESS    Deprecated in favor of --listen.  If both are given, $NVIM_LISTEN_ADDRESS is ignored.

## Regarding filename

It works now. 

> if that works, then we can temporarily disable shellslash within this function.

Since function is supposed to be called from the terminal we can just set whichever option we like without having to worry about setting it back. But the if (in the commit) works well and is more general approach which is nice.

## Regarding the terminal popping up

`cmd /c start /min "" nvim --headless -c "VimtexInverseSearch %l '%f'"`

With this command the cmd still shows up but just for a moment before it hides again.
It as very short flash.

From here: https://stackoverflow.com/a/67300159
Although the comment says "`I've never seen a window flash when calling it via cmd /c start /min "".` I am afraid it still does in my case. But it is the shortest one I managed to find and make it worrk.

lervag:
Thanks; I'm at work so I'll only comment that I of course intended to use the environment stuff, I just removed it while testing on Linux and forgot to restore. The fix is pushed now. I'll comment further later tonight (I hope).

BertrandSim:
Regarding windows filenames, 

Unfortunately it does not work when 'shellslash' is set. 
Based on what I've found, for windows, the `\` separator should be used, *independent* of the 'shellslash' setting. 

For that reason, 
I propose to modify the following lines
```
function! vimtex#paths#s(path) abort " {{{1
  " Handle shellescape issues and simplify path
  " let l:path = exists('+shellslash') && !&shellslash
  "      \ ? tr(a:path, '/', '\')
  "      \ : a:path

  " Handle windows file separator issues and simplify path
  let l:path = vimtex#util#is_win()
        \ ? tr(a:path, '/', '\')
        \ : a:path

  return simplify(l:path)
endfunction
```

BertrandSim:
My previous comment about 'shellslash' was in relation to the use of `globpath()`. However, since `globpath()` is no longer used here, I think one can drop the 'shellslash' consideration altogether

lervag:
> My previous comment about 'shellslash' was in relation to the use of `globpath()`. However, since `globpath()` is no longer used here, I think one can drop the 'shellslash' consideration altogether

I'm sorry, I'm a little bit confused. Do I understand correctly that things work without doing the change you suggested, so that I can disregard it?

lervag:
> > Good idea. The alternative p.info['name'] in ('nvim', 'nvim.exe') looks better, though.
> 
> The disadvantage of this is that it does not include 'nvim-qt.exe', which we can add. I checked Neovide and it's process name is 'nvim.exe'. However, they are plenty more GUIs out there and we either have to check them and keep an up to date list or wait for someone to complain that it is not working with X GUI.
> 
> On the other hand 'nvim' in 'name' is too broad, however if we are going to check afterwards if it has $NVIM_LISTEN_ADDRESS or $NVIM_LISTEN_ADDRESS_VIMTEX or something else then I think we are good.

A compromise, then: let's do `nvim in name` when we also look for the env var, and then use the more explicit variant in the other case.

> > psutil is a standard lib, so I think in reality the dependency here is python3 and pynvim. And those seem quite common and I don't expect it to be considered a problem.
> 
> On my windows installation for both python 3.9 and 3.10, and fresh Conda installations 3.9 and 3.10 is not installed by default and must be installed.

Sorry, my mistake. You are right, `psutil` is not part of the standard libraries.

> I do not consider them a problem either, however it should be mentioned on the README/doc for anyone interested. For example, I do not use any plugin with pynvim dependency so I would be wondering what is going on.

Agreed; however, I've been thinking about it, and I don't think we need `pynvim` here. We only need the python provider itself and `psutil`.

> > Yes, this reference seems to indicate that we can't properly resolve this right now. So, perhaps we could fall back to something inspired by @jdhao here. That is, instead of doing the python stuff, we can write the servernames to a file ...
> 
> That would remove the psutil and pynvim dependency which is nice but it would add reading and writing files which is I personally prefer even less.

I'm curious: why?

> In any case, if you decide on this approach, neovim has `stdpath('cache')` (see `:help stdpath`) which can be used as a default.

Yes; I believe we can use the `g:vimtex_cache_root` location for this.

> With this it works now.

The current version should include these patches and so I'll mark it as verified for neovim on Windows. Thanks!

> Alternatively, we can check only for "NVIM_LISTEN_ADDRESS" and then compare the filename only as we already do. So strictly speaking we do not need "NVIM_LISTEN_ADDRESS_VIMTEX" in this scenario.
> 
> Btw, NVIM_LISTEN_ADDRESS has a note (Deprecated). Does this mean they will remove it completely in the future? Maybe using `v:servername` to assign `NVIM_LISTEN_ADDRESS_VIMTEX` is safer?

I think setting the environment variable is good. The deprecation notice at the very least indicates that we should not rely on `NVIM_LISTEN_ADDRESS`.

Notice, I tried to make the environment variable approach work on Linux, but here the `p.environ()` did _not_ include these variables. It's strange, and I suspect it may be that the environment presented by `p.environ()` is not updated after the process itself has updated it.

> ## Regarding the terminal popping up
> 
> `cmd /c start /min "" nvim --headless -c "VimtexInverseSearch %l '%f'"`
> 
> With this command the cmd still shows up but just for a moment before it hides again. It as very short flash.
> 
> From here: https://stackoverflow.com/a/67300159 Although the comment says "`I've never seen a window flash when calling it via cmd /c start /min "".` I am afraid it still does in my case. But it is the shortest one I managed to find and make it worrk.

Ok, I'll update the docs accordingly. I'm curious, did you try the `start /b` as previously suggested?

PanagiotisS:
All look OK now for Neovim on Windows.

> A compromise, then: let's do nvim in name when we also look for the env var, and then use the more explicit variant in the other case.

I am not sure what you mean for the other variant but the code look fine now!

> Agreed; however, I've been thinking about it, and I don't think we need pynvim here. We only need the python provider itself and psutil.

I was thinking the same. And that is why I made a new conda environment to test it.
I tried again. Here is the terminal output

```terminal
(base) PS C:\Users\ps> conda create --name test_pynvim
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: C:\tools\miniconda3\envs\test_pynvim


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate test_pynvim
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(base) PS C:\Users\ps> conda activate test_pynvim
(test_pynvim) PS C:\Users\ps> conda install python==3.9
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: done



  added / updated specs:
    - python==3.9


The following NEW packages will be INSTALLED:

  ca-certificates    pkgs/main/win-64::ca-certificates-2021.10.26-haa95532_2
  certifi            pkgs/main/win-64::certifi-2021.10.8-py39haa95532_0
  openssl            pkgs/main/win-64::openssl-1.1.1l-h2bbff1b_0
  pip                pkgs/main/win-64::pip-21.2.4-py39haa95532_0
  python             pkgs/main/win-64::python-3.9.0-h6244533_2
  setuptools         pkgs/main/win-64::setuptools-58.0.4-py39haa95532_0
  sqlite             pkgs/main/win-64::sqlite-3.36.0-h2bbff1b_0
  tzdata             pkgs/main/noarch::tzdata-2021e-hda174b7_0
  vc                 pkgs/main/win-64::vc-14.2-h21ff451_1
  vs2015_runtime     pkgs/main/win-64::vs2015_runtime-14.27.29016-h5e58377_2
  wheel              pkgs/main/noarch::wheel-0.37.0-pyhd3eb1b0_1
  wincertstore       pkgs/main/win-64::wincertstore-0.2-py39haa95532_2


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
(test_pynvim) PS C:\Users\ps> conda install psutil
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: C:\tools\miniconda3\envs\test_pynvim

  added / updated specs:
    - psutil


The following NEW packages will be INSTALLED:

  psutil             pkgs/main/win-64::psutil-5.8.0-py39h2bbff1b_1


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
(test_pynvim) PS C:\Users\ps> nvim --headless -c "VimtexInverseSearch 8 'C:\Users\ps\Documents\MyDocuments\test.tex'"
(test_pynvim) PS C:\Users\ps> conda install -c conda-forge pynvim
Solving environment: done

## Package Plan ##

  environment location: C:\tools\miniconda3\envs\test_pynvim

  added / updated specs:
    - pynvim


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    certifi-2021.10.8          |   py39hcbf5309_1         145 KB  conda-forge
    greenlet-1.1.2             |   py39h415ef7b_1          92 KB  conda-forge
    msgpack-python-1.0.2       |   py39h2e07f2f_2          81 KB  conda-forge
    ------------------------------------------------------------
                                           Total:         319 KB

The following NEW packages will be INSTALLED:

  greenlet           conda-forge/win-64::greenlet-1.1.2-py39h415ef7b_1
  msgpack-python     conda-forge/win-64::msgpack-python-1.0.2-py39h2e07f2f_2
  pynvim             conda-forge/win-64::pynvim-0.4.3-py39hcbf5309_0
  python_abi         conda-forge/win-64::python_abi-3.9-2_cp39
  pyuv               conda-forge/win-64::pyuv-1.4.0-py39hb82d6ee_1003

The following packages will be UPDATED:

  certifi            pkgs/main::certifi-2021.10.8-py39haa9~ --> conda-forge::certifi-2021.10.8-py39hcbf5309_1

The following packages will be SUPERSEDED by a higher-priority channel:

  openssl              pkgs/main::openssl-1.1.1l-h2bbff1b_0 --> conda-forge::openssl-1.1.1l-h8ffe710_0


Proceed ([y]/n)? y


Downloading and Extracting Packages
certifi-2021.10.8    | 145 KB    | ##################################################################################################################################################################### | 100%
msgpack-python-1.0.2 | 81 KB     | ##################################################################################################################################################################### | 100%
greenlet-1.1.2       | 92 KB     | ##################################################################################################################################################################### | 100%
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
(test_pynvim) PS C:\Users\ps> nvim --headless -c "VimtexInverseSearch 8 'C:\Users\ps\Documents\MyDocuments\cv\test.tex'"
['\\.\pipe\nvim-14000-0']
(test_pynvim) PS C:\Users\ps> conda remove pynvim
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: C:\tools\miniconda3\envs\test_pynvim

  removed specs:
    - pynvim


The following packages will be REMOVED:

  greenlet-1.1.2-py39h415ef7b_1
  msgpack-python-1.0.2-py39h2e07f2f_2
  pynvim-0.4.3-py39hcbf5309_0
  python_abi-3.9-2_cp39
  pyuv-1.4.0-py39hb82d6ee_1003

The following packages will be UPDATED:

  ca-certificates    conda-forge::ca-certificates-2021.10.~ --> pkgs/main::ca-certificates-2021.10.26-haa95532_2

The following packages will be SUPERSEDED by a higher-priority channel:

  certifi            conda-forge::certifi-2021.10.8-py39hc~ --> pkgs/main::certifi-2021.10.8-py39haa95532_0
  openssl            conda-forge::openssl-1.1.1l-h8ffe710_0 --> pkgs/main::openssl-1.1.1l-h2bbff1b_0


Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
(test_pynvim) PS C:\Users\ps> nvim --headless -c "VimtexInverseSearch 8 'C:\Users\ps\Documents\MyDocuments\test.tex'"
(test_pynvim) PS C:\Users\ps>
```

You can see the output of `165     unsilent echo l:socket_ids` when is running ok (btw was it left intentionally there?)

>> That would remove the psutil and pynvim dependency which is nice but it would add reading and writing files which is I personally prefer even less.

> I'm curious: why?

Not a strong preference honestly. Mostly comes from my frustration that ALL the applications dumping a file in my $HOME directory which then becomes a huge mess. There is a chaos there of files. _(Which btw was a "huge" advantage of neovim as vim and the plugins I was using were also part of the problem_).  But jokes aside, if you can configure where the file will be then is fine.

> Notice, I tried to make the environment variable approach work on Linux, but here the p.environ() did not include these variables. It's strange, and I suspect it may be that the environment presented by p.environ() is not updated after the process itself has updated it.

That is interesting! I suppose here is why https://unix.stackexchange.com/questions/29128/how-to-read-environment-variables-of-a-process

> Ok, I'll update the docs accordingly. I'm curious, did you try the start /b as previously suggested?

I did try a lot of things and options for both cmd and nvim. I did not remember which exactly now.
Testing again the following:

1. `cmd /c start /min "" nvim --headless -c "VimtexInverseSearch %l '%f'"` : Instant flash
2. `nvim --headless -c "VimtexInverseSearch %l '%f'"`: "long" popup cmd window
3. `start /b "" nvim --headless -c "VimtexInverseSearch %l '%f'"`: does not work. When running from powershell terminal it is asking for "Windows PowerShell credential request" and is asking for username and password. When running from cmd terminal it works without popup (as expected).
4. `cmd /c start /b nvim --headless -c "VimtexInverseSearch %l '%f''`: Similar to 2. when run from SumatraPDF, however when run from the terminal (windows terminal with powershell) is instantaneous. Weird difference.
5. `PowerShell.exe -windowstyle hidden {nvim --headless -c "VimtexInverseSearch %l '%f''"}`: Instant flash. Similar speed to 1. when run from SumatraPDF. No popup when run from the powershell terminal. However, as the default size of powershell is bigger the flashing window is bigger. And blue instead of black. (Not that it matters much).

Adding `cmd /q /q` makes no difference.

So what I think happens in 3. is that we need to open cmd, in order to run start, which runs nvim. So the short flash is from cmd.

https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start
https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd

BertrandSim:
> > My previous [comment](https://github.com/lervag/vimtex/pull/2219#issuecomment-958699644) about 'shellslash' was in relation to the use of `globpath()`. However, since `globpath()` is no longer used here, I think one can drop the 'shellslash' consideration altogether
> 
> I'm sorry, I'm a little bit confused. 

So sorry for the confusion. I think part of the confusion stems from which "previous comment" I was referring to. (I have now linked to the correct "comment" in the quote above)  

Summarizing that "previous comment" (and rephrasing),
* with globpath() and 'shellslash', does not work
* with globpath() and 'noshellslash', works
* with `\`        and 'shellslash', works
* with `\`        and 'noshellslash', works


> Do I understand correctly that things work without doing the change you suggested, so that I can disregard it?

Not quite, without the suggested change, the inverse search only works when 'shellslash' is not set; it should also work when 'shellslash' is set. 

> > I think one can drop the 'shellslash' consideration altogether

By this, I mean that one should always replace `/` with `\` whether 'shellslash' is set or not (ie, always use `\` for windows paths; there is no need to consider the 'shellslash' option), hence my suggestion [here](https://github.com/lervag/vimtex/pull/2219#issuecomment-960613312)

I hope that clarifies, please write back to me otherwise :)

PanagiotisS:
@BertrandSim 

I was actually curious so I tried it vim on Windows.
It works with both `set shellslash` and `set noshellslash` as it is expected as this exists in the code

https://github.com/lervag/vimtex/pull/2219/files#diff-75ba36348a62468eaed3de511d31c8407867b60597d17d2de485c8de587ed10dR34-R36

so `shellslash` should be irrelevant. Have you tried pulling the latest changes?
The code does not use `globpath`.

Here is a minimal vimrc

```vim
if &compatible
  set nocompatible
endif

set runtimepath+=C:\Users\ps\.cache\dein\repos\github.com\Shougo\dein.vim

call dein#begin('C:\Users\ps\.cache\dein')
call dein#add('C:\Users\ps\.cache\dein\repos\github.com\Shougo\dein.vim')
call dein#add('lervag/vimtex', {'rev': 'feat/better-inverse-search'})
call dein#end()

filetype plugin indent on
syntax enable

set number
set shellslash
```

@lervag 
Regarding cmd window with vim it actually has different behaviour.
Any combination of cmd and start (f.e. `cmd /q /c start /b`) leaves the cmd window open with the following message
```
♀
C:\Users\ps\AppData\Local\SumatraPDF>
```
No clue where this symbol comes from.

Using `cmd /c vim -T dumb --not-a-term -n -c "VimtexInverseSearch %l '%f'"` works fine although the cmd window stays open for <1sec. (longer than `cmd /c start /min` in neovim case which is almost instant).

Finally, now it works for me with both vim and neovim on Windows.

BertrandSim:
@PanagiotisS 

> so shellslash should be irrelevant. 

Agreed, That was exactly my point.

> The code does not use globpath

Again, that was exactly what I was getting at. The code does not use globpath, so there is no need to use shellslash in the code.

> It works with both set shellslash and set noshellslash as it is expected as this exists in the code

No, I am afraid it does not work with shellslash set.
In the code
```
function! vimtex#paths#s(path) abort " {{{1
  " Handle shellescape issues and simplify path
  let l:path = exists('+shellslash') && !&shellslash
        \ ? tr(a:path, '/', '\')
        \ : a:path

  return simplify(l:path)
endfunction
```

It only makes the correction (`tr(...)`) if `!&shellslash`, ie. when shellslash is not set. `tr(...)` is not applied if `&shellslash`.

user202729:
Regarding `start`: it's a Windows cmd's built-in, that's the likely reason why it
does not work in Sumatra PDF spawn.

As I mentioned above in the general case VBScript may be needed.

* https://stackoverflow.com/questions/23557720/execute-batch-file-without-command-line-visible
* https://stackoverflow.com/questions/4277963/how-to-call-cmd-without-opening-a-window
* https://www.faqforge.com/windows/how-to-execute-powershell-scripts-without-pop-up-window/
* https://www.raymond.cc/blog/hidden-start-runs-batch-files-silently-without-flickering-console/
* https://superuser.com/questions/62525/run-a-batch-file-in-a-completely-hidden-way

But for gvim in particular, it may be possible without VBScript:

I tried running some commands. The "nothing pops up" cases are highlighted.


	gvim --remote a

(assuming there's a GVIM remote server) A terminal window pop up very quickly then exits

	"C:\Program Files (x86)\Vim\vim82\gvim" --remote a

**Nothing pops up**

	"C:\Program Files (x86)\Vim\vim82\vim" --remote a

(assuming there's a VIM remote server) A terminal pops up quickly then exits

	"C:\Program Files (x86)\Vim\vim82\gvim" -c ":call mkdir('D:\a\' .. rand()) | quit"

Gvim window pop up quickly then exits

	gvim --cmd ":call mkdir('D:\a\' .. rand()) | quit"

A terminal window pop up very quickly then exits

	"C:\Program Files (x86)\Vim\vim82\gvim" --cmd ":call mkdir('D:\a\' .. rand()) | quit"

**Nothing pops up**

	"C:\Program Files (x86)\Vim\vim82\vim" --cmd ":call mkdir('D:\a\' .. rand()) | quit"

A terminal window pop up (longer than gvim's terminal window) then
exits

The cause of some of the popups is that `gvim` is shorthand for `gvim.bat`
file the installer created to be able to use it from command-line.

lervag:
@PanagiootisS

> All look OK now for Neovim on Windows.

Glad to hear it!

> > Agreed; however, I've been thinking about it, and I don't think we need pynvim here. We only need the python provider itself and psutil.
> 
> I was thinking the same. And that is why I made a new conda environment to test it. I tried again. Here is the terminal output

Thanks; I understand this as evidence that `pynvim` _is_ required.

> You can see the output of `165 unsilent echo l:socket_ids` when is running ok (btw was it left intentionally there?)

No, sorry - I'll remove. Thanks for noticing!

> > > That would remove the psutil and pynvim dependency which is nice but it would add reading and writing files which is I personally prefer even less.
> 
> > I'm curious: why?
> 
> Not a strong preference honestly. Mostly comes from my frustration that ALL the applications dumping a file in my $HOME directory which then becomes a huge mess.

Well, I _strongly_ relate to this frustration and I'll of course avoid messing with your home.

I plan to look into this idea in a follow-up PR after I merge this one.

> That is interesting! I suppose here is why https://unix.stackexchange.com/questions/29128/how-to-read-environment-variables-of-a-process

Thanks, this seems to confirm my suspicion.

> I did try a lot of things and options for both cmd and nvim. I did not remember which exactly now. Testing again the following: ...

I'm left with the impression that the current version of the docs are satisfactory, then. Is that right?

> EDIT: Also, this command assumes `nvim` is in PATH obviously. Otherwise, the full path must be used. For example `cmd /c start /min "" "C:\tools\neovim\Neovim\bin\nvim.exe" --headless -c "VimtexInverseSearch %l '%f'"`

Yes, good point. But I would think it is not necessary to document this, it seems an edge case?


@PanagiootisS and @BertrandSim

> so shellslash should be irrelevant. Have you tried pulling the latest changes?

> No, I am afraid it does not work with shellslash set.

Again I'm slightly confused, but my understanding is that I _should_ apply the patch suggested by @BertrandSim [here](https://github.com/lervag/vimtex/pull/2219#issuecomment-960613312). Do you both agree?

> Regarding cmd window with vim it actually has different behaviour. ...
> 
> Using `cmd /c vim -T dumb --not-a-term -n -c "VimtexInverseSearch %l '%f'"` works fine although the cmd window stays open for <1sec.

@BertrandSim Is this the same on your end?

> Finally, now it works for me with both vim and neovim on Windows.

Great, glad to hear it! This is quite close to ready, then!


@user202729

> So in conclusion
>
>     "C:\Program Files (x86)\Vim\vim82\gvim" --cmd "VimtexInverseSearch %l '%f'
>
> is likely to work. (drop x86 on 32-bit Windows)

Thanks; it seems the above `cmd /c` versions should also work well - would you still recommend to update/change/append something about this to the docs?

user202729:
Up to you. The disadvantage is that you need to know the "real" path to gvim to do it.

Alternatively add it (only) if many people ask for it.

PanagiotisS:
> I'm left with the impression that the current version of the docs are satisfactory, then. Is that right?

Github had the doc file folded so I missed. I though you were planning this last thing or something.

> 5257         cmd /c start /min "" nvim --headless -c "VimtexInverseSearch %l '%f'"

Based on my experience a vim example is needed. I suppose you are waiting for more feedback before updating this, but just in case I wanted to point it.

 > 297 protocol in the same fashion as the clientserver. Both Vim and neovim **has**
 > 298  the |v:servername| variable that contains the name/location of the server
 > 299 with which we need to communicate.

Shouldn't this be "Both Vim and Neovim have" ?

> \*vimtex-faq-neovim\*

We probably still need this for the "pynvim" part. And probably a reference from the `*vimtex-requirements*` session in a similar manner to windows reference would be really nice "and Neovim users should also read |vimtex-faq-neovim|"

>>    EDIT: Also, this command assumes nvim is in PATH obviously. Otherwise, the full path must be used. For example cmd /c start /min "" "C:\tools\neovim\Neovim\bin\nvim.exe" --headless -c "VimtexInverseSearch %l '%f'"

> Yes, good point. But I would think it is not necessary to document this, it seems an edge case?

Nowdays I am not sure as I do use mostly Linux. It really depends on how one installs (n)vim. In the past, I certainly had to edit PATH a lot of times and I still do every now and then. For example, I had to manually add mingw bin to the path a couple of days ago. And lets not start with Python and Windows and the huge mess that existed already, which Windows 10 made worse with their intrusive tactics. And (I am sorry for giving once again a personal example) considering that most people I know that they use latex (engineering, in academia) probably do not know what exactly PATH is or how it works (nor they care). So, having an explicit one-line example (or try"this" if it doesn't work) would be very helpful for them, if they ever came to use this plugin (some of them are using vim but without plugins). But then again I do not know. You can maybe leave it off and if someone asks you can add it later. 

### Regarding path and shellslash

@BertrandSim I am sorry. You are right. With shellslash it does not work. I messed it up apparently. (I did not take into account that half of the code is executed in a different (n)vim instance than the one we start with.)

@lervag 

> Again I'm slightly confused, but my understanding is that I should apply the patch suggested by @BertrandSim here. Do you both agree?

Yes. So it seems that regardless of `shellslash` setting if we do 'tr(a:path, '/', '\')` we are good.

To see what is going on I will put a couple of echoms in the code and posting them here so you can see what's going on,

in `autoload/vimtex/view.vim`

```vim
function! vimtex#view#inverse_search(line, filename) abort " {{{1
  " Only activate in VimTeX buffers
  if !exists('b:vimtex') | return -1 | endif

  echom '-----------'
  echom 'Shellslash = ' . &shellslash

  " Only activate in relevant VimTeX projects
  let l:file = resolve(a:filename)
  let l:sources = copy(b:vimtex.sources)
  echom '1: l:file = ' . l:file
  if vimtex#paths#is_abs(l:file)
    call map(l:sources, {_, x -> vimtex#paths#join(b:vimtex.root, x)})
  endif
  echom '4: l:sources = ' . string(l:sources)
  if index(l:sources, l:file) < 0 | return -2 | endif
```

in `autoload/vimtex/paths.vim`
```vim
function! vimtex#paths#s(path) abort " {{{1
  " Handle shellescape issues and simplify path
  echom '2: a:path = ' . a:path
  " let l:path = exists('+shellslash') && &shellslash
        " \ ? tr(a:path, '/', '\')
        " \ : a:path
  let l:path = tr(a:path, '/', '\')
  echom '3: l:path = ' . l:path

  return simplify(l:path)
endfunction
```

And the result is (for both vim and neovim),

```
-----------
Shellslash = 1
1: l:file = C:\Users\ps\Documents\MyDocuments\test.tex
2: a:path = C:/Users/ps/Documents/MyDocuments/test.tex
3: l:path = C:\Users\ps\Documents\MyDocuments\test.tex
4: l:sources = ['C:\Users\ps\Documents\MyDocuments\test.tex']

-----------
Shellslash = 0
1: l:file = C:\Users\ps\Documents\MyDocuments\test.tex
2: a:path = C:\Users\ps\Documents\MyDocuments/test.tex
3: l:path = C:\Users\ps\Documents\MyDocuments\test.tex
4: l:sources = ['C:\Users\ps\Documents\MyDocuments\test.tex']  
```

lervag:
> > I'm left with the impression that the current version of the docs are satisfactory, then. Is that right?
> 
> Github had the doc file folded so I missed. ...

Thanks for the comments; I've updated the docs now.

> > *vimtex-faq-neovim*
> 
> We probably still need this for the "pynvim" part. And probably a reference from the `*vimtex-requirements*` session in a similar manner to windows reference would be really nice "and Neovim users should also read |vimtex-faq-neovim|"

Ok, thanks; I'll add/update that.

> Nowdays I am not sure as I do use mostly Linux. It really depends on how one installs (n)vim. In the past, I certainly had to edit PATH a lot of times and I still do every now and then. For example, I had to manually add mingw bin to the path a couple of days ago. And lets not start with Python and Windows and the huge mess that existed already, which Windows 10 made worse with their intrusive tactics. And (I am sorry for giving once again a personal example) considering that most people I know that they use latex (engineering, in academia) probably do not know what exactly PATH is or how it works (nor they care). So, having an explicit one-line example (or try"this" if it doesn't work) would be very helpful for them, if they ever came to use this plugin (some of them are using vim but without plugins). But then again I do not know. You can maybe leave it off and if someone asks you can add it later.

I'll consider to add something here.

> ### Regarding path and shellslash
> 
> @BertrandSim I am sorry. You are right. With shellslash it does not work. I messed it up apparently. (I did not take into account that half of the code is executed in a different (n)vim instance than the one we start with.)

Great, I'll fix and update. Thanks for looking into it so thoroughly!

----

I'll now take the liberty of merging this PR. Please don't hesitate to open issues with ideas for improving the docs or this new feature.

And again, thanks to everyone who took their time to pitch in! I very much appreciate it!

lervag:
Again, thank you all. This was more work than anticipated, but I think and hope it will make the experience better for inverse search with VimTeX.

PR #2226, opened 2021-10-29
More conservative filetype detection for .cls file

Hans-Guenter:
The filetype detection for tex files with extension cls is more restrictive in the vim repository https://github.com/vim/vim/blob/master/runtime/filetype.vim.
This change mimics this logic.

### DIFF
--- a/ftdetect/cls.vim
+++ b/ftdetect/cls.vim
@@ -6,4 +6,7 @@
 
 if !get(g:, 'vimtex_enabled', 1) | finish | endif
 
-autocmd BufRead,BufNewFile *.cls set filetype=tex
+autocmd BufNewFile,BufRead *.cls
+	\ if getline(1) =~ '^%' |
+	\  set filetype=tex |
+	\ endif


lervag:
> Conflicted. I stand by my request for the current behavior as being more convenient, but I see how diverging from upstream could add undesirable friction. In particular, if VimTeX has _any_ customized filetype detection, users need to override it not in `init.vim` but in an `after/plugin/*.vim` (👈🏻 @Hans-Guenter). So the only real alternative to the current behavior is to remove the `*.cls` autocommand altogether.

Ok, thanks for the input!

> (In a twist of irony, I've switched to https://github.com/nathom/filetype.nvim, which disables _all_ `filetype.vim` detection -- including VimTeX's. So I'm not even benefitting from the current behavior anymore...)

Interesting! I assume you don't set your filetype manually, though?

> I myself want to keep this a low profile issue. I equally can go with after/plugin approach. Thank you for your considerations.

Ok, thanks. I don't see a very good reason for changing this. I don't mind discussion and considering it, but I would prefer to have a more complete example that shows how the current implementation breaks expectations.

So, as such, I'll close the issue for now, as I have the understanding that you've resolved this with @clason's suggestion.

clason:
> > (In a twist of irony, I've switched to https://github.com/nathom/filetype.nvim, which disables _all_ `filetype.vim` detection -- including VimTeX's. So I'm not even benefitting from the current behavior anymore...)
> 
> Interesting! I assume you don't set your filetype manually, though?

I haven't, because I just recently realized that 😅 

I'll probably just add an override to `filetype.nvim` (or whatever its inevitable builtin version will be) for it.

PR #2238, opened 2021-11-09
Added lhd (◁) and rhd (▷) to conceal symbols

yehoshuapw:
Added conceal symbols for `\lhd` (◁) and `\rhd` (▷).

thanks for all the great work!


lervag:
Thanks, both for the PR and the kind words :)

PR #2242, opened 2021-11-10
fix: use unicode character for `\times` conceal

savq:
Concealing `\times` with a letter `x` can make expressions like
`x \times x` awkward to read.


lervag:
Thanks!

PR #2244, opened 2021-11-11
biblatex syntax:  support for concealment of additional biblatex commands

krissen:
Adds support for concealment of additional biblatex commands.

(Additions to the test biblatex.tex file weren't require as the citation commands were already there. Previous to the changes suggested in this PR, several commands were not concealed. With it, all are.)

### DIFF
--- a/autoload/vimtex/syntax/p/biblatex.vim
+++ b/autoload/vimtex/syntax/p/biblatex.vim
@@ -39,21 +39,20 @@ function! vimtex#syntax#p#biblatex#load(cfg) abort " {{{1
 
   if g:vimtex_syntax_conceal.cites
     let s:re_concealed_cites = '\v\\%(' . join([
-			    \ '%([Tt]ext|[Ss]mart|[Aa]uto)cite[s]?',
-			    \ '(foot)?cite[tp]?',
-			    \ '[Aa]utocite[s]?',
-			    \ 'bibentry',
-			    \ '[Cc]ite%(title|author|year%(par)?|date)[s]?',
-			    \ '[Pp]arencite[s]?',
-			    \ '[Ppf]?[Nn]otecite',
-			    \ '[pPfFsStTaA]?[Vv]olcite[s]?',
-			    \ '[Ss]upercite[s]?',
-			    \ 'cite%(num|text|url|field|list|name)',
-			    \ 'citeal[tp]',
-			    \ 'foot%(full)?cite%(text)?',
-			    \ 'footcite%(s|texts)',
-			    \ 'fullcite[s]?',
-			    \ ], '|') . ')>\*?'
+          \ '%([Tt]ext|[Ss]mart|[Aa]uto)cite[s]?',
+          \ '(foot)?cite[tp]?',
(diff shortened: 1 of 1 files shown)


lervag:
But do we _want_ all of them concealed? Including `\bibentry`?

@whisperity and @Melkster I believe you were part of the original discussion on this topic. What do you think?

autoload/vimtex/syntax/p/biblatex.vim
```diff
@@ -39,9 +39,21 @@ function! vimtex#syntax#p#biblatex#load(cfg) abort " {{{1
...
+			    \ '[Pp]arencite[s]?',
+			    \ '[Ppf]?[Nn]otecite',
+			    \ '[pPfFsStTaA]?[Vv]olcite[s]?',
+			    \ '[Ss]upercite[s]?',
+			    \ 'cite%(num|text|url|field|list|name)',
+			    \ 'citeal[tp]',
+			    \ 'foot%(full)?cite%(text)?',
+			    \ 'footcite%(s|texts)',
+			    \ 'fullcite[s]?',
+			    \ ], '|') . ')>\*?'
```
lervag, line 56: You've changed the indents to tabs. I prefer spaces, and in VimTeX I want to keep a consistent style.

krissen:
> But do we _want_ all of them concealed? Including `\bibentry`?
> 

Indeed, `bibentry` was the one I was unsure about. The rest I do believe are all used *in-text*, and thus ought to be concealed. I'd think that would be the logic. "Should cites be concealed? *Looks at variable for cite-concealment.* Yes. Alright, all citation commands used in-text will be concealed."

However, if one would like to exclude some citation commands --- instead of adding a separate variable to test (i.e. `vimtex_syntax_conceal_cites` or some such with a list of types of citation commands to conceal), how about simply referring to `vimtex_syntax_custom_cmds` for un-concealment of individual commands? (`conceal: 0`).

autoload/vimtex/syntax/p/biblatex.vim
krissen: Fixed.

krissen:
> Indeed, `bibentry` was the one I was unsure about.

Removed bibentry from cite-concealment.

lervag:
Thanks! I've merged. Please note: I prefer if you don't merge master into your branch _unless_ necessary. I like to keep a logically clean Git history, and it helps when I can easily rebase a branch on top of master before I merge.

I think you are right that this makes sense, and let's instead have the discussion if someone now should oppose this.

krissen:
> Please note: I prefer if you don't merge master into your branch _unless_ necessary. I like to keep a logically clean Git history, and it helps when I can easily rebase a branch on top of master before I merge.

Got it. New to contributing through PR:s. Merged master in between, thinking it was proper practice, but see why it would've been better not to. Will keep it in mind should there be any future PR:s! 👍

lervag:
I guess it is a matter of taste; I won't claim that my practise is correct or better, but it is my preferred workflow and I try to enforce it when I can :)

Issue #2046, opened 2021-05-06
Request: Wayland Support


lorenzleutgeb:
Currently, some vimtex Features, e.g. cursor sync with Zathura, are not compatible with Wayland. I'd like to request making it compatible. This involves finding out which features are *not* compatible, first.

The topic has come up in (at least) the two issues linked below. I create this request to allow tracking of Wayland support and to invite other Wayland users to give their "thumbs up" to show it's importance.

**Is your feature request related to a problem?**
Please refer to
#1775 
#1719 

**Describe the solution you'd like**
vimtex should work support interaction with Wayland, with minimal user-configuration/-interaction. 

**Describe alternatives you've considered**
None. Note that I don't think use "Use X11." is an alternative worth considering.

**Additional context**
None.

lervag:
> Currently, some vimtex Features, e.g. cursor sync with Zathura, are not compatible with Wayland. I'd like to request making it compatible. This involves finding out which features are _not_ compatible, first.

I don't mind having Wayland support. Or, more correctly: I would be very happy if I could claim VimTeX works perfectly well on Wayland.

As you say, the first step is to know exactly what is lacking on Wayland. As far as I know, the main hurdle is that the Zathura and MuPDF viewers rely on the `xdotool` command to avoid multiple windows (both) and for some concenient controls (MuPDF).

Another obvious hurdle: I don't use Wayland yet myself. The main, and big, implication: I won't actively work to solve this issue. But I will be more than interested in helping anyone who is interested and able to suggest a solution!

> The topic has come up in (at least) the two issues linked below. I create this request to allow tracking of Wayland support and to invite other Wayland users to give their "thumbs up" to show it's importance.

Since I don't use Wayland myself, I think the main impact the thumbs up's may give here is attention to the issue, and perhaps motivation for someone who can help me solve this.

lervag:
I'm closing this now; not because I'm not interested in progress, but because I won't act on it myself. I would be very happy to see PRs or concrete suggestions, but as long as I don't use Wayland I will require active users who can help both with implementing and testing.

thetic:
Can we reopen this? I think its still valuable even if it isn't a priority right now.

lervag:
It's not about priority - instead, it is due to me not having any plan of working on it. I need (competent) users to help with this.

I can reopen and tag this as "help needed", though.

mvasi90:
Hello everyone.

I'm using Wayland and I don't have the `xdotool`.
[...]

PR #2259, opened 2021-11-26
Make zathura plugin usable on Wayland

ftilde:
This PR changes the zathura viewer to not be a specialization of the xwin template and thus not require or use xdotool. The only functionality of the xwin template that was really used previously was the check for existing windows which used the xwin id. Now we use the pid instead for which there is a function already. In order for that function to robustly, all locations that interact with a zathura process and its arguments (i.e., calls to zathura and the pgrep call in get_pid) use normalize the pdf path to be relative to the current directory. This matches how the latex compiler seems to open the viewer. This could be made even more robust by normalizing all paths to be absolute paths, but I'm not sure how to do that for the zathura instance that is created when starting the compiler. As far as I can tell, the xwin template simply searches for a window with the given viewer name/title (i.e., "Zathura"), however, so the current change is already a step up in that regard.

This would solve #2046.

### DIFF
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -20,8 +20,7 @@ function! vimtex#view#zathura#new() abort " {{{1
     endif
   endif
 
-  " Use the xwin template
-  return vimtex#view#_template_xwin#apply(deepcopy(s:zathura))
+  return vimtex#view#_template#apply(deepcopy(s:zathura))
 endfunction
 
 " }}}1
@@ -33,6 +32,7 @@ let s:zathura = {
       \}
 
 function! s:zathura.start(outfile) dict abort " {{{1
+  let l:outfile = fnamemodify(a:outfile, ':.')
   let l:cmd  = 'zathura'
   if self.has_synctex
     let l:cmd .= ' -x "' . s:inverse_search_cmd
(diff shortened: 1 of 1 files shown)


autoload/vimtex/view/zathura.vim
```diff
@@ -33,6 +32,7 @@ let s:zathura = {
       \}
 
 function! s:zathura.start(outfile) dict abort " {{{1
+  let l:outfile = fnamemodify(a:outfile, ':.')
```
lervag, line 35: Why do you need this modification?

lervag:
> The only functionality of the xwin template that was really used previously was the check for existing windows which used the xwin id.

This is not strictly true, because there is also an option to e.g. focus the viewer with inverse search. This relies on xdotool. And a user may apply further customization that relies on xdotool things.

However, I do agree it would be better to simplify these things and avoid using xdotool unless absolutely necessary. So I think you are taking things in the right direction. I think the xwin template should be removed and instead the important functions there should be _functions_ instead of an extra _class_ (or something like that).

> Now we use the pid instead for which there is a function already. In order for that function to robustly, all locations that interact with a zathura process and its arguments (i.e., calls to zathura and the pgrep call in get_pid) use normalize the pdf path to be relative to the current directory. This matches how the latex compiler seems to open the viewer. This could be made even more robust by normalizing all paths to be absolute paths, but I'm not sure how to do that for the zathura instance that is created when starting the compiler.

Ah, ok; that answers my separate question. I don't know how to make latexmk use absolute paths when calling the viewer, so it seems you are right to instead adjust from VimTeX side. I would have added a comment to explain why we will not use the supplied absolute path, though.

> As far as I can tell, the xwin template simply searches for a window with the given viewer name/title (i.e., "Zathura"), however, so the current change is already a step up in that regard.

Yes and no; there is also the element of timing involved. If we search for a Zathura window just after it was opened, we will usually find the correct window.

----

Ok, so: I think you are on to something, but it needs more work to avoid regression for some users. In fact, I think it requires a refactoring which may be slightly complicated. But I'm motivated to do it when I get the time.

lervag:
I've now done the refactoring I mentioned. And so we are getting ready to incorporate your suggestions. However, there are some questions first. I would be happy if you could contribute - please see #2271.

ftilde:
On 07.12.2021 01:07, Karl Yngve Lervåg wrote:
>I've now done the refactoring I mentioned. And so we are getting ready to incorporate your suggestions. However, there are some questions first. I would be happy if you could contribute - please see #2271.

Alright! I've briefly looked over the refactoring. And so far it seems 
sensible to me. I can unfortunately only help to test this on Linux as 
well, but I'm happy to do so. Also, I can try to rebase this PR on top 
of #2271 some time in the next few days and see if any issues arise.

lervag:
> Alright! I've briefly looked over the refactoring. And so far it seems sensible to me. I can unfortunately only help to test this on Linux as well, but I'm happy to do so. Also, I can try to rebase this PR on top of #2271 some time in the next few days and see if any issues arise.

Great, thanks!

If you do rebase this PR on top, then please do some tests as well. Specifically: your fix relies on keeping the actual command line invocation of zathura the same across VimTeX and latexmk. At zeroth order, I think your fix works. But does it continue to work if the user specifies a different output directory with the `build_dir` option? If it doesn't, is there an easy way to make it work?

Also, note that one possible simplification would be to overload the `.out()` method for Zathura. You may need to copy some of the boiler plate code in that case, but feel free to do so. This may reveal a weak point that may need further refactoring from my side.

lervag:
Any updates here?

ftilde:
Unfortunately, no. :( Work has continued to be super busy. If someone 
else wants to take this over, feel free. If not, I definitely plan to 
come back and finish this up.

lervag:
No problem and no rush, take your time :)

ftilde:
So, as promised, I just came back to work on this. However, it seems like with the refactorings done in the meantime (especially the pid matching for the zathura instance) it all seems to work flawlessly for me on wayland (sway) with the current master version (including bidirectional position sync between vim and zathura, i.e. vimtex-view etc.). Excellent!

If I encounter any issues in the future, I'll try to fix them, but for now I'm closing this PR now.
Thank you for your work!

lervag:
Great, I'm glad to hear it works better for you now! :D

PR #2262, opened 2021-11-30
Fix zathura inverse search on NixOS

### LINKED ISSUE #2261: Allow configuring the nvim path used inside s:inverse_search_cmd
On NixOS, if one chooses to configure neovim declaratively, then the neovim binary gets wrapped inside a wrapper.
What the wrapper does is it calls the neovim binary with suitable arguments and environment variables. One of the arguments passed is `-u /nix/store/h97rm0ckx2p9vqcc5hsdkskykw882j87-init.vim` which is how the plugins such as vimtex get loaded.
This usually works fine, but when vimtex uses zathura for viewing pdf, and since [`get(v:, 'progpath')`](https://github.com/lervag/vimtex/blob/d9e05f5e8010ce43670d19655058a6dc8cad300e/autoload/vimtex/view/zathura.vim#L113) returns the path to the unwrapped binary, the unwrapped binary gets passed to zathura as part of the synctex command. This immediately causes a problem: since the unwrapped binary can't possibly load vimtex, the command will always fail.

One way to solve this is issue to allow configuring the nvim path used inside s:inverse_search_cmd so people using NixOS (and possibly GuixSD) can manually set the nvim path to the wrapper. Unfortunately, I'm don't know much vimscript so I cannot implement this.
See also: https://github.com/NixOS/nixpkgs/issues/78033

poscat0x04:
closes #2261

### DIFF
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -110,7 +110,8 @@ endfunction
 
 " }}}1
 
-let s:inverse_search_cmd = get(v:, 'progpath', get(v:, 'progname', ''))
+let s:inverse_search_cmd = get(g:, 'vimtex_callback_progpath',
+      \                      get(v:, 'progpath', get(v:, 'progname', '')))
       \ . (has('nvim')
       \   ? ' --headless'
       \   : ' -T dumb --not-a-term -n')
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -2927,6 +2927,15 @@ OPTIONS                                                        *vimtex-options*
 
   Default value: 1
 
+*g:vimtex_callback_progpath*
+  The path to the vim binary that will be passed to viewers (for example
+  Zathura) for synctex callbacks. If unset, vimtex will fallback to use
+  |v:progpath|. You usually don't have to touch this variable, but if your
+  vim loads vimtex using a wrapper then it needs to be set otherwise
+  inverse search calls from viewers won't work.
+
+  Default value: unset
+
 *$VIMTEX_OUTPUT_DIRECTORY*
   This environment variable allows to specify the output directory of
   auxiliary LaTeX files. If it exists and is a valid path, this path will be


lervag:
Thanks for raising an issue and also proposing a resolution!

autoload/vimtex/view/zathura.vim
```diff
@@ -110,7 +110,7 @@ endfunction
 
 " }}}1
 
-let s:inverse_search_cmd = get(v:, 'progpath', get(v:, 'progname', ''))
+let s:inverse_search_cmd = get(g:, 'vimtex_view_zathura_vim_path', get(v:, 'progpath', get(v:, 'progname', '')))
```
lervag, line 113: Although this is currently used only for Zathura, I think it could be wise to use a more general name. How about `g:vimtex_callback_progpath` or simply `g:vimtex_progpath`?

lervag: Also, I prefer to keep lines below 80 columns unless there's a good reason not to.

doc/vimtex.txt
```diff
@@ -2927,6 +2927,14 @@ OPTIONS                                                        *vimtex-options*
 
   Default value: 1
 
+g:vimtex_view_zathura_vim_path
+  The path to the vim binary that will be passed to Zathura. If unset, vimtex
```
lervag, line 2931: I would generalize this text somewhat and raise Zathura as an example.

lervag (changes requested):
I only have minor comments. Let me know what you think.

autoload/vimtex/view/zathura.vim
poscat0x04: fixed

doc/vimtex.txt
poscat0x04: done.

poscat0x04:
I've tested this PR on my machine and it works.

lervag:
Thanks!

PR #2267, opened 2021-12-04
doc: fix the example for latexmk config

poscat0x04:
This example is [outdated](https://github.com/lervag/vimtex/blob/178e58a9d65f4e8cd0e646f0b1def0923e24c6f5/autoload/vimtex/compiler.vim#L45-L47) (and misleading!)


lervag:
Thanks!

PR #2271, opened 2021-12-07
Refactor viewer backend

lervag:
The viewer backend has been somewhat messy for quite some time. It was time to clean up and refactor it.

Either as part of the PR or after the PR I can consider to also incorporate #2259 - specifically to avoid the `xdo_exists` method for Zathura and instead rely on checking the PID. This will allow Zathura to work more or less as expected on Wayland. Note that this requires a "hack" on the pdf file name to work, and this must be checked properly if it works as expected e.g. with different build directories.

I would be happy to have some testers test this stuff. @clason, can you test Skim no OSX?

I've personally tested all backends on Linux and it seems to (continue to) work as expected.


clason:
> I would be happy to have some testers test this stuff. @clason, can you test Skim no OSX?

Can do! Although there's a slight niggle in that Apple has decided to break scriptability in the macOS 12.1 betas, _hard_ -- which completely breaks the `displayline` bundled with Skim. Not sure whether and how it's possible to work around that. Just a heads-up that there'll likely be a bunch of reports once macOS 12.1 is released...

clason:
Doing `\lv` immediately gives me the error
```
Error detected while processing function vimtex#view#view[2]..49[17]..61:
line    1:
E121: Undefined variable: outfile
E116: Invalid arguments for function join
```
(PDF exists in the current directory.)

lervag:
Sorry, I think that's fixed now.

lervag:
> Although there's a slight niggle in that Apple has decided to break scriptability in the macOS 12.1 betas, _hard_ -- which completely breaks the `displayline` bundled with Skim. Not sure whether and how it's possible to work around that. Just a heads-up that there'll likely be a bunch of reports once macOS 12.1 is released...

Huh. Well, you should all just convert to Linux... :p

So, this will clearly raise the question of how we can address this challenge in the future. That should be a different thread, obviously.

clason:
Indeed, works now!

lervag:
Great, glad to hear it! :)

clason:
> Huh. Well, you should all just convert to Linux... :p

Nah, I'm not going back :) (Although issues like this do not make a strong case for switching to macOS -- sometimes I wish for a Linux-based alternative...)

> So, this will clearly raise the question of how we can address this challenge in the future. That should be a different thread, obviously.

I don't think VimTeX can do anything about this... Either 
1. this gets fixed (eventually), then nothing needs to be done, or 
2. Skim finds another way to be controlled from the command line (then I'll make a PR to change `skim.vim`), or
3. we just need to document that forward search/automatic launching is not supported on Skim (and to blame Apple).

clason:
(Skim is still broken on the 12.1 release candidate, unfortunately -- this means option 1 is pretty much out :( )

lervag:
I'm sad to hear this. I hope Skim devs will implement an alternative approach, then.

Perhaps another alternative can be to recommend a different viewer such as Zathura?

clason:
Probably not for macOS -- people like their native apps...

Maybe the `open -a` workaround can be added as a fallback? Does VimTeX see error messages from shell tools?

clason:
Ah, that won't work -- the callback to reload the PDF after compilation finished also requires Apple Script.

Yes, I don't see an alternative to recommending using a different viewer, as Apple completely broke Skim for use as a LaTeX previewer...

lervag:
Perhaps one could open an issue with the Skim project tracker and link it here to possibly have a joint discussion on this?

clason:
Oh, they are aware of it: https://sourceforge.net/p/skim-app/bugs/1438/

lervag:
That's good. If you (or any other VimTeX user who are reading this) joins that discussion, feel free to also link this issue back there.

lervag:
Ok, I'll just merge this. It seems to me to work well, and I'll rather fix things as they pop up after users start to use it.

clason:
Brief update: macOS 12.1 is out, with the breaking changes. The Skim dev has a test version that works, at least for the update callback, although it looks like the AppleScript snippets need adapting for the forward search. 

Expect some breakage (reports) for some days until the new version is released and I can adapt the forward view script.

PR #2282, opened 2021-12-15
fix: add label env to texMathText

ld0891:
Added `\label{}` environment to text-inside-math detection, along with a few tests.

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -527,6 +527,7 @@ function! vimtex#syntax#core#init() abort " {{{1
         \ 'text%(normal|rm|up|tt|sf|sc)?',
         \ 'intertext',
         \ '[mf]box',
+        \ 'label',
         \]
     execute 'syntax match texMathCmdText'
           \ '"\v\\' . l:re_cmd . '>"'
--- a/test/test-syntax/test-core.tex
+++ b/test/test-syntax/test-core.tex
@@ -115,6 +115,14 @@ Line with $inline math$ here.\footnote{here {asd} with \cmd also}.
   Lorem ipsum \dots
 \end{theorem}
 
+\begin{equation}
+    e^{2 \pi i} = 1
+    \text{foo}
+    \intertext{bar}
+    \mbox{foo}
+    \label{bar}
+\end{equation}
+
 \begin{thebibliography}{1}
   \bibitem{abc} a citation.
   \bibitem[label]{abc} a citation.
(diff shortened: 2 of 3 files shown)


lervag:
I'm not sure; `\label` is already matched with `texCmdRef`. The argument for `\label` is not normal text mode - it is simply a reference string for use in `\ref` commands. So I'm curious, what do you think is wrong with the current version?

ld0891:
Ah I see, thanks for the prompt reply. The current logic in the function `vimtex#syntax#in_mathzone(...)` is first detecting `texMathZone` as well as `texMathText`, then excluding those matching `texMathText`. Now I understand that `\label` doesn't count as normal text mode, but should `texCmdRef` as well as `texRefArg` be excluded in that function as well? If so, this change is more suitable to be made in the function itself rather than the syntax part.

lervag:
> Ah I see, thanks for the prompt reply. The current logic in the function `vimtex#syntax#in_mathzone(...)` is first detecting `texMathZone` as well as `texMathText`, then excluding those matching `texMathText`. Now I understand that `\label` doesn't count as normal text mode, but should `texCmdRef` as well as `texRefArg` be excluded in that function as well? If so, this change is more suitable to be made in the function itself rather than the syntax part.

Yes, perhaps; do you currently have any issues related to this particular thing? E.g. unexpected behaviour because the `\label` is matched as math environment?

ld0891:
Yeah it's about completion. Right now there are different types of completion going on, one is math operators like `\frac` within math mode, another is environment like `\begin{align}` within text mode, yet another one is the reference strings you mentioned for `\ref`, which is in text mode as well. Plugins like UltiSnips use `vimtex#syntax#in_mathzone(...)` to decide which type of suggestions to make or expand. Right now if it's a `\label` within math mode, it still suggests/expands math operators, which is not the intended behavior. Please let me know what you think.

lervag:
> Right now if it's a `\label` within math mode, it still suggests/expands math operators, which is not the intended behavior. Please let me know what you think.

Yes, I agree, except I don't see quite how this would be a real problem. I also use Ultisnips, but I never try to expand a snippet within `\label{...}`. I also have a couple of minor imaps (`:help vimtex-imaps`), but again, these are not relevant inside a label command and I don't have any issues with this.

I might still fix this, but it would be nice to see an actual real problem instead of just the theoretic fact that this is inconsistent/wrong. To expand on this: there are always going to be things that are not perfect, and I don't aim for that. Instead, I try to be pragmatic and address things that make a difference.

lervag:
I believe this is addressed now.

Issue #180, opened 2015-06-07
Change window focus behaviour on `\ll`: Focus the newly created viewer window, not vim


nasenatmer:
First, thanks for your great work on this project again :)

Although we talked about it recently and said it would be more intuitive if `vimtex` returns window focus to vim after the viewer has been started due to successful compilation by `\ll`, I think I've changed my view here.

Whenever I have a new window in front of any other window, I assume that this is the one in focus and start using key strokes (eg with `zathura` in order to set width, scroll down or something like that). However, these keys are sent to vim as the focus is returned to it when the viewer is opened for the firmt time, which often irritates me.

Although I'd appreciate if the default behaviour would change, I'd equally be happy to know which part of `vimtex#util#execute` needs changed in my personal fork so I can change this behaviour locally.

lahwaacz:
Wait, the vim window is focused but it still remains in "background"? This sounds more like a problem with the window manager...

nasenatmer:
Indeed. Workflow is as follows:

1. Open vim to edit tex file
2. press `\ll`, compilation starts
3. when compilation has ended successfully, `zathura` opens
4. `openbox` returns focus to vim, while `zathura` is still in the foreground

Did I mis-configure something with openbox? Here is my [rc.xml](http://paste2.org/GBg5bLdE), the only thing I remember to have configured that is related to this, is that I want mouse cursor to focus a window and to place new windows under mouse cursor.

lahwaacz:
I'm not sure, I haven't used Openbox for quite some time now... Perhaps it's disabled by the `raiseOnFocus` value? But allegedly the windows would be raised even as you move the mouse, which is probably not what you want.

Anyway I think that the focus should be changed only as new windows open, so ideally vimtex itself would never set focus. Unfortunately starting Zathura again to sync when the compilation finishes means that it will get focus, which is then set back to vim by the callback in vimtex. This also means that focus on a third window will not be preserved, so isn't there a way to sync the PDF viewer without starting the viewer again, which would also let the window manager manage the focus?

lervag:
I've not directly addressed this issue yet, but I've pushed a patch that should fix some minor itches. Could you please test the latest version and see if it helps somewhat?

nasenatmer:
Wow, so much replies here, but I first have to apologise, it seems I didn't make my problem clear enough.

The whole scenario I was talking about was only about the **first time a viewer is opened** by vimtex' `\ll` command.

I was not concerned with the behaviour of `\lv`.

[...]

Issue #2279, opened 2021-12-14
Skim forward search not working on macOS 12.1


Zeta611:
### Description

I recently got an M1 Mac (if relevant) and setup Neovim + vimtex almost identically to my previous intel Mac.
However, forward search does not work anymore.

I installed both Neovim and Skim via Homebrew.

### Steps to reproduce

I invoked `neovim` with `NVIM_LISTEN_ADDRESS=/tmp/nvimsocket nvim -u minimal.vim minimal.tex`.

```vim
set nocompatible

let &runtimepath  = '~/.local/share/nvim/plugged/vimtex,' . &runtimepath
let &runtimepath .= ',~/.local/share/nvim/plugged/vimtex/after'

filetype plugin indent on
syntax on

let g:vimtex_view_method = 'skim'
```

```tex
\documentclass{article}

\begin{document}
Hello, \TeX!
\end{document}
```

### Expected behavior

Forward search should work with `\ll` or `:VimtexView`.

### Actual behavior

Skim does not get focused nor 'forward-search'ed.

Initial `\ll` does wake Skim and open the document (without current line being focused).

Backward search does work properly.

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info:
  OS: macOS 12.1 (21C52)
  Vim version: NVIM v0.6.0
  Has clientserver: true
  Servername: /tmp/nvimsocket

VimTeX project: test
  base: test.tex
  root: /Users/jay/Developer/TeXProjects/test
  tex: /Users/jay/Developer/TeXProjects/test/test.tex
  main parser: current file verified
  document class: article
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: Skim
  qf method: LaTeX logfile
```

clason:
That was quick :) Yes, macOS 12.1 broke Skim's forward search functionality, see https://sourceforge.net/p/skim-app/bugs/1438/

There's unfortunately nothing that can be done about it right now; we'll have to wait for the Skim dev to release a new version to work around these changes (if that is possible).

Zeta611:
So is it broken regardless of the CPU architecture in macOS 12.1?

lervag:
[...]

PR #2286, opened 2021-12-26
add option to control whether Skim gets focus upon first open

yongrenjie:
Just a tiny PR which introduces an option `g:vimtex_view_skim_activate_on_start`, to control whether Skim is brought to the foreground when it's first opened (using `\ll` or `\lv`). The default value is kept to `1` to match the current behaviour, i.e. bring to foreground. Seems to work nicely on my Mac (Monterey 12.0 here).

(Somewhat related to #180; but rather less general, because it's only Skim and hence only macOS.)

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -426,6 +426,7 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_view_mupdf_send_keys', '')
   call s:init_option('vimtex_view_sioyek_exe', 'sioyek')
   call s:init_option('vimtex_view_skim_activate', 0)
+  call s:init_option('vimtex_view_skim_activate_on_start', 1)
   call s:init_option('vimtex_view_skim_reading_bar', 1)
   call s:init_option('vimtex_view_zathura_options', '')
   call s:init_option('vimtex_view_zathura_check_libsynctex', 1)
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -426,7 +426,7 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_view_mupdf_send_keys', '')
   call s:init_option('vimtex_view_sioyek_exe', 'sioyek')
   call s:init_option('vimtex_view_skim_activate', 0)
-  call s:init_option('vimtex_view_skim_activate_on_start', 1)
+  call s:init_option('vimtex_view_skim_latexmk_previewer', 'open -a Skim')
   call s:init_option('vimtex_view_skim_reading_bar', 1)
   call s:init_option('vimtex_view_zathura_options', '')
   call s:init_option('vimtex_view_zathura_check_libsynctex', 1)
--- a/autoload/vimtex/view/skim.vim
+++ b/autoload/vimtex/view/skim.vim
@@ -13,7 +13,6 @@ endfunction
 
 let s:viewer = vimtex#view#_template#new({
       \ 'name' : 'Skim',
-      \ 'startskim' : 'open -a Skim' . (g:vimtex_view_skim_activate_on_start ? '' : ' -g'),
       \})
 
 function! s:viewer.compiler_callback(outfile) dict abort " {{{1
@@ -74,7 +73,7 @@ endfunction
 
 " }}}1
 function! s:viewer._latexmk_append_argument() dict abort " {{{1
-  return vimtex#compiler#latexmk#wrap_option('pdf_previewer', self.startskim)
+  return vimtex#compiler#latexmk#wrap_option('pdf_previewer', g:vimtex_view_skim_latexmk_previewer)
 endfunction
 
 " }}}1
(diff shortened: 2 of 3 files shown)


lervag:
Thanks. However, I think this could be solved in a better way. Right now, the `startskim` attribute seems out of place, because it is only used as the string passed to latexmk to specify the viewer opened by latexmk. I.e., this setting is only used when the viewer is opened _implicitly_ by latexmk after we start compilation with `\ll`. So, your proposed setting would not have any effect on the `\lv` mapping.

Instead, perhaps we could instead make these changes:

* Remove the `startskim` key from the `s:viewer` dictionary.
* Add option `g:vimtex_view_skim_latexmk_option` which by default would be `open -a Skim`.

This would allow the configuration possibility you want in a way that seems to
me to be more clear. What do you think?

@clason - comments?

clason:
I don't quite understand what the difference is to the already existing `g:vimtex_view_skim_activate` option?

yongrenjie:
> I.e., this setting is only used when the viewer is opened implicitly by latexmk

Ah, I didn't look into the code very thoroughly, which is probably the source of my misunderstanding — but now I see what you mean. I agree with your suggestion @lervag — it is more technically correct and that's always a good thing! I can rework this a bit later.

> I don't quite understand what the difference is to the already existing `g:vimtex_view_skim_activate` option

@clason As I understand it (and in my experience), that option only applies to forward search, not the initial compilation (with latexmk).

Happy to hear any other suggestions!

clason:
I mean, why not use the `_start` (or `callback`) method for the initial compilation as well? Is there an actual need to have a third, potentially different, method?

yongrenjie:
That I'm not entirely sure about; after a bit of experimentation, it appears that the latexmk previewer option might not be needed, because the compiler callback opens Skim for the initial compilation anyway. With my PR, one would be able to control the initial compilation & subsequent compilations independently. (Indeed, the current default behaviour is already different, because Skim is focused initially and not subsequently.) But if that's not worth having, then it seems that it could be simplified.

clason:
>  With my PR, one would be able to control the initial compilation & subsequent compilations independently. 

Yes, the question is exactly whether that is desired. (Not for me, but other people may have different opinions, and those I'd like to hear.)

> Not sure where _start comes in, yet. It seems to me that that's triggered on forward searching, not compilation?

Yes, that is correct.

yongrenjie:
Ok, good, I'm on the same page! :smile:

Anyway, I renamed the option for the time being according to @lervag's comment. Feel free to ignore if you all want to move in a different direction.

clason:
If I understand correctly, the motivation for this change is _not_ to have Skim steal focus when first opening after a compilation?

If so, my proposal would be to make a different change:
1. remove the `latexmk_previewer` variable and invocation completely and just rely on the compiler callback;
2. add a `sync` variable to control whether the callback includes a forward search.

@lervag what is the role of the `s:viewer._latexmk_append_argument()` function? Can it simply be removed (I suspect that it is left over from a time before the current callback architecture)?

lervag:
> > I don't quite understand what the difference is to the already existing `g:vimtex_view_skim_activate` option
>
> As I understand it (and in my experience), that option only applies to forward search, not the initial compilation (with latexmk).

> what is the role of the `s:viewer._latexmk_append_argument()` function? Can it simply be removed (I suspect that it is left over from a time before the current callback architecture)?

Yes - the point is that latexmk has its own feature for opening the viewer, and that VimTeX has traditionally "played nice" with this feature by passing on the necessary option to make latexmk open the correct viewer in the correct way. This would typically imply that things are more or less seamless.

However, as is apparent e.g. from this issue, the result is not fully seamless, as we have things like `g:vimtex_view_skim_activate` that is not relevant if Skim is opened implicitly by latexmk.

> If I understand correctly, the motivation for this change is _not_ to have Skim steal focus when first opening after a compilation?
> 
> If so, my proposal would be to make a different change (not in this PR necessarily):
> 
>     1. remove the `latexmk_previewer` variable and invocation completely and just rely on the compiler callback;
> 
>     2. add a `sync` variable to control whether the callback includes a forward search.
> 
> **EDIT:** I could combine this with the switch to JavaScript, see [#2279 (comment)](https://github.com/lervag/vimtex/issues/2279#issuecomment-1002322442)

It's not an unreasonable proposal. But let's consider it properly. So, the idea is to generally prevent latexmk from opening the viewer and instead open the viewer from VimTeX through callbacks. The benefits:

* It is more or less already implemented and making this change should be easy.
* VimTeX can have full control of the viewer, and we don't need extra code/options to "patch" things to make them seamless.

The cons:

* I believe there are no real cons here. But there may be an issue with requirements that I'm not thinking of right now.

I don't mind having all of this implemented in #2289. So perhaps we should continue the discussion there?

yongrenjie:
OK, I checked out #2289 and it "works for me" from the perspective of a user. When I made this PR I had in mind to not *remove* any configurability / break somebody's workflow, but personally, I totally don't care about having separate behaviour for compilation & forward search. Essentially, I never want Skim to be focused :-)

I'll close this now and look forward to that being merged into master. Thanks for picking it up, @clason.

PR #2289, opened 2021-12-29
refactor skim viewer

### LINKED ISSUE #2279: Skim forward search not working on macOS 12.1

clason:
Refactor the Skim viewer backend to
* use JavaScript instead of AppleScript (nicer, and should fix #2279)
* make the initial start respect `g:vimtex_view_skim_activate`  (alternative to #2286, @yongrenjie )
* factor out common script parts into script function
* correct off-by-one line number for forward search
* introduce new option `g:vimtex_view_skim_sync` to perform forward  search after compilation (syncing viewer and editor), default false
* change default for `g:vimtex_view_skim_reading_bar` to false (location is highlighted anyway on sync)

### DIFF
--- a/autoload/vimtex/view/skim.vim
+++ b/autoload/vimtex/view/skim.vim
@@ -4,42 +4,59 @@
 " Email:      karl.yngve@gmail.com
 "
 
+let s:viewer = vimtex#view#_template#new({ 'name' : 'Skim' })
+
 function! vimtex#view#skim#new() abort " {{{1
   return s:viewer.init()
 endfunction
 
 " }}}1
 
+function! s:make_cmd_view(outfile, sync) abort " {{{1
+  let cmd_view = join([ 'osascript -l JavaScript -e ''',
+        \ 'var app = Application("Skim");',
+        \ 'var theFile = Path("' . a:outfile . '");',
+        \ 'try { var theDocs = app.documents.whose({ file: { _equals: theFile }});',
+        \ 'if (theDocs.length > 0) app.revert(theDocs) }',
+        \ 'catch (e) {};',
+        \ 'app.open(theFile);' ])
+  if a:sync
+    let cmd_view .= join([
+          \ 'app.documents[0].go({ to: app.texLines[' . (line('.')-1) . '],',
+          \ 'from: Path("'. expand('%:p') . '")',
+          \ (g:vimtex_view_skim_reading_bar ? ', showingReadingBar: true' : ''),
+          \ '});' ])
+  endif
+  if g:vimtex_view_skim_activate
+    let cmd_view .= 'app.activate();'
+  endif
+  return cmd_view . ''''
+endfunction
 
-let s:viewer = vimtex#view#_template#new({
-      \ 'name' : 'Skim',
-      \ 'startskim' : 'open -a Skim',
-      \})
+" }}}1
 
 function! s:viewer.compiler_callback(outfile) dict abort " {{{1
-  let self.cmd_view_callback = join([
-        \ 'osascript',
-        \ '-e ''set theFile to POSIX file "' . a:outfile . '"''',
-        \ '-e ''set thePath to POSIX path of (theFile as alias)''',
-        \ '-e ''tell application "Skim"''',
-        \ '-e ''try''',
-        \ '-e ''set theDocs to get documents whose path is thePath''',
-        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
-        \ '-e ''end try''',
-        \ '-e ''open theFile''',
-        \ '-e ''end tell''',
-        \])
+  call vimtex#jobs#run(s:make_cmd_view(a:outfile, g:vimtex_view_skim_sync))
+endfunction
 
-  call vimtex#jobs#run(self.cmd_view_callback)
+" }}}1
+
+function! s:viewer._start(outfile) dict abort " {{{1
+  call vimtex#jobs#run(s:make_cmd_view(a:outfile, 1))
+endfunction
+
+" }}}1
+
+function! s:viewer._latexmk_append_argument() dict abort " {{{1
+  return vimtex#compiler#latexmk#wrap_option('pdf_previewer',
+        \ 'open -a Skim' . g:vimtex_view_skim_activate ? ' -g' : '')
 endfunction
 
 " }}}1
 
 function! s:viewer._check() dict abort " {{{1
-  " Check if Skim is installed
   let l:output = vimtex#jobs#capture(
-        \ 'osascript -e '
-        \ . '''tell application "Finder" to get id of application "Skim"''')
+        \ 'osascript -l JavaScript -e ''Application("Skim").id()''')
 
   if l:output[0] !~# '^net.sourceforge.skim-app'
     call vimtex#log#error('Skim is not installed!')
@@ -47,34 +64,6 @@ function! s:viewer._check() dict abort " {{{1
   endif
 
   return v:true
-endfunction
-
-" }}}1
-function! s:viewer._start(outfile) dict abort " {{{1
-  let self.cmd_view = join([
-        \ 'osascript',
-        \ '-e ''set theLine to ' . line('.') . ' as integer''',
-        \ '-e ''set theFile to POSIX file "' . a:outfile . '"''',
-        \ '-e ''set thePath to POSIX path of (theFile as alias)''',
-        \ '-e ''set theSource to POSIX file "' . expand('%:p') . '"''',
-        \ '-e ''tell application "Skim"''',
-        \ '-e ''try''',
-        \ '-e ''set theDocs to get documents whose path is thePath''',
-        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
-        \ '-e ''end try''',
-        \ '-e ''open theFile''',
-        \ '-e ''tell front document to go to TeX line theLine from theSource',
-        \ g:vimtex_view_skim_reading_bar ? 'showing reading bar true''' : '''',
-        \ g:vimtex_view_skim_activate ? '-e ''activate''' : '',
-        \ '-e ''end tell''',
-        \])
-
-  call vimtex#jobs#run(self.cmd_view)
-endfunction
-
-" }}}1
-function! s:viewer._latexmk_append_argument() dict abort " {{{1
-  return vimtex#compiler#latexmk#wrap_option('pdf_previewer', self.startskim)
 endfunction
 
 " }}}1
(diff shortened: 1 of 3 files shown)


lervag:
> * use JavaScript instead of AppleScript (nicer, and should fix [Skim forward search not working on macOS 12.1 #2279](https://github.com/lervag/vimtex/issues/2279))

Great!

> * make the initial start respect `g:vimtex_view_skim_activate`  (alternative to [add option to control whether Skim gets focus upon first open #2286](https://github.com/lervag/vimtex/pull/2286), @yongrenjie )

Great!

> * factor out common script parts into `self` variables

Not quite sure if this has any specific merits..?

> * correct off-by-one line number for forward search

Great!

> * introduce new option `g:vimtex_view_skim_sync` to perform forward  search after compilation (syncing viewer and editor), default true

Does this sync on _every_ callback, or just the initial one? I believe there may be three main categories here:

* The user does not want any forward search when the viewer starts or on callbacks.

* The user wants a forward search just when the viewer starts (either manually or through callback).

* The user wants forward search on every callback.

> * change default for `g:vimtex_view_skim_reading_bar` to false (location is highlighted anyway on sync)

I don't have any opinions on this.

autoload/vimtex/view/skim.vim
```diff
@@ -10,24 +10,33 @@ endfunction
...
-        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
-        \ '-e ''end try''',
-        \ '-e ''open theFile''',
-        \ '-e ''end tell''',
+        \ 'osascript -l JavaScript',
+        \ '-e ''var theFile = Path("' . a:outfile . '")''',
+        \ '-e ''var theLine = ' . (line('.')-1) . '''',
+        \ '-e ''var theSource = Path("' . expand('%:p') . '")''',
+        \ self.show,
+        \ g:vimtex_view_skim_sync ? self.sync : '',
```
lervag, line 39: Do I remember correctly that this stuff is required to force Skim to update after successful compilation? I.e., without this, Skim would still show the old version of the PDF?

If possibly, I think it would be good to add support for `g:vimtex_view_automatic` here similar to how it's done for Zathura and MuPDF, see here:

https://github.com/lervag/vimtex/blob/f31b723d16fa00137ccbe51fa771a90dc642b05d/autoload/vimtex/view/_template.vim#L236-L253

autoload/vimtex/view/skim.vim
```diff
@@ -10,24 +10,33 @@ endfunction
...
+      \ '-e ''if (theDocs.length > 0) app.revert(theDocs)''',
+      \ '-e ''}''',
+      \ '-e ''catch (e) {}''',
+      \ '-e ''app.open(theFile)''',
+      \]),
+      \ 'sync' : join([
+      \ '-e ''app.documents[0].go({ to: app.texLines[theLine], from: theSource',
+      \ g:vimtex_view_skim_reading_bar ? ', showingReadingBar: true' : '',
+      \ '})''',
+      \]),
```
lervag, line 29: I think the main purpose for adding these variables as dictionary keys is to prevent duplicate code. As far as I can see, the `start` key is only used once. `show` and `sync` are used twice. I therefore believe it would be better to remove the `start` key and instead type it out verbosely (unless we end up removing the `_latexmk_append_argument` stuff altogether). I'm also wondering if the code could be slightly cleaner if the `show` and `sync` keys are instead changed to script variables; it would never make sense for a user outside of this scripts scope to inspect these variables, right?

autoload/vimtex/view/skim.vim
lervag: Is it possible to implement an `_exists` method that would be reliable for Skim?

lervag (changes requested):

clason:
> Not quite sure if this has any specific merits..?

Makes it a bit easier to maintain (it's hard enough not to lose track of all the quotes for the embedded scripting...). An alternative would be to move the complete thing out into a (private) `view` function that takes options to control the behavior, depending on whether it's run from a callback or a forward search.

> Does this sync on every callback, or just the initial one? I think there are two different scenarios here:

It syncs on every callback. I agree that this may not be desirable (in fact, my motivation for the original refactor to get rid of `displayline` was exactly _not_ to do that ;)), but there's a tradeoff here between features and complexity; this strikes me as a reasonable one, at least as long as VimTeX doesn't easily provide the information whether it's the first callback or subsequent ones?

autoload/vimtex/view/skim.vim
clason: No, that is for doing a forward search (syncing editor and viewer, which I know some people like). The update happens in every case (via the `app.revert(theDocs)` line) as I don't see any reason _not_ to have this.

clason: I can, however, put this behind a `g:vimtex_view_automatic` guard -- if there is a good reason for it?

autoload/vimtex/view/skim.vim
clason: I think the main question is when the config variables are evaluated. I believe the reason to have them in the template is to allow updating it via `:VimtexReload`?

But this relates to the comment below: I'm happy to do the refactoring in another way; I just wanted to stay as close to the current setup as I don't fully understand the control flow here.

clason:
I think there are three scenarios:
1. a forward search, which I believe should sync unconditionally (after all, that's the meaning of "forward search");
2. a compiler callback, which some people would like to keep synced while others would hate that, hence should be an option;
3. a compiler callback that also starts the viewer -- optionally followed by a forward search, which I believe most people would want.

As far as I can tell, the current architecture doesn't allow me to easily handle 3., since VimTeX itself doesn't know about this, and the `latexmk` option can't run vimscript functions?

clason:
@lervag I've refactored again to extract all the common JavaScript lines into a single `make_cmd_view` script-level _function_; that indeed looks cleaner (although the nested ternary is rather ugly).

The `sync` now defaults to `false` to preserve old behavior; I think this is now a strict improvements. Further improvements are possible but would need some architecture changes, I believe.

lervag:
Thanks @clason! I've merged now. But please note: I've made some important general changes based on the discussion we've had. I've now forced `-view=none` with the latexmk backend so as to have full control of the viewer from vimtex. I expect this to fail somewhat, actually, but I believe it is better to address those issues and find a more robust solution than to stay intertwined with latexmk.

So: Please test the latest version and let me know how it works. Especially: with the default settings with Skim, do things still work as expected?

clason:
Yes, looks fine to me! Except for the reading bar, I get the same behavior as before with the default settings (and `g:vimtex_view_skim_sync` works as expected).

lervag:
Does Skim start automatically after compilation?

Also, I think we want to support `g:vimtex_view_automatic` also for Skim.

clason:
> Does Skim start automatically after compilation?

Yes, it does.

> Also, I think we want to support `g:vimtex_view_automatic` also for Skim.

Why (honestly asking)? From what I've seen, this is optional since for many viewers, it requires workarounds using external tools and whatnot, which may have undesired side effects (such as slowdown). However, Skim supports this out of the box with the same interface, so it's a freebie. 

Can you think of any situation where you would _not_ want this?

lervag:
Ok, so, to be more clear about what I'm thinking about, please see how the general viewer now works wrt. callbacks:

https://github.com/lervag/vimtex/blob/72e6285b91a606940ec21c8cd5e587f842094ca4/autoload/vimtex/view/_template.vim#L80-L91

The callbacks are "activated" here (they are active in all cases):

https://github.com/lervag/vimtex/blob/72e6285b91a606940ec21c8cd5e587f842094ca4/autoload/vimtex/view.vim#L25-L29

And this shows how they are abstracted:

https://github.com/lervag/vimtex/blob/72e6285b91a606940ec21c8cd5e587f842094ca4/autoload/vimtex/view.vim#L50-L66

lervag:
> Why (honestly asking)? From what I've seen, this is optional since for many viewers, it requires workarounds using external tools and whatnot, which may have undesired side effects (such as slowdown). However, Skim supports this out of the box with the same interface, so it's a freebie.
> 
> Can you think of any situation where you would _not_ want this?

Yes, simple: I don't want `\ll` to open the viewer automatically. So, I add `let g:vimtex_view_automatic = 0` to prevent it. That just works. I use `\lv` to open the viewer when I want it.

Currently, this workflow is not possibly with Skim, because it does not support `g:vimtex_view_automatic`.

clason:
Interesting. But I don't see the issue then: Isn't it trivial to add that approach to the current callback, as in
```vim
function! s:viewer.compiler_callback(outfile) dict abort " {{{1 
   if !g:vimtex_view_automatic 
       \ || has_key(self, 'started_through_callback') | return | endif 
  
   call vimtex#jobs#run(s:make_cmd_view(a:outfile, g:vimtex_view_skim_sync))
   let self.started_through_callback = 1 
 endfunction 
```

lervag:
For other people reading here - me and @clason discussed things on Gitter and I've pushed a few more updates that we believe resolve these things for now. Feel free to open new issues if something broke!

PR #2308, opened 2022-01-27
Rewrite vimtex_syntax_conceal_disable doc

davidlowryduda:
When reading the documentation before, I was initially confused. It used
to refer to the option g:vimtex_syntax_conceal, and the following
sentence referred to *this option*. I had parsed that to mean
*g:vimtex_syntax_conceal* instead of *g:vimtex_syntax_conceal_disable*.

This commit rewords this part of the documentation.

As always, thank you for this wonderful plugin!


lervag:
Thanks, and thanks for the kind words! I pushed an additional update that I believe both simplifies and improves the docs further.

PR #2319, opened 2022-02-01
Improve highlight of commented code

julio-b:
First commit is a simple fix for `\iffalse <commented-code>[\else <not-commented-code>]\fi ` command.
In the second commit i tried to add highlights for `\iftrue <not-commented-code> [\else <commented code>]\fi ` command.
I've never used vim syntax before, so i'm not sure if this is the correct way to implement \iftrue.

I must admit that this pr is not very useful. But it allows you, for example, a quick switch draft mode by simply replacing `\iftrue` with `\iffalse` etc.
2022-02-01-210318_547x116_scrot
2022-02-01-210334_559x122_scrot

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -436,6 +436,25 @@ function! vimtex#syntax#core#init() abort " {{{1
   syntax region texComment matchgroup=texCmd
         \ start="^\s*\\iffalse\>" end="\\\(fi\|else\)\>"
         \ contains=texCommentConditionals
+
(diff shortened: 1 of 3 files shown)


lervag:
Thanks! I made a couple of minor adjustments myself, I hope you don't mind. I think you did a good implementation, but I want to keep it in my "style". And I used `containedin` to avoid the clusters.

julio-b:
Thank you! I don't mind at all, in fact your adjustments helped me to better understand the way that VimL combines regions with each other.
Thank you for this plugin!

lervag:
Great; glad to hear it :)

PR #2369, opened 2022-04-08
Adjust path parsed result for pplatex

JingMatrix:
Latex log assumes project root to be the start directory

### DIFF
--- a/autoload/vimtex/qf/pplatex.vim
+++ b/autoload/vimtex/qf/pplatex.vim
@@ -87,7 +87,9 @@ function! s:qf.addqflist(tex, log) abort dict " {{{1
   let l:tmp = fnamemodify(a:log, ':r') . '.pplatex'
 
   call vimtex#jobs#run(printf('pplatex -i "%s" >"%s"', a:log, l:tmp))
+  call vimtex#paths#pushd(b:vimtex.root)
   call vimtex#qf#u#caddfile(self, l:tmp)
+  call vimtex#paths#popd()
   call delete(l:tmp)
 endfunction
 


autoload/vimtex/qf/pplatex.vim
```diff
@@ -87,7 +87,10 @@ function! s:qf.addqflist(tex, log) abort dict " {{{1
   let l:tmp = fnamemodify(a:log, ':r') . '.pplatex'
 
   call vimtex#jobs#run(printf('pplatex -i "%s" >"%s"', a:log, l:tmp))
+  let l:pwd = getcwd()
+  execute 'lcd' b:vimtex.root
```
lervag, line 91: It would be better to use the built-in api: `call vimtex#paths#pushd(b:vimtex.root)`.

Similar to how it's done in `jobs#run`:

https://github.com/lervag/vimtex/blob/master/autoload/vimtex/jobs.vim#L40-L42

autoload/vimtex/qf/pplatex.vim
```diff
@@ -87,7 +87,10 @@ function! s:qf.addqflist(tex, log) abort dict " {{{1
...
   call vimtex#qf#u#caddfile(self, l:tmp)
+  execute 'lcd' l:pwd
```
lervag, line 93: `call vimtex#paths#popd()`

lervag (changes requested):
I don't see any reason not to accept, but I'll allow you to adjust as commented.

JingMatrix:
I force updated my commit now, the first one is a rebase for your commits, and the second one is to adopt your advices.

Thanks for your advice on `popd` and `pushd`, it is intuitive just as the shell language.

lervag:
Great, thanks! I'll merge now. :)

Issue #2342, opened 2022-03-03
[regression] zathura forward search opens a new instance when pwd is a symlink


MoetaYuko:
### Description

This is a regression after https://github.com/lervag/vimtex/commit/71554107aa2ca28aa93099bdb46cc0c0fdf69a52

When the real place of tex file is ~/foo and I'm working in a symlink ~/bar -> ~/foo, `<localleader>ll` opens a zathura _window A_, `<localleader>lv` opens another _window B_, where:
* backward search works in _window A_ only
* forward search works in _window B_ only

Moreover, when having the autocmd to close viewer on exit, only _window A_ gets closed.

### Steps to reproduce

1. mkdir foo
2. ln -s foo bar
3. cd bar
4. vim -u minimal.vim minimal.tex

minimal.vim
```vim
call plug#begin()
 Plug 'lervag/vimtex'
call plug#end()

set nocp

call remote_startserver('VIM')
let g:vimtex_view_method = 'zathura'

" Close viewers when VimTeX buffers are closed
function! CloseViewers()
  if executable('xdotool')
        \ && exists('b:vimtex.viewer.xwin_id')
        \ && b:vimtex.viewer.xwin_id > 0
    call system('xdotool windowclose '. b:vimtex.viewer.xwin_id)
  endif
endfunction
augroup vimtex_event_2
  au!
  au User VimtexEventQuit call CloseViewers()
augroup END
```
minimal.tex
```tex
\documentclass{minimal}
\begin{document}
Hello world!
\end{document}
```

### Expected behavior

forward search works in _window A_ and no new window opens

### Actual behavior

as described before

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info:
  OS: Arch Linux
  Vim version: VIM 8.2 (1-4464)
  Has clientserver: true
  Servername: VIM

VimTeX project: minimal
  base: minimal.tex
  root: /home/dianlujitao/foo
  tex: /home/dianlujitao/foo/minimal.tex
  main parser: current file verified
  document class: minimal
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: Zathura
    xwin id: 0
  qf method: LaTeX logfile
```

lervag:
Thanks! Should work as expected now.

MoetaYuko:
[...]

PR #2377, opened 2022-04-24
Use absolute path for Zathura forward search

GianCarloMilanese:
Hello,

I have always had issues with Zathura forward search (backward works fine on the other hand).

The only way to fix it for me is to use absolute paths instead of relative paths.

Maybe someone else has the same issue, I hope this is somewhat useful.

Thanks for the plugin, I have been using it for quite a while :)

### DIFF
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -99,7 +99,7 @@ function! s:cmdline(outfile, synctex, start) abort " {{{1
           \ ' --synctex-forward %d:%d:%s',
           \ line('.'), col('.'),
           \ vimtex#util#shellescape(
-          \   vimtex#paths#relative(expand('%:p'), b:vimtex.root)))
+          \   expand('%:p')))
   endif
 
   return l:cmd . ' '


lervag:
> I have always had issues with Zathura forward search (backward works fine on the other hand).
> 
> The only way to fix it for me is to use absolute paths instead of relative paths.

The suggested change addresses code that was last modified for #2342, and there's been no complaint on this issue since then (until now). I know for a fact that a lot of people use VimTeX + Zathura on Linux machines, so: It would be very useful to have an example where I can reproduce the problem.

Can you provide:

0. Some basic information on your system, such as OS, vim version, etc. Most of this is provided by the output of `:VimtexInfo`.

1. A minimal example (`test.vim` and `test.tex`) for reproduction of the problem. E.g.

  ```vim
  " test.vim
  set nocompatible
  set runtimepath^=~/.local/plugged/vimtex
  set runtimepath+=~/.local/plugged/vimtex/after
  filetype plugin indent on
  syntax enable

  " other options/config?

  silent edit test.tex
  ```

  ```tex
  % test.tex
  \documentclass{minimal}
  \begin{document}

  Hello World!

  \end{document}
  ```

2. Steps to reproduce, e.g. `nvim --clean -u test.vim` + `\ll`. Go to "Hello" and do `\lv` - should perform forward search.

> Thanks for the plugin, I have been using it for quite a while :)

Glad you like it!

GianCarloMilanese:
Thanks a lot for your time. I'm sorry for the late reply.

My OS is Ubuntu 20.04.3 LTS, and my VIM version is NVIM v0.8.0-dev.

I have tried with the following files:

```vim
" test.vim
set nocompatible
set runtimepath^=~/.local/plugged/vimtex
set runtimepath+=~/.local/plugged/vimtex/after
filetype plugin indent on
syntax enable

let g:vimtex_view_method = "zathura"

silent edit test.tex
```

```tex
% test.tex
\documentclass{minimal}
\usepackage{lipsum}
\begin{document}

\lipsum[1]

\newpage
\lipsum[2]

\newpage
\lipsum[3]

\end{document}
```

I have cloned the latest version of vimtex in `~/.local/plugged/vimtex` and run `nvim --clean -u test.vim`.

After `\ll`, the pdf opens on the first page. Doing `\lv` while my cursor is on any line of the `tex` file will always move the pdf to the first page. E.g., if the pdf is on page 3 and I do `\lv` while on the line containing `\lipsum[2]`, the pdf will focus on the first page instead of the second page.

After changing line 102 of `autoload/vimtex/view/zathura.vim` like in the PR so that it uses absolute paths, the forward search works fine.

Thanks again. Let me know if you need any other information.

lervag:
I still can't reproduce this. I do this:

```sh
cd ~
mkdir test
cd test
nvim test.vim # add content from the minimal vimrc
nvim test.tex # add content from the minimal tex file
nvim --clean -u test.vim
```

then

1. `\ll` - observe that compilation is successful and a three page pdf opens in Zathura.
2. `\lv` - works as expected for me generally.

Note, I don't mind making the change if I can observe specifically that it is necessary, but I hesitate to do it before I am sure it actually solves anything. There is a risk that it may _break_ something for some users as well, which is why I want to reproduce this first.

lervag:
Could this be related to synctex version?

lervag:
I believe #2392 is a duplicate of this issue. I'm closing this due to inactivity; I hope this will be resolved after fixing #2392.

PR #2387, opened 2022-05-08
add support for TeXShop on macOS (MacVim)

michaelliebling:
Add support for TeXShop on macOS. TeXShop offers:

1. a TeX-aware text editor (irrelevant for vim users)
2. a PDF viewer compatible with backward search
3. a compiler front-end for various TeX-related software (pdfTeX, LuaTeX, etc.).
 
The proposed features are related to 2. (backward search from PDF to MacVim) and 3. (backward search from compiler console to MacVim). The latter might stand out compared to other tools (viewers) vimtex usually interacts with (alternative to quickfix).

The following files are affected (or added) by this pull request:

* `README.md`: mention TeXShop in the list of supported viewers.
* `autoload/health/vimtex.vim`: add a command to check whether TeXShop exists
* `autoload/vimtex/options.vim`: add the two options  'vimtex_view_texshop_activate' and 'vimtex_view_texshop_sync'.
* `autoload/vimtex/view/texshop.vim`: viewer interaction with TeXShop (forward search) based on the latest viewer for Skim.
* `doc/vimtex.txt`: add documentation on a. how to use vimtex with TeXShop (succinct, since it will just work like for other viewers on the vim side) and b. on how to set up TeXShop to play with MacVim (a bit longer, in the FAQ, since that's more unique).

### DIFF
--- a/autoload/vimtex/fold/markers.vim
+++ b/autoload/vimtex/fold/markers.vim
@@ -35,10 +35,10 @@ endfunction
 " }}}1
 function! s:folder.level(line, lnum) abort dict " {{{1
   if a:line =~# self.re.start
-    let s:self.opened = 1
+    let self.opened = 1
     return 'a1'
   elseif a:line =~# self.re.end
-    let s:self.opened = 0
+    let self.opened = 0
     return 's1'
   endif
 endfunction


autoload/vimtex/view/texshop.vim
```diff
@@ -0,0 +1,104 @@
...
+        \ '-e ''set theDocs to get documents whose path is thePath''',
+        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
+        \ '-e ''end try''',
+        \ a:open ? '-e ''open theFile''' : '',
+        \ g:vimtex_view_texshop_activate ? '-e ''activate''' : '',
+        \ '-e ''end tell''',
+        \]
+  let l:script = join(l:scriptview)
+
+" Define variables for the source file, line and column numbers:
```
lervag, line 63: I prefer to keep the comments indented at same level as the code.

autoload/vimtex/view/texshop.vim
```diff
@@ -0,0 +1,104 @@
...
+  let sourcefile = shellescape(expand('%'), 1)
```
lervag, line 64: I prefer to be explicit about variable locality, i.e. please prepend `l:` here and for similar variables. That is, change `let sourcefile` to `let l:sourcefile`.

`l:` specifies that the variable is function local. In this context, it will be local by default, so this is only a minor thing. But I prefer explicit.

autoload/vimtex/view/texshop.vim
```diff
@@ -0,0 +1,104 @@
...
+        \ g:vimtex_view_texshop_activate ? '-e ''activate''' : '',
+        \ '-e ''    return 0''',
+        \ '-e ''  end tell''',
+        \ '-e ''end tell''',
+        \]
+" An alternative way of defining the command could have been
+" to use a bash script that encapsulates the Applescript (has
+" the advantage that the bashscript can be tested on its own)
+"  let l:cmd = '/usr/local/bin/ExternalSync '
+"  let l:cmd .= join([linenr, colnr , sourcefileFull, outfile])
```
lervag, line 95: I'm not sure if the above comment has any specific value here? Perhaps just remove it; unless you think the alternative way is superior somehow?

doc/vimtex.txt
```diff
@@ -5284,6 +5285,22 @@ Note: There is a known issue with VimTeX + SumatraPDF when you use `xelatex`,
...
+call TeX programs to process source files, and provides a PDF viewer
+that supports syncing with the corresponding source. |vimtex| supports
+forward searches (source to PDF) with TeXShop.
+
+Configuration: >
+  let g:vimtex_view_method = 'texshop'
+
+Note: See |vimtex-faq-texshopviewer| for more info on how to properly set
+up the TeXShop (viewer and compilation front-end) for backward search.
+
```
lervag, line 5303: Please add `Associated settings:` paragraph (see docs for Skim for an example). Also: Please add docs for the options themselves (e.g. somewhere close to line 2927).

doc/vimtex.txt
```diff
@@ -5808,6 +5825,90 @@ A: Yes, it should work.  The following recipe has been reported to work [0].
...
+<
+   If you wish for the TeXShop window to pop to the front during forward
+   search, set the followin variable (e.g. in your `~/.vimrc` file): >
+
+   let g:vimtex_view_texshop_activate=1
+<
+   Otherwise (to keep the TeXShop window in the background) set: >
+
+   let g:vimtex_view_texshop_activate=0
+<
```
lervag, line 5891: This may be redundant if/when you document the option directly. Then it could suffice to refer to the option with the tag `|g:vimtex_view_texshop_activate|`.

lervag (changes requested):
Thanks! This is really close, I only have a few more minor comments - I hope you don't mind.

One more important question, though: In December/January I added an improved method for doing inverse search. Perhaps you could check if this is possible to integrate also for TeXShop? Please read `:help vimtex-synctex-inverse-search` once in detail, I believe it may already be more or less fully relevant. Perhaps also relevant to read the `:help vimtex-view-skim` sections for an example on how I document the inverse search for Skim.

I am not surprised if we keep the faq section in general, but perhaps it is possible to simplify things more?

lervag:
@michaelliebling I'm curious, do you want me to close this PR, or should I expect a reply to my comments?

autoload/vimtex/view/texshop.vim
michaelliebling: Done: cleaned up comments, matched indentation of comments to that of code.

autoload/vimtex/view/texshop.vim
michaelliebling: Done: I've given all variables explicit scope

autoload/vimtex/view/texshop.vim
michaelliebling: Done: removed that comment

doc/vimtex.txt
michaelliebling: Done: added an `Associated settings:` entry (it matches skim's format for a similar feature)

doc/vimtex.txt
michaelliebling: Removed this text and instead added a description for this option with the tag `|g:vimtex_view_texshop_activate|` as suggested.

michaelliebling:
Thanks for the comments and checks. I've shortened the docs to match those for other viewers (in particular, skim) only kept extra text for the extra features that TeXshop provides (i.e. when using its compiler front-end, one can Inverse Search the error during interactive compilation) and its expectations regarding the existence of an external bash script that calls the external editor. 

As you suggested, the generic `VimtexInverseSearch` approach does indeed work well for TeXshop as well. One thing that took me a while to figure out: somehow I had missed the `-v` option in the call to mvim. This turns out to be *essential* to avoid having MacVim reload the window with an annoying opening animation at each Inverse Search. 

I've moved the explicit script example to the `vimtex-faq-texshopviewer` section to keep the main documentation lean (again, matching skim's). In the faq, I describe using `VimtexInverseSearch` first, but nevertheless kept the mention to the `--remote-silent` alternative as it appears to be the only way to force-open a hidden buffer in MacVim (or even have the file opened by MacVim if it isn't open yet).

lervag:
Thank you! I think this is very good and I will be glad to merge it. However, I don't want the messy merge commits here. Can you please make a new fork of VimTeX and commit the changes on top of that as a single commit, then open a new PR? I'll merge it asap after that.

Alternatively, I can do this myself on my end. That means less work for you, but I may end up loosing your git signatures - that is, then it may seem like I authored the work. I would therefore prefer that you update on your end so that the authorship is correctly preserved.

PR #2403, opened 2022-06-01
Add support for tabularx

MarcelRobitaille:
This properly highlights the arguments of the `tabularx` environment, which takes two arguments unlike `tabular` that only takes one. The reason I added this was so that the second argument would properly get recognized as `@texClusterTabular`. Otherwise, the column specifiers get marked as a misspelled word when using `set spell`. This is a big problem for me because I use `]s` and `[s` to jump around and fix things. It is also preferable to have proper highlight groups for more reasons than just spelling.

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -308,22 +308,6 @@ function! vimtex#syntax#core#init() abort " {{{1
         \ 'contains': '@texClusterTabular'
         \})
 
-  syntax match texCmdTabularx "\\begin{tabularx}"
-        \ skipwhite skipnl
-        \ nextgroup=texTabularxOpt,texTabularxWidth
-        \ contains=texCmdEnv
-  call vimtex#syntax#core#new_opt('texTabularxOpt', {
-        \ 'next': 'texTabularxWidth',
-        \ 'contains': 'texComment,@NoSpell',
-        \})
-  call vimtex#syntax#core#new_arg('texTabularxWidth', {
-        \ 'next': 'texTabularxArg',
-        \})
-  call vimtex#syntax#core#new_arg('texTabularxArg', {
-        \ 'contains': '@texClusterTabular'
-        \})
-
-
   syntax match texTabularCol   "[lcr]" contained
   syntax match texTabularCol   "p"     contained nextgroup=texTabularLength
   syntax match texTabularAtSep "@"     contained nextgroup=texTabularLength
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -308,6 +308,22 @@ function! vimtex#syntax#core#init() abort " {{{1
         \ 'contains': '@texClusterTabular'
         \})
 
+  syntax match texCmdTabularx "\\begin{tabularx}"
+        \ skipwhite skipnl
+        \ nextgroup=texTabularxOpt,texTabularxWidth
+        \ contains=texCmdEnv
+  call vimtex#syntax#core#new_opt('texTabularxOpt', {
+        \ 'next': 'texTabularxWidth',
+        \ 'contains': 'texComment,@NoSpell',
+        \})
+  call vimtex#syntax#core#new_arg('texTabularxWidth', {
+        \ 'next': 'texTabularxArg',
+        \})
+  call vimtex#syntax#core#new_arg('texTabularxArg', {
+        \ 'contains': '@texClusterTabular'
+        \})
+
+
   syntax match texTabularCol   "[lcr]" contained
   syntax match texTabularCol   "p"     contained nextgroup=texTabularLength
   syntax match texTabularAtSep "@"     contained nextgroup=texTabularLength
--- a/autoload/vimtex/syntax/p/tabularx.vim
+++ b/autoload/vimtex/syntax/p/tabularx.vim
@@ -7,22 +7,24 @@
 function! vimtex#syntax#p#tabularx#load(cfg) abort " {{{1
   call vimtex#syntax#packages#load('array')
 
+  " The format is \begin{tabularx}{WIDTH}[POS]{PREAMBLE}
   syntax match texCmdTabularx "\\begin{tabularx}"
         \ skipwhite skipnl
-        \ nextgroup=texTabularxOpt,texTabularxWidth
+        \ nextgroup=texTabularxWidth
         \ contains=texCmdEnv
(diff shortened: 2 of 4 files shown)


autoload/vimtex/syntax/core.vim
```diff
@@ -308,6 +308,22 @@ function! vimtex#syntax#core#init() abort " {{{1
...
+        \ 'next': 'texTabularxWidth',
+        \ 'contains': 'texComment,@NoSpell',
+        \})
+  call vimtex#syntax#core#new_arg('texTabularxWidth', {
+        \ 'next': 'texTabularxArg',
+        \})
+  call vimtex#syntax#core#new_arg('texTabularxArg', {
+        \ 'contains': '@texClusterTabular'
+        \})
+
```
lervag, line 325: Nice, this looks good. But it should be moved to [`autoload/vimtex/syntax/p/tabularx.vim`](https://github.com/lervag/vimtex/blob/master/autoload/vimtex/syntax/p/tabularx.vim), as this is not core syntax stuff.

lervag (changes requested):
Thanks! In addition to my comment to the code, could you also add a test file for this under `test/test-syntax/test-tabularx.{vim,tex}`?

lervag:
Also sorry for the late reply!

MarcelRobitaille:
> Also sorry for the late reply!

Don't worry about it. I know how it is to be an open source developer :)

I moved the files and added some tests.

I know I'm new around here, but have you considered using something like vader for the tests? It would be really nice to run a single test at a time. As far as I can tell, you can't with this makefile.

lervag:
> > Also sorry for the late reply!
> 
> Don't worry about it. I know how it is to be an open source developer :)

:)

> I moved the files and added some tests.

Great; thanks!

> I know I'm new around here, but have you considered using something like vader for the tests? It would be really nice to run a single test at a time. As far as I can tell, you can't with this makefile.

I actually much prefer my way of testing. It is really quite easy to run a single test at a time:

```sh
cd $VIMTEX/test

# Run ALL tests
make

# Run all test-syntax tests
make -C test-syntax

# Run all test-syntax tests (alternative)
cd test-syntax
make

# Run only the tabularx test
make test-tabularx

# Run the tabularx tests but stay within neovim/Vim
nvim --clean -u test-tabularx.vim
vim -u test-tabularx.vim
```

test/test-syntax/test-tabularx.tex
```diff
@@ -0,0 +1,14 @@
+\documentclass{article}
+
+\usepackage{tabularx}
+
+\begin{document}
+	\begin{tabularx}{\linewidth}{llXX}
+		1 & 2 & 3 & 4 \\
+	\end{tabularx}
+
+	% Tabularx does not seem to have options
```
lervag, line 10: What do you mean with this? In the implementation, you do allow options..?

lervag:
I think the only thing lacking now is to set the default highlight groups. I think you can use this as an inspiration:

https://github.com/lervag/vimtex/blob/master/autoload/vimtex/syntax/p/array.vim#L56-L66

test/test-syntax/test-tabularx.tex
MarcelRobitaille: I added `texTabularxOpt` for consistency with normal `tabular`, but as far as I can tell reading the [`tabularx` docs](https://texdoc.org/serve/tabularx.pdf/0), there are no supported `[]` options for this environment. Should we keep it like this for consistency and in case they decide to add some, or should we take it away?

MarcelRobitaille:
I added some default highlight groups. How does that look?

test/test-syntax/test-tabularx.tex
lervag: I think we should aim to implement the correct behaviour. If you are right, then we should not include the option group. I don't think you are wrong, but if so, it is easy to adjust later if someone shows a relevant proof.

autoload/vimtex/syntax/p/tabularx.vim
```diff
@@ -21,6 +21,10 @@ function! vimtex#syntax#p#tabularx#load(cfg) abort " {{{1
   call vimtex#syntax#core#new_arg('texTabularxArg', {
         \ 'contains': '@texClusterTabular'
         \})
+
+  highlight def link texTabularxArg         texOpt
+  highlight def link texTabularxWidth       texOpt
```
lervag, line 26: Perhaps it would be better to link this to `texLength`? If it is a number, at least?

lervag:
Looks more or less good! I have one comment, and also a response regarding the option group. I looked at the docs, and I notice the syntax description looks like this:

```tex
\begin{tabularx}{WIDTH}[POS]{PREAMBLE}
```

Thus, there is a possible option group after the WIDTH argument. The WIDTH argument does not need a default highlight group, but it should contain the `texLength` group to match length specifications. Give it one more shot, and in the next iteration I'll either merge (possibly with minor adjustments) or be more concrete and helpful!

MarcelRobitaille:
Sorry for the delay. I think I got what you mean. Also, thanks for explaining how to run only the tests for `test-tabularx`. I am not an expert with Makefiles.

I just realise that the first link for `tabularx` points to the doc for the 1999 version of tabularx, which has no `[pos]` as far as I can tell. I see the option in the 2020 version.

autoload/vimtex/syntax/p/tabularx.vim
MarcelRobitaille: I might be getting the vocabulary wrong, but I think it's a length, not a number. For example, it's quite common to use `\tabularx{\linewidth}...`.

lervag: Yes, I believe it is a length. The `texLength` group matches things like `1em` and similar, and I believe it is a better default highlight group.

lervag:
> Sorry for the delay. I think I got what you mean. Also, thanks for explaining how to run only the tests for `test-tabularx`. I am not an expert with Makefiles.

No problem!

> I just realise that the first link for `tabularx` points to the doc for the 1999 version of tabularx, which has no `[pos]` as far as I can tell. I see the option in the 2020 version.

Ah, cool. Then we should keep the `[pos]` argument. It seems it remains to change the default highlight group for `texTabularxWidth`?

MarcelRobitaille:
> It seems it remains to change the default highlight group for texTabularxWidth?

Sorry, I thought you meant to put the `'contains': 'texLength'` like I did on line 17. Did I get it right now?

lervag:
Thanks, I've merged now with a few minor adjustments.

PR #2404, opened 2022-06-01
Support `\autoref{}` and `\cref{}`

MarcelRobitaille:
Add support for the `\autoref{}` command and the [`cleveref`](https://ctan.org/pkg/cleveref) packages `\cref{}` and `\Cref{}` commands. This properly treats the argument as `texRefArg`, highlighting it correctly and removing spell checking.

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -258,7 +258,7 @@ function! vimtex#syntax#core#init() abort " {{{1
   " Reference and cite commands
   syntax match texCmdRef nextgroup=texRefArg           skipwhite skipnl "\\nocite\>"
   syntax match texCmdRef nextgroup=texRefArg           skipwhite skipnl "\\label\>"
-  syntax match texCmdRef nextgroup=texRefArg           skipwhite skipnl "\\\(page\|eq\)ref\>"
+  syntax match texCmdRef nextgroup=texRefArg           skipwhite skipnl "\\\(page\|eq\|auto\|[cC]\)ref\>"
   syntax match texCmdRef nextgroup=texRefArg           skipwhite skipnl "\\v\?ref\>"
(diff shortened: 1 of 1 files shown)


lervag:
The `\cref` and `\Cref` commands _are_ properly matched already. The code is here:

https://github.com/lervag/vimtex/blob/master/autoload/vimtex/syntax/p/cleveref.vim#L9-L10

Notice also that the relevant changes should go in this file. The `\autoref` is probably missing, though, so I'll be glad to add that. Do you want to update the PR, or should I just add it?

MarcelRobitaille:
Hmm. `\Cref{}` is was not working for me before this. Should I open an issue then?

lervag:
Yes, or we can just continue the discussion here. Please note that the syntax packages such as for cleveref are only loaded when the packages is recognized. Perhaps that's the issue here?

In any case, I think it is relevant to include examples and/or tests also for PRs.

image

lervag:
I noticed I missed an example for `\Cref` in the test file; I just added that. You can manually test on your side like this:

```sh
cd $VIMTEX/test/test-syntax

# Neovim
nvim --clean -u test-cleveref.vim

# Vim
vim -u test-cleveref.vim
```

MarcelRobitaille:
Ok, that little example is working for me. Maybe I am experiencing this because the file I am editing does not use `\usepackage{cleveref}`, but rather is imported from another file.

lervag:
Things should still work as expected in multi file projects, as long as VimTeX recognizes the main file from the imported file. You can check with `:VimtexInfo` - it should show the detected main file. See also `:help vimtex-multi-file`. Feel free to open an issue or continue this discussion if you have further problems.

MarcelRobitaille:
Thanks, I will try that next time it comes up.

lervag:
Does it mean we can close this PR?

PR #2411, opened 2022-06-12
sioyek forward-search column number

JingMatrix:
Well, please don't merge it now; otherwise it will break.
It depends on my pull request https://github.com/ahrm/sioyek/pull/271 , which is waiting for merging.


JingMatrix:
That request might be hard to merge, but a new request is more likely to be accepted: https://github.com/ahrm/sioyek/pull/273, so I should reopen it.

lervag:
Thanks! I'll look into this when the sioyek PR has been merged; ok?

JingMatrix:
The pull request is merged now, but for stability reason, current pull-request should wait for next release of `sioyek`.

lervag:
Thanks! I noticed that the latest release was in August, which means I should probably have merged this a long time ago. Better late then never, I guess :)

PR #2414, opened 2022-06-15
Enable custom commands for inverse search in Zathura

gabrielbdsantos:
This PR adds `g:vimtex_view_zathura_synctex_command` for calling custom inverse search commands in Zathura. Even though  `VimtexInverseSearch` works in most cases, users could make use of a custom command for finer control.

(Not so) minimal working example.

minimal.vim
```vim
set nocompatible
let &rtp = '../..,' . &rtp
let &rtp .= ',../../after'
filetype plugin indent on
syntax enable

nnoremap q :qall!<cr>

let g:vimtex_view_automatic = 0
let g:latex_view_general_viewer = 'zathura'
let g:vimtex_view_method = 'zathura'
let g:vimtex_compiler_method = 'latexmk'
let g:vimtex_compiler_progname = 'nvr'

" Using neovim-remote. Requires `pip install neovim neovim-remote`.
let g:vimtex_view_zathura_synctex_command =
      \ '"nvr --servername ' .
      \ v:servername .
      \ ' --remote-silent +%{line} %{input}"'

" or using just neovim.
" let g:vimtex_view_zathura_synctex_command =
"       \ '"nvim --server ' .
"       \ v:servername .
"       \ ' --remote-send '':e %{input} | :%{line}<CR>''"'

silent edit main.tex
```

main.tex
```tex
\documentclass{minimal}
\begin{document}
Hello world!

Line 5

Line 7

Line 9
\end{document}
```

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -461,6 +461,7 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_view_skim_reading_bar', 0)
   call s:init_option('vimtex_view_zathura_options', '')
   call s:init_option('vimtex_view_zathura_check_libsynctex', 1)
+  call s:init_option('vimtex_view_zathura_synctex_command', '')
 
   " Fallback option
   if g:vimtex_context_pdf_viewer ==# 'NONE'
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -88,9 +88,13 @@ function! s:cmdline(outfile, synctex, start) abort " {{{1
   if a:start
     let l:cmd .= ' ' . g:vimtex_view_zathura_options
     if a:synctex
-      let l:cmd .= printf(
-            \ ' -x "%s -c \"VimtexInverseSearch %%{line} ''%%{input}''\""',
-            \ s:inverse_search_cmd)
+      if empty(g:vimtex_view_zathura_synctex_command)
+        let l:cmd .= printf(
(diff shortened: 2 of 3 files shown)


gabrielbdsantos:
Well, it is not that the current implementation does not work. For me, the motivation is somewhat different: the possibility to combine inverse search with other commands.

For example, I do a lot of folding in my documents, nesting up to two or three levels. I believe this is a common practice among (Neo)Vim users, especially when dealing with long files. It helps to keep things clean and organized. Anyway, when doing inverse search, I have to manually open _all_ the folds. So, a custom command could also automatically open the folds. It would be something like:

```vim
let g:vimtex_view_zathura_synctex_command =
      \ '"nvim --server ' . v:servername .
      \ ' --remote-send '':e %{input} | :%{line} | :call feedkeys(\"zv\")<CR>''"'
```

This could also be done by simply appending extra arguments to the current implementation. However, in this case, I thought it would make more sense to have an entirely user-defined command instead of extra arguments.

lervag:
> For me, the motivation is somewhat different: the possibility to combine inverse search with other commands.

You already can, see `:help VimtexEventViewReverse`. I do more or less the same as you with regard to folding. I use:

```vim
augroup init_vimtex
  autocmd!
  autocmd User VimtexEventViewReverse normal! zMzvzz
augroup END
```

And, considering, as you say:

> Well, it is not that the current implementation does not work.

I don't think this suggested feature is necessary.

gabrielbdsantos:
Woow, I've completely missed this one feature. Sorry for that. And thanks for the tip :)

lervag:
No problem; there's a lot of stuff, so easy it is easy to miss. Glad to help!

Issue #2415, opened 2022-06-17
(view) `b:vimtex.viewer.xwin_id` corresponds to unexpected window when multiple PDFs are open


ejmastnak:
### Description

In an admittedly unusual scenario (described below in the steps to reproduce) the variable `b:vimtex.viewer.xwin_id` is set to a window that is different from the window holding the PDF corresponding to the LaTeX file edited with VimTeX.

Not a big issue, but it is inconvenient when automatically closing windows using the `VimtexEventQuit` autocommand and `CloseViewers()` function workflow given as an example in `:help vimtex-events`---in certain cases the `CloseViewers()` function, which uses `b:vimtex.viewer.xwin_id`, will close the "wrong" PDF, i.e. not the PDF associated with the edited LaTeX file.

This workflow arises when I first open a third-party PDF (e.g. a paper, textbook, etc...), then open a LaTeX file in which I take notes about the content of the third-party PDF. Exiting VimTeX will then close the third-party PDF, instead of the PDF associated with my notes.

#### Files

`minimal.tex`
```tex
\documentclass{minimal}
\begin{document}
Hello world!
\end{document}
```

`minimal.vim`
```vim
set nocompatible
let &runtimepath  = './vimtex,' . &runtimepath
let &runtimepath .= ',./vimtex/after'
filetype plugin indent on
syntax enable

" Relevant options and VimTeX configuration below.
let g:vimtex_view_method = 'zathura'

" Close viewers when VimTeX buffers are closed (see :help vimtex-events)
function! CloseViewers()
  if executable('xdotool')
        \ && exists('b:vimtex.viewer.xwin_id')
        \ && b:vimtex.viewer.xwin_id > 0
    call system('xdotool windowclose '. b:vimtex.viewer.xwin_id)
  endif
endfunction

augroup vimtex_event_close
  au!
  au User VimtexEventQuit call CloseViewers()
augroup END

" Only for personal convenience when testing
let mapleader = " "
nmap <leader>i <plug>(vimtex-info)
nmap <leader>v <plug>(vimtex-view)
nmap <leader>c <plug>(vimtex-compile)
```

#### Possible change

(I thought I would get your opinion before diving headfirst into PRs, since you know the codebase much better and will probably have a better idea than I would.)

First, it seems the problem is only relevant to `mupdf` and `zathura`, since only those viewers have a `xwin_id` property.

[...]

PR #2417, opened 2022-06-20
[Viewer] Make search for X11 window IDs more robust

ejmastnak:
This PR addresses and solves #2415, wherein the variable `b:vimtex.viewer.xwin_id` would sometimes be set to an X11 window ID that was not the ID of the actual viewer. The issue is solved on my end (Arch Linux 5.18.2 with Zathura), but it would be prudent to test on other systems.

The PR is a refactor of existing code more than anything else; here is a quick summary:

- Move winID-searching logic (by viewer PID, name and class) to three dedicated functions (to avoid code repetition; see [5671b1a](https://github.com/lervag/vimtex/commit/5671b1aed1628ea9ff47f966eb3b573b0705ca01) for details)
- Makes the `xdo_get_id` function in `_template.vim` attempt search by PID and search by window name before attempting search by class (previously only search by class was used).

More details in the commit messages.

refer: #2415

### DIFF
--- a/autoload/vimtex/view/_template.vim
+++ b/autoload/vimtex/view/_template.vim
@@ -132,28 +132,17 @@ endfunction
 function! s:viewer.xdo_get_id() dict abort " {{{1
   if !self.xdo_check() | return 0 | endif
 
-  if self.xwin_id <= 0
-    " Allow some time for the viewer to start properly
-    sleep 500m
+  if self.xwin_id > 0 | return self.xwin_id | endif
 
-    " Note: Since `xdo_find_win_id_by_{name,pid,class}()` return zero if no
-    " window ID is found, they can be used to directly set `self.xwin_id`
-    " without introducing and if/else testing intermediate variables like
-    " `l:win_id_by_{name,pid,class}`.
+  " Allow some time for the viewer to start properly
+  sleep 500m
 
-    " First try search by viewer's PID
-    let self.xwin_id = self.xdo_find_win_id_by_pid()
-
-    " If search by PID failed, try search by viewer's window name
-    if self.xwin_id <= 0
-      let self.xwin_id = self.xdo_find_win_id_by_name()
-    endif
-
-    " Fallback to search by viewer class
-    if self.xwin_id <= 0
-      let self.xwin_id = self.xdo_find_win_id_by_class()
-    endif
-  endif
+  " Try finding viewer's window ID, first attempting search by PID, then by
+  " window name, and finally, as a fallback, by window class.
+  for l:method in ['pid', 'name', 'class']
+    execute "let self.xwin_id = self.xdo_find_win_id_by_" . l:method . "()"
+    if self.xwin_id > 0 | break | endif
+  endfor
 
   if self.xwin_id <= 0
     call vimtex#log#warning('Viewer cannot find ' . self.name . ' window ID!')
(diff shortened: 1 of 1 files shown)


autoload/vimtex/view/_template.vim
```diff
@@ -136,23 +136,37 @@ function! s:viewer.xdo_get_id() dict abort " {{{1
     " Allow some time for the viewer to start properly
```
lervag, line 136: Let's simplify further. Something like this:

```vim
if self.xwin_id > 0 | return self.xwin_id | endif

" Allow some time for the viewer to start properly
sleep 500m

" EXPLANATORY COMMENT HERE
for l:method in ['name', 'pid', 'class']
  let self.xwin_id = self.xdo_find_win_id_by_{l:method}()
  if self.xwin_id > 0 | break | endif
endfor

if self.xwin_id <= 0
  call vimtex#log#warning('Viewer cannot find ' . self.name . ' window ID!')
endif

return self.xwin_id
```

autoload/vimtex/view/_template.vim
```diff
@@ -162,31 +176,67 @@ function! s:viewer.xdo_exists() dict abort " {{{1
...
+    if self.xwin_id > 0
+      let self.xwin_id = self.xdo_find_win_id_by_name()
     endif
   endif
 
   return self.xwin_id > 0
 endfunction
 
+" }}}1
+" Attempts to find the viewer's X window ID using `xdotool` search by window
```
lervag, line 193: I prefer to have the comments at the top of the function, but not above.

autoload/vimtex/view/_template.vim
```diff
@@ -162,31 +176,67 @@ function! s:viewer.xdo_exists() dict abort " {{{1
...
+" class. Returns the viewer's window ID if one is found, and `0` otherwise.
+function! s:viewer.xdo_find_win_id_by_class() dict abort " {{{1
+  let l:xwin_ids = vimtex#jobs#capture('xdotool search --class ' . self.name)
+  if len(l:xwin_ids) == 0
```
lervag, line 197: I would do it like this:

```
return empty(l:xwin_ids) ? 0 : l:xwin_ids[-1]
```

autoload/vimtex/view/_template.vim
```diff
@@ -162,31 +176,67 @@ function! s:viewer.xdo_exists() dict abort " {{{1
...
+  return 0
+endfunction
+
+" }}}1
+" Attempts to find the viewer's X window ID using `xdotool` search by the
+" viewer's process ID. Returns the viewer's window ID if one is found, and `0`
+" otherwise.
+function! s:viewer.xdo_find_win_id_by_pid() dict abort " {{{1
+  let l:pid = has_key(self, 'get_pid') ? self.get_pid() : 0
+  if l:pid > 0
```
lervag, line 230: I would simplify like this:

```vim
if l:pid <= 0 | return 0 | endif

let xwin_ids = vimtex#jobs#capture(
      \   'xdotool search --all --pid ' . l:pid
      \ . ' --name ' . fnamemodify(self.out(), ':t'))
return get(xwin_ids, 0)
```

lervag:
Thanks! I think this looks very good and is almost ready for merge already. I only had a few minor comments and suggestions.

autoload/vimtex/view/_template.vim
ejmastnak: Very elegant! Note that the code...
```vim
let self.xwin_id = self.xdo_find_win_id_by_{l:method}()
```
...give me the error `E716: Key not present in Dictionary: "xdo_find_win_id_by_{l:method}()"`.

I've taken the liberty of replacing it with
```vim
execute "let self.xwin_id = self.xdo_find_win_id_by_" . l:method . "()"
```
which I believe does the trick.

autoload/vimtex/view/_template.vim
ejmastnak: Nice, I like the C-style conditional `?:`.
Will update.

lervag:
Thanks! I made a few adjustments on top, I hope you don't mind. Let me know if I accidentily broke something :p

ejmastnak:
> Thanks! I made a few adjustments on top, I hope you don't mind. Let me know if I accidentily broke something :p

Looks good! Just tested locally and everything seems to work. Thanks for your suggestions and help implementing this!

PR #2436, opened 2022-07-13
Fix outdir quoting issue for tectonic compiler

thezeroalpha:
For the tectonic compiler: if the path to the file contains spaces, outdir is not quoted properly and hence the build command does not work. This commit fixes that by escaping the path with `fnameescape()`.

---

Minimal example of the fixed issue:

1. Create the directory and file:

```
mkdir ~/'dir with spaces'
touch ~/'dir with spaces'/main.tex
```

2. Edit `main.tex` to create a valid document.
3. Set `let g:vimtex_compiler_method='tectonic'`
4. Try to compile, observe an error. `:VimtexCompileOutput` shows that `'spaces'` in the path is treated as a separate argument, though it should not be.


lervag:
Thanks!

PR #2440, opened 2022-07-15
Add `textcite` and `citeauthor` to the `texCmdRef` syntax item

thezeroalpha:
These commands are often used with Biblatex, so in my opinion they should be a part of the `texCmdRef` for correct highlighting (and avoiding spell-checking of the cite key).
It might be useful to also add other [citation commands from Biblatex](https://mirrors.nic.cz/tex-archive/macros/latex/contrib/biblatex/doc/biblatex.pdf); if you agree, I can do that.
Also, I'm not sure if it's better to have separate `syn match` definitions (i.e. one for `citeauthor` and one for `textcite`), or both in a regular expression in a single `syn match` (ie.g. `\\\(text\)\?cite\(author\)\?\>`). Multiple `syn match` definitions are more readable, but there might be a performance difference.


lervag:
Thanks for noticing a problem and trying to solve it; the effort is appreciated!

However, this is actually not necessary. `\textcite` and `\citeauthor` are already supported via the [biblatex syntax package](https://github.com/lervag/vimtex/blob/e56daa62a38160fbba45aca9f1943c14cefe4876/autoload/vimtex/syntax/p/biblatex.vim#L14-L18). The package is only loaded when biblatex is recognized, though, which may be why you have not noticed that this already works.

Please see `:help vimtex-syntax-package` for more info, and feel free to ask if something is unclear!

PR #2461, opened 2022-08-17
Allow inverse search to work even when VimTex is loaded only when filetype=tex

dylandronnier:
When VimTex is loaded only when filetype=tex, the editor command VimtexInverseSearch is not recognized.

Specifying the filetype when calling vim/nvim solve this issue.

### DIFF
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -89,7 +89,7 @@ function! s:cmdline(outfile, synctex, start) abort " {{{1
     let l:cmd .= ' ' . g:vimtex_view_zathura_options
     if a:synctex
       let l:cmd .= printf(
-            \ ' -x "%s -c \"VimtexInverseSearch %%{line} ''%%{input}''\""',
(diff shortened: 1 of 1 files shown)


lervag:
No, sorry. You are loading VimTeX wrongly. VimTeX should _not_ be put behind a filetype guard. Can you show me how you install VimTeX?

dylandronnier:
Hello lervag and thank you for this amazing plugin. I installed VimTex using the Packer plugin. I load vimtex using the ft option

`use { 'lervag/vimtex', ft  = 'tex'}`

lervag:
> Hello lervag and thank you for this amazing plugin.

Glad you like it! :)

> I installed VimTex using the Packer plugin. I load vimtex using the ft option
> 
> `use { 'lervag/vimtex', ft = 'tex'}`

There's your problem. Don't use `ft = 'tex'` - VimTeX is a filetype plugin and will already lazily load as needed. By adding `ft = 'tex'`, you are preventing it from loading correctly, which explains why you needed this PR. Change to

```
use 'lervag/vimtex'
```

PR #2495, opened 2022-09-22
register the (arg)?mini[!*e]?, (arg)?maxi[!*]? environment from  to region_math

huweiATgithub:
The package [optidef](https://ctan.org/pkg/optidef) provides mini, maxi, argmini, argmaxi environments that should be considered math environments.

### DIFF
--- a/autoload/vimtex/syntax/p/optidef.vim
+++ b/autoload/vimtex/syntax/p/optidef.vim
@@ -1,0 +1,16 @@
+" VimTeX - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+"
+
+scriptencoding utf-8
+
+function! vimtex#syntax#p#optidef#load(cfg) abort " {{{1
+  call vimtex#syntax#core#new_region_math('\(arg\)\?mini')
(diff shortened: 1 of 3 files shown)


autoload/vimtex/syntax/p/optidef.vim
```diff
@@ -0,0 +1,16 @@
+" VimTeX - LaTeX plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+"
+
+scriptencoding utf-8
```
lervag, line 7: This is not necessary - it is only necessary when a scriptfile has special utf8 symbols.

lervag (commented):
Thanks! This looks good and I don't see any immediate problems here. But I would kindly ask you to also add an example within `test/test-syntax`. Create two files, `test-optidef.vim` and `test-optidef.tex`. For this simple stuff you don't need to create any real tests, but having the example there makes it much easier for further development or bug fixing in the future in case something breaks.

autoload/vimtex/syntax/p/optidef.vim
huweiATgithub: sure!

huweiATgithub:
I added some simple tests.

lervag:
Thank you!

PR #2528, opened 2022-10-24
Use json to store fontawesome icons

Freed-Wu:
Add a github workflow to check if a json is sorted


autoload/vimtex/syntax/p/fontawesome5.vim
```diff
@@ -4,7 +4,9 @@
 " Email:      karl.yngve@gmail.com
 "
 
-scriptencoding utf-8
+let s:path = fnamemodify(resolve(expand('<sfile>:p')), ':h:h:h:h:h')
+let s:json_path = s:path . '/assets/json/fontawesome.json'
+let s:fontawesome = json_decode(join(readfile(s:json_path), ''))
```
lervag, line 9: I prefer to have the script local stuff at the bottom and the main "public" code at the top.

autoload/vimtex/syntax/p/fontawesome5.vim
```diff
@@ -16,7 +18,14 @@ function! vimtex#syntax#p#fontawesome5#load(cfg) abort " {{{1
         \ 'opts': 'conceal containedin=texCmdFontawesome',
         \})
 
-  for [l:name, l:symbol] in s:fontawesome
+  for [l:name, l:symbol] in items(s:fontawesome)
+    " 500px does not have the \faName variant
+    if l:name ==# '500px'
```
lervag, line 23: Yes, ok. But this will add a conditional on every iteration. It is small, but still, there are O(1000) symbols, so it may still be significant. Should at least check if it is relevant.

lervag (changes requested):
Thanks! I'll have to check this before accepting. You can also check. I would actually not be surprised if this now fails at `\faFile*`.

Can you explain the github workflow for me? I'm not an expert on that, and I'm curious if we should make it more specific? I.e., now the workflow triggers on all json files, but that does not seem correct if there should be more json files added?

Freed-Wu:
> Can you explain the github workflow for me

```sh
          for json in assets/json/*.json ; do
            diff <(jq -S . $json) $json || exit 1
          done
```

if a json is not sorted, jq -S . will sort it and print the sorted json to 
stdout, so diff the stdout and original json will return 1, `|| exit 1` will 
quit the for-loop, Like this:

[screenshot](https://user-images.githubusercontent.com/32936898/197423696-c44a5f2b-45d5-4c40-9d85-1191f7e786c5.png)

autoload/vimtex/syntax/p/fontawesome5.vim
Freed-Wu: OK, I'll fix.

autoload/vimtex/syntax/p/fontawesome5.vim
Freed-Wu: OK, I'll fix it.

lervag:
Thanks. As I wrote above:

> I'll have to check this before accepting. You can also check. I would actually not be surprised if this now fails at `\faFile*`.

I've checked, and as you see, it now fails:

image

The reason is that the Vim dictionary does not preserve a sorted order, necessarily. So, this means we must either sort the `item(DICTIONARY)` or store the data as a sorted list in the json file.

Regarding the workflow: I'm still wondering if it is a good idea to trigger this workflow on _any_ json file and to specify that all json files must be sorted? Would it not be more appropriate to restrict it to this specific file?

Freed-Wu:
> Would it not be more appropriate to restrict it to this specific file?

Fixed.

> store the data as a sorted list in the json file.

Fixed.

lervag:
I did some minor profiling, and fascinatingly, doing the `json_decode(fileread())` is faster than inlining in vimscript.

I'm merging this now. Thanks for contributing!

lervag:
I made a few minor changes to the workflow (https://github.com/lervag/vimtex/blob/master/.github/workflows/assets.yml), mostly that I fixed the path to the fontawesome file. That will avoid any surprises if I were to add more json files to the assets directory. I believe I should probably collect all assets within that subdirectory, so I might look into that in the not so far away future.

lervag:
Hmm. For some reason, the updated workflow does not run properly now:

https://github.com/lervag/vimtex/actions/runs/3332974138

Perhaps you could assist me here? Here's the diff of the workflow file from your version to the current:

image

Freed-Wu:
Fixed :+1:

lervag:
Thanks! I'll look into the new PRs later today!

PR #2531, opened 2022-10-27
Fix wrong .github/workflows/assets.yml

Freed-Wu:
github should be env. :bug:

And `# yamllint disable-line rule:truthy` just let yamllint ignore `on` because
`on` in yaml equal to true, which the latter is preferred. If you don't like it,
you can delete it :smile:

Screenshot from 2022-10-27 10-30-05


lervag:
Thanks!

PR #2537, opened 2022-10-29
replaces nvim-config-local with exrc.nvim

krishnakumarg1984:
nvim-config-local is unmaintained, and has known bugs. In the last few months, I have successfully managed to switch to `exrc.nvim`, a lua-only plugin for neovim.


lervag:
Thanks

PR #2548, opened 2022-10-30
Replace texlab url with github link

krishnakumarg1984:
When trying to navigate to the original texlab netlify URL in the browser, it redirects to the github page for the project.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -759,7 +759,7 @@ Language servers~
   conjunction with VimTeX as an alternative to |vimtex-grammar|.
 
   [0]: https://langserver.org/
-  [1]: https://texlab.netlify.app/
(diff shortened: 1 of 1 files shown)


clason:
Yes. That is intentional.

krishnakumarg1984:
Yup. So, maybe let's just provide people the github url directly in the documentation?

lervag:
I don't mind accepting this. @clason Is there a reason _not_ to use the github url here? E.g., perhaps a more dedicated page is expected or planned and the current URL is a placeholder for that?

clason:
I don't care either way, but the "official" URL links to the github page by design, since there's no (need) for a separate webpage.

lervag:
Thanks @krishnakumarg1984, I see no good reason not to accept the PR.

krishnakumarg1984:
Thanks!

PR #2551, opened 2022-10-30
 Improvement of the calculation of the usable width of current window fixes #2545

### LINKED ISSUE #2545: Cosmetic issues with formatting of the text displayed for a closed fold
### Description

The formatting of the text displayed for a closed fold is based on an incorrect estimation of the full width of the current window. This issue is actually two-folds.
1) If there is no signcolumn, then the full width is underestimated of 1. In the case of a folded labelled environment, this results in an unaethestic '·' at the end of the text displayed for the closed fold.
2) If there is a signcolumn, its width is not taken into account in the estimation of the full width of the current window. If the width is larger than 1, this results in the fact that the label of a folded labelled environment is not entirely displayed.

### Steps to reproduce

[minimal.vim.zip](https://github.com/lervag/vimtex/files/9895962/minimal.vim.zip)
[minimal.tex.zip](https://github.com/lervag/vimtex/files/9895965/minimal.tex.zip)

1. `nvim -u minimal.vim ~/Desktop/minimal.tex`
2. The equation environment should be folded and a '·' is visible at the end of the text displayed for the closed fold.
3. Type `:set scl=yes:5`
4. The label of the folded equation is no longer entirely visible.

### Expected behavior

The closing parentheses of the label of a folded labelled environment should be the last character of the line in any case.

### Actual behavior

_No response_

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
System info:
  OS: macOS 13.0 (22A380)
  Vim version: NVIM v0.8.0
  Has clientserver: true
  Servername: /tmp/nvimsocket

VimTeX project: minimal
  base: minimal.tex
  root: /Users/patrick.ballard/Desktop
  tex: /Users/patrick.ballard/Desktop/minimal.tex
  document class: minimal
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: General
  qf method: LaTeX logfile
```

PatrBal:
I believe that this pull request provides a fix for issue #2545. It is strongly based on https://stackoverflow.com/questions/26315925/get-usable-window-width-in-vim-script.

### DIFF
--- a/autoload/vimtex/fold/envs.vim
+++ b/autoload/vimtex/fold/envs.vim
@@ -81,9 +81,42 @@ function! s:folder.text(line, level) abort dict " {{{1
     let caption = self.parse_caption(a:line)
   endif
 
+  " let width = winwidth(0)
+  "       \ - (&number ? &numberwidth : 0)
+  "       \ - str2nr(matchstr(&foldcolumn, '\d\+$'))
+
   let width = winwidth(0)
-        \ - (&number ? &numberwidth : 0)
-        \ - str2nr(matchstr(&foldcolumn, '\d\+$'))
+  let numberwidth = max([&numberwidth, strlen(line('$')) + 1])
+  let numwidth = (&number || &relativenumber) ? numberwidth : 0
+  let foldwidth = str2nr(matchstr(&foldcolumn, '\d\+$'))
+
+  if &signcolumn == 'yes'
+    let signwidth = 2
+  elseif &signcolumn =~ 'yes'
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! I refactored the code into a separate function and I made some simplifications. I believe it should still be accurate, although I updated the logic of the `auto` case somewhat.

Issue #2471, opened 2022-08-27
Compilation error due to failed mainfile detection!  for standalone files


Rmano:
### Description

After the last upgrade (of `vimtex` as well as of the operating system), I have often the following error on simple files:

![image](https://user-images.githubusercontent.com/6414907/187039299-35910e49-b748-4d6c-b027-6cc20540aad6.png)

this is with `lalla.tex`

```latex
\documentclass[border=10pt]{standalone}
\usepackage[siunitx, RPvoltages]{circuitikz}
  \begin{document}
  \begin{tikzpicture}[]
      \draw[red] (0,0) to[R] ++(2,0);
  \end{tikzpicture}
\end{document}
```
opened with `vim -u ~/software/bugs/vimtex/minivimrc lalla.tex` where the `minivimrc` is:

```vim
set nocompatible
let &rtp  = '~/.vim/plugged/vimtex,' . &rtp
let &rtp .= ',~/.vim/plugged/vimtex/after'
filetype plugin indent on
syntax enable
```

and `~/.latexmkrc` 

```
$pdflatex = 'pdflatex -synctex=1 %O %S';
$pdf_update_method = 0;
$dvi_mode = 0;
$pdf_mode = 1;
$preview_continuous_mode = 1;
$pdf_previewer = 'start evince';

push @generated_exts, "cb";
push @generated_exts, "cb2";
push @generated_exts, "spl";
push @generated_exts, "nav";
push @generated_exts, "snm";
push @generated_exts, "tdo";
push @generated_exts, "nmo";
push @generated_exts, "brf";
push @generated_exts, "nlg";
push @generated_exts, "nlo";
push @generated_exts, "nls";
push @generated_exts, "synctex.gz";
push @generated_exts, "tex.latexmain";
push @generated_exts, "run.xml";
```

That seems to happen only when I try to compile files with `standalone` class...

```yaml
System info:
  OS: Ubuntu 22.04.1 LTS
  Vim version: VIM 8.2 (1-3995)
  Has clientserver: true
  Servername: undefined (vim started without --servername)

VimTeX project: lalla
  base: lalla.tex
  root: /home/romano/tmp
  tex: /home/romano/tmp/lalla.tex
  main parser: fallback current file
  document class: standalone
  packages: circuitikz
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
[...]

PR #2556, opened 2022-11-03
TeX root directive needs reloading after change

fauust:
See: https://github.com/lervag/vimtex/issues/2471


lervag:
I think you are pointing out something real, but the suggested update is not quite correct. You don't need to restart, you can use `:VimtexReload`. Also, it should be clarified that you need to reload or restart whenever you _change_ the tex root directive.

fauust:
Ok, I have just improved the doc. I am not a native English speaker so don't hesitate to suggest the exact wording ;)

lervag:
Thanks! I tried to further improve slightly, let me know what you think!

Issue #2557, opened 2022-11-04
main file detection strictness


shuber2:
### Description

**Issue.**
Commit e5de00e2 adds more strict mainfile detection and with commit fa5cbe4f an error is emitted when mainfile detection fails.

I have a use case where I produce lecture note slides; each lecture unit is a beamer presentation. The preamble for each `lecture-xx.tex` is almost the same and hence hence each file starts with an `\input{slides.inc}` to reduce code duplication. In particular, the \documentclass line is contained in the `slides.inc`.

This now breaks the mainfile detection, because a \documentclass line became necessary. (Actually, having a \documentclass is not sufficient: It cannot be certain documentclass variants and the document-environment is also necessary.)

**Proposal.**
I would like to propose two things:

1. Add the required properties for the mainfile detection logic into the documentation (`:help`). I reverse engineered the properties through git bisect and the corresponding vimscript code.
2. Maybe rethink whether the existence of `\begin{document}…\end{document}` would not be a sufficient property for a file to be a mainfile.

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
System info:
  OS: Debian GNU/Linux 11 (bullseye)
  Vim version: NVIM v0.7.0
  Has clientserver: true
  Servername: /tmp/nvimqvrkzB/0

VimTeX project: foo
  base: foo.tex
  root: /tmp
  tex: /tmp/foo.tex
  main parser: fallback current file
  document class: 
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: General
  qf method: LaTeX logfile
```

lervag:
> **Issue.** Commit [e5de00e](https://github.com/lervag/vimtex/commit/e5de00e2e7577fca3b39d533732de259c8a0abf6) adds more strict mainfile detection and with commit [fa5cbe4](https://github.com/lervag/vimtex/commit/fa5cbe4fab0e9b7c62ee318a158297a529c6d7b5) an error is emitted when mainfile detection fails.
> ...
> This now breaks the mainfile detection, because a \documentclass line became necessary. (Actually, having a \documentclass is not sufficient: It cannot be certain documentclass variants and the document-environment is also necessary.)

[...]

Issue #2483, opened 2022-09-09
Vimtex can't recognize the main.tex file


Yorkwn:
### Description

For some reason the last 2 weeks or so i've been unable to compile the main file of my latex project. I've explained the general issue also in the neovim subreddit so i won't repeat what i explained there. On the minimal directory,

main.tex includes:
```tex
\input{preamble.tex}

\begin{document}
Hello World
\end{document}
```

and preamble.tex:

```tex
\documentclass[11pt]{article}
\usepackage[utf8]{inputenc}
```

### Steps to reproduce

1. `vim minimal.vim` same as the example file but with plugged/ instead of bundle/
2. `mkdir Documents/minimal`
3. `cd Documents/minimal/`
4. `touch main.tex`
5. `touch preamble.tex`
6. `vim -u ~/minimal.vim main.tex`

### Expected behavior

The expected behavior would be to recognize the main.tex file and compile the document.

### Actual behavior

Instead i get this error:

```
VimTeX: Compilation error due to failed mainfile detection!
Please ensure that VimTeX can locate the proper main .tex file.
Read ":help vimtex-multi-file" for more info.
```

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info:
  OS: Fedora release 36 (Thirty Six)
  Vim version: VIM 9.0 (1-348)
  Has clientserver: false

VimTeX project: main
  base: main.tex
  root: /home/gk/Documents/minimal
  tex: /home/gk/Documents/minimal/main.tex
  main parser: fallback current file
  document class: article
  packages: inputenc
  source files:
    main.tex
    preamble.tex
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: General
  qf method: LaTeX logfile
```

Rmano:
This seems the new main file detection code; he does not find a `\documentclass` do it thinks it's not the main file. 

If you add:

```
%! TeX root = main.tex
```

at the top of your `main.tex` file. and exit and re-enter `vim`,   it should work (it does here).

lervag:
> For some reason the last 2 weeks or so i've been unable to compile the main file of my latex project. I've explained the general issue also in the neovim subreddit so i won't repeat what i explained there. On the minimal directory, …

[...]

Issue #2465, opened 2022-08-20
Root detection fails when file contains the strings `\documentclass[` and `\begin{document}` even if they are commented


rrueger:
**_Edit: The issue is incorrectly described in this post. [See below ](https://github.com/lervag/vimtex/issues/2465#issuecomment-1221418204)for correct description._**

I have a unified preamble that I use for many projects at `$LATEXTEMPLATES/preamble.tex`.

I initialise new projects as follows

    mkdir project
    cp $LATEXTEMPLATES/main.tex project/project-name.tex
    ln -s $LATEXTEMPLATES/preamble.tex project/preamble.tex
    # edit project/project-name.tex with project specific content

It is important that `preamble.tex` is a symbolic link here to observe the unwanted behaviour.
Obviously `project/project-name.tex` should not be a symbolic link, because the content is different for each project.

(I use a symbolic link for `preamble.tex` so that any changes that I make to the `preamble.tex` in any individual project will be observed globally. For my workflow, this makes sense.)

At the moment, when I edit `preamble.tex` VimTeX tries to compile `preamble.tex` as a standalone file.
The auto-root-detection seems to fail.

I would expect one of two things to happen, when I edit `project/preamble.tex -> $LATEXTEMPLATES/preamble.tex`

1. VimTeX follows the symbolic link, and sees that `$LATEXTEMPLATES/main.tex` imports `$LATEXTEMPLATES/preamble.tex` so it successfully compiles `$LATEXTEMPLATES/main.tex`.
2. VimTeX doesn't follow the symbolic link, and sees that `project/project-name.tex` imports `project/preamble.tex` so it successfully compiles `project/main.tex`.

Since `project/project-name.tex` is different for each project, I cannot add a vimtex-tex `ROOT` directive to `preamble.tex`. On the other hand, it seems like kind of a hack to add `$LATEXTEMPLATEX/main.tex` as the `ROOT` in the unified preamble.

I think there are two options for fixing this unwanted behaviour:

1.  "Fixing" the root detection, so that VimTeX **doesn't** follow the symbolic link, and realises that `project/project-name.tex` imports `project/preamble.tex`. I think it is important that VimTeX doesn't follow the symbolic link, because it (imho) be an entirely legal and sensible approach to keep a unified preamble without a corresponding `main.tex` that imports it.
2. Adding a `VIMTEX_DISABLE` directive, to disable VimTeX for this specific file.

I hope this makes sense!
Perhaps I have missed something in the documentation?

[...]

PR #2558, opened 2022-11-04
Improve the main-file detection

lervag:
As a result of #2465, I added a restriction to the compiler module where it won't compile the document unless the main file was _verified_ to be a valid main file. This helped to ensure robustness, at the cost of breaking the work flow for several users.

This PR introduces a minor change to how VimTeX checks if a file is a main file that should resolve most of the issues reported after the more restrictive compiler module. However, I won't merge it immediately, because it _may_ come at a cost of reduced init performance. That is, slower load time for VimTeX. This should only affect large projects, and only if they consist of many/multiple files. But still, it may be a relevant blocker.

Therefore, before I merge this, I would be very happy if users who have large LaTeX documents with multiple files (e.g. books, large theses, etc) could test if this branch introduce a noticable reduced startup time. Please also report if you do not see a reduced performance.

Related issues:
* #2557
* #2471
* #2483

People who may have an interest in this issue: @Rmano, @fauust, @Yorkwn, @shuber2, @krishnakumarg1984, @clason

### DIFF
--- a/autoload/vimtex/state.vim
+++ b/autoload/vimtex/state.vim
@@ -475,23 +475,23 @@ endfunction
 function! s:file_is_main(file) abort " {{{1
   if !filereadable(a:file) | return 0 | endif
 
+  let l:preamble = vimtex#parser#preamble(a:file, {
+        \ 'inclusive' : 1,
+        \ 'root' : fnamemodify(a:file, ':p:h'),
+        \})
+
   " Check if a:file is a main file by looking for the \documentclass command,
   " but ignore the following:
-  "
-  "   \documentclass[...]{subfiles}
-  "   \documentclass[...]{standalone}
-  "
-  let l:lines = readfile(a:file, 0, 50)
+  " * \documentclass[...]{subfiles}
+  " * \documentclass[...]{standalone}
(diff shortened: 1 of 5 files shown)


shuber2:
Regarding startup time, I tested startup time on three larger projects:

- A 100+ pages research center proposal
- A 140+ pages book
- A 100+ pages lecture notes

They all loaded quickly. Only a 3800+ lines tex file of the research proposal took some two seconds to load. But that was also true for the current master branch.

krishnakumarg1984:
I see virtually no difference when typesetting my (admittedly too long) 329 page PhD thesis. The new commits don't seem to have any impact either way.

Rmano:
I checked on my biggest file (the manual for `circuitikz`, which is a single-file project tough) and I noticed no difference at all.

clason:
No measurable difference for me (using `hyperfine`), so 👍

ejmastnak:
+1 No appreciable difference for me between the feature branch and master on a ~300 page book with multiple chapters in separate files

lervag:
Thank you all for the quick feedback! I see no reason to merge this, then.

PR #2560, opened 2022-11-05
doc: Note on lazy loading and inverse search

shuber2:
The provided command lines for inverse search do not work when VimTeX is loaded lazily only for tex files, as provided by various plugin managers. Add a note on how to cope with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ Or use some other plugin manager:
 * [neobundle](https://github.com/Shougo/neobundle.vim)
 * [pathogen](https://github.com/tpope/vim-pathogen)
 
+Note that some plugin managers provide mechanisms for lazy loading of plugins,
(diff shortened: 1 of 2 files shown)


lervag:
I don't agree with this, because there is simply _no_ reason to lazily load VimTeX. It is already lazily loaded by virtue of being a filetype plugin. People should _not_ cope with it as you suggest, they should just _not_ lazily load VimTeX.

However, it may be worth mentioning this somewhere, because this particular issue does show up once in a while. Perhaps we should add a note in the [README](https://github.com/lervag/vimtex/blob/8824c2415c60f0ff215a034385237ce2d5874117/README.md#installation)?

shuber2:
That is a good and valid point. Thank you for pointing it out. I changed the proposed pull request to adding notes on discouragement of lazy loading VimTeX.

lervag:
Thanks! I took the liberty of rephrasing it while merging.

Issue #2569, opened 2022-11-15
Inverse search with Skim and Neovim open XQuartz


PatrBal:
### Description

I use Vimtex with Neovim in iTerm2 on macOS with Skim as the viewer. The reverse search in Skim is configured as suggested in the help file:

`nvim --headless -c "VimtexInverseSearch %l '%f'"`

When I perform a reverse search from Skim. It opens XQuartz systematically. This is annoying because it makes the first reverse search very slow (it is only performed after XQuartz has started). The culprit is here:

https://github.com/lervag/vimtex/blob/e3aa3ccc434321fdd0cd0c036a74d40214e34da4/autoload/vimtex/view.vim#L118

Whenever I run `call b:vimtex.viewer.xdo_focus_vim()` as an ex command from Neovim, it starts XQuartz. I have commented out this line and it solves the problem for me.  Reverse search from Skim works flawlessly and XQuartz is no longer started. I have not dug to find the aim this function, but I believe that it should not be called when the `g:vimtex_view_method` is set to `'skim'`.

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
System info:
  OS: macOS 13.0 (22A380)
  Vim version: NVIM v0.8.1
  Has clientserver: true
  Servername: /tmp/nvimsocket

VimTeX project: Leray-Lions
  base: Leray-Lions.tex
  root: /Users/patrick.ballard/Documents/Science/Articles/Leray-Lions
  tex: /Users/patrick.ballard/Documents/Science/Articles/Leray-Lions/Leray-Lions.tex
  main parser: current file verified
  document class: article
  packages: amsbsy amsfonts amsgen amsmath amsopn amssymb amstext babel bbm calc epstopdf-base etoolbox fontenc geometry graphics graphicx iftex ifvtex inputenc keyval lmodern mathrsfs mathtools mhsetup microtype pgf pgfcomp-version-0-65 pgfcomp-version-1-18 pgfcore pgffor pgfkeys pgfmath pgfrcs pgfsys tikz trig xcolor
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
    job: 
      jobid: 9
      output: /var/folders/4w/nx4x9z8j3wz4mmjfvs4xlzxh0000gn/T/nvim.patrick.ballard/TzATPa/2
[...]

PR #2570, opened 2022-11-15
Auto-commit: saved _template.vim

PatrBal:
This PR aims at solving issue #2569


lervag:
Thanks; I appreciate the effort of trying to solve this, but I think this is not quite what we want. Let's continue the discussion in the issue thread.

PR #2571, opened 2022-11-16
Fix TeX directive recognition

tyckesak:
This one simple typo—probably left over from some old version—rendered TeX directives like `%! TeX program = lualatex` useless and prevented Vimtex from recognising directives at all (very annoying!)

It seems likely that this very typo is littered throughout the codebase, but I haven't had the time to establish that; for now at least, it makes TeX directives usable again.

### DIFF
--- a/autoload/vimtex/state/class.vim
+++ b/autoload/vimtex/state/class.vim
@@ -19,9 +19,7 @@ function! vimtex#state#class#new(main, main_parser, preserve_root) abort " {{{1
 
   let l:ext = fnamemodify(a:main, ':e')
   " Recognise (La)TeX-related file extensions
-  let l:new.tex = empty(filter(['^(la)?tex$', '^dtx$', '^tikz$'],
-                             \  'l:ext =~? v:val'))
-                \ ? a:main : ''
+  let l:new.tex = l:ext =~? '^\(\(la\)\=tex\|dtx\|tikz\)$' ? a:main : ''
 
   " Get preamble for some state parsing
   let l:preamble = !empty(l:new.tex)
--- a/autoload/vimtex/state/class.vim
+++ b/autoload/vimtex/state/class.vim
@@ -18,8 +18,10 @@ function! vimtex#state#class#new(main, main_parser, preserve_root) abort " {{{1
   endif
 
   let l:ext = fnamemodify(a:main, ':e')
-  let l:new.tex = index(['tex', 'dtx', 'tikz', 'ins'], l:ext) >= 0
-        \ ? a:main : ''
+  " Recognise (La)TeX-related file extensions
+  let l:new.tex = empty(filter(['^(la)?tex$', '^dtx$', '^tikz$'],
+                             \  'l:ext =~? v:val'))
+                \ ? a:main : ''
 
   " Get preamble for some state parsing
   let l:preamble = !empty(l:new.tex)
@@ -130,7 +132,7 @@ function! s:vimtex.get_tex_program() abort dict " {{{1
   let l:tex_program_re =
         \ '\v^\c\s*\%\s*!?\s*tex\s+%(ts-)?program\s*\=\s*\zs.*\ze\s*$'
 
-  let l:lines = vimtex#parser#preamble(self.base, {'root' : self.root})[:20]
+  let l:lines = vimtex#parser#preamble(self.tex, {'root' : self.root})[:20]
   call map(l:lines, 'matchstr(v:val, l:tex_program_re)')
   call filter(l:lines, '!empty(v:val)')
   return tolower(get(l:lines, -1, '_'))
--- a/autoload/vimtex/state/class.vim
+++ b/autoload/vimtex/state/class.vim
@@ -18,8 +18,7 @@ function! vimtex#state#class#new(main, main_parser, preserve_root) abort " {{{1
   endif
 
   let l:ext = fnamemodify(a:main, ':e')
-  " Recognise (La)TeX-related file extensions
-  let l:new.tex = l:ext =~? '^\(\(la\)\=tex\|dtx\|tikz\)$' ? a:main : ''
+  let l:new.tex = l:ext =~? '\v^%(%(la)?tex|dtx|tikz|ins)$' ? a:main : ''
 
   " Get preamble for some state parsing
   let l:preamble = !empty(l:new.tex)
--- a/autoload/vimtex/state/class.vim
+++ b/autoload/vimtex/state/class.vim
@@ -130,7 +130,7 @@ function! s:vimtex.get_tex_program() abort dict " {{{1
   let l:tex_program_re =
         \ '\v^\c\s*\%\s*!?\s*tex\s+%(ts-)?program\s*\=\s*\zs.*\ze\s*$'
 
-  let l:lines = vimtex#parser#preamble(self.tex, {'root' : self.root})[:20]
+  let l:lines = vimtex#parser#preamble(self.base, {'root' : self.root})[:20]
   call map(l:lines, 'matchstr(v:val, l:tex_program_re)')
   call filter(l:lines, '!empty(v:val)')
   return tolower(get(l:lines, -1, '_'))
--- a/autoload/vimtex/doc.vim
+++ b/autoload/vimtex/doc.vim
@@ -57,7 +57,7 @@ function! vimtex#doc#make_selection(context) abort " {{{1
   endif
 
   if len(a:context.candidates) == 1
-    if g:vimtex_doc_confirm_single && vimtex#ui#confirm([
+    if !g:vimtex_doc_confirm_single || vimtex#ui#confirm([
           \ 'Open documentation for ' . a:context.type . ': ',
           \ ['VimtexSuccess', a:context.candidates[0]],
           \ '?'


lervag:
That's strange. I'm quite sure that tex directives work exactly as intended on my end. Could you provide a concrete, minimal example with a description to show that it does not work? Please include info on your system as well, as it may be relevant.

lervag:
Notice, though, that `b:vimtex.tex` is the absolute path to the main file of a given project, whereas `b:vimtex.base` is only the base path, i.e. relative to the root directory. Thus, your proposed change is likely not good without _more_ changes. E.g., if you open a file `/path/chapter/1.tex` included in a main file `/path/main.tex` and if you have `set autochdir` (some people do!), then calling `vimtex#parser#preamble(state.base, ...)` _will_ fail.

tyckesak:
> [...] a concrete, minimal example [...]?

You got it.

We are talking about a MacVim setup on macOS 13.0: typing `:version` gives
```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Sep 24 2021 14:13:28)
macOS version - x86_64
Included patches: 1-3455
Compiled by GitHub Actions
Huge version with MacVim GUI.  Features included (+) or not (-):
+acl               +file_in_path      -mouse_sysmouse    +tag_binary
+arabic            +find_in_path      +mouse_urxvt       -tag_old_static
+autocmd           +float             +mouse_xterm       -tag_any_white
+autochdir         +folding           +multi_byte        -tcl
-autoservername    -footer            +multi_lang        +termguicolors
+balloon_eval      +fork()            -mzscheme          +terminal
+balloon_eval_term +fullscreen        +netbeans_intg     +terminfo
+browse            +gettext           +num64             +termresponse
++builtin_terms    -hangul_input      +odbeditor         +textobjects
+byte_offset       +iconv             +packages          +textprop
+channel           +insert_expand     +path_extra        +timers
+cindent           +ipv6              +perl/dyn          +title
+clientserver      +job               +persistent_undo   +toolbar
+clipboard         +jumplist          +popupwin          +transparency
+cmdline_compl     +keymap            +postscript        +user_commands
+cmdline_hist      +lambda            +printer           +vartabs
+cmdline_info      +langmap           +profile           +vertsplit
+comments          +libcall           +python/dyn        +virtualedit
+conceal           +linebreak         +python3/dyn       +visual
+cryptv            +lispindent        +quickfix          +visualextra
+cscope            +listcmds          +reltime           +viminfo
+cursorbind        +localmap          +rightleft         +vreplace
+cursorshape       +lua/dyn           +ruby/dyn          +wildignore
+dialog_con_gui    +menu              +scrollbind        +wildmenu
+diff              +mksession         +signs             +windows
+digraphs          +modify_fname      +smartindent       +writebackup
+dnd               +mouse             -sodium            -X11
-ebcdic            +mouseshape        -sound             -xfontset
+emacs_tags        +mouse_dec         +spell             +xim
+eval              -mouse_gpm         +startuptime       -xpm
+ex_extra          -mouse_jsbterm     +statusline        -xsmp
+extra_search      +mouse_netterm     -sun_workshop      -xterm_clipboard
-farsi             +mouse_sgr         +syntax            -xterm_save
```
`Boom.latex`
```latex
%! TeX program = lualatex
\documentclass{article}
\directlua{ zzz = 108 }
\begin{document}
Where did that bring you? Back to me.
\end{document}
```
`~/.vimrc`
```vim
set encoding=utf-8
set nocompatible
call plug#begin('~/.vim/plugs')
Plug 'lervag/vimtex'
call plug#end()

let g:vimtex_syntax_enabled = 1 
let g:vimtex_view_method = 'skim'
```

Hit `\ll` and you will be greeted with an error message revealing that the `lualatex` directive was not taken into account—at all:
```
Rule 'pdflatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
  Boom.latex
  pdflatex
------------
------------
------------
Running 'pdflatex  -file-line-error -synctex=1 -interaction=nonstopmode -recorder  "Boom.latex"'
------------
This is pdfTeX, Version 3.141592653-2.6-1.40.24 (TeX Live 2022) (preloaded format=pdflatex)
 restricted \write18 enabled.

   [...omitted]

./Boom.latex:3: Undefined control sequence.
<recently read> \directlua 
                           
l.3 \directlua
              { zzz = 108 }

./Boom.latex:3: LaTeX Error: Missing \begin{document}.

See the LaTeX manual or LaTeX Companion for explanation.

   [...omitted]

```

I tracked the bug down to `self.tex` being the empty string `''` instead of anything useful; my gut feeling tells me there should be a reasonable file path there instead. As for how correct my fix is...I admit that it is a rather haphazard one—ignoring the fact that `self.base` might need the whole path prepended to it—and that I might have been too quick to open a pull request over this, but I'm glad I got your attention over this!

tyckesak:
The first couple lines of the Vimtex log might be interesting to have (before I forget to include it):
```
Rc files read:
  NONE
Latexmk: This is Latexmk, John Collins, 17 Mar. 2022. Version 4.77, version: 4.77.
======= Need to update make_preview_continuous for target files
Not using a previewer
------------
Running 'echo vimtex_compiler_callback_compiling'
------------
vimtex_compiler_callback_compiling
Latexmk: applying rule 'pdflatex'...
Rule 'pdflatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
  Boom.latex
  pdflatex
   ...
   ...
   ... [you got the rest]
```

lervag:
Thanks! Could you also state, explicitly, the steps you take in a terminal?

For me, if I create the specified files, then do `vim -u test.vim test.tex`, then do `\li` (or `:VimtexInfo`), I get this:

```
...
VimTeX project: test
  base: test.tex
  root: /home/lervag/test
  tex: /home/lervag/test/test.tex
  document class: article
  compiler: latexmk
    engine: -lualatex
    ...
```

tyckesak:
Done.

```bash
~ % cd Garage/Literacy/VimtexPlayground
~ % mvim Boom.latex
```

tyckesak:
And `:VimtexInfo` gives
```
System info:
  OS: macOS 13.0 (22A380)
  Vim version: VIM 8.2 (1-3455)
  Has clientserver: true
  Servername: VIM2

VimTeX project: Boom
  base: Boom.latex
  root: /Users/josip/Garage/Literacy/VimtexPlayground
  tex: 
  document class: 
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: Skim
  qf method: LaTeX logfile
```

lervag:
Ok. Does this mean that you apply the minimal configuration as your regular `.vimrc` file? Notice that it is usually more convenient to make a `test.vim` file similar to the one your provide in a separate directory, then use `mvim -u test.vim test.tex`; this way, you don't need to change your regular config.

The `:VimtexInfo` output shows the main problem here, which is that the `tex` property is empty. That should _not_ happen, and this is the issue. We need to figure out why it happens. The first place to look is here:

https://github.com/lervag/vimtex/blob/e5343cf7902021bcea5e9432b1da4c7e1528e37e/autoload/vimtex/state/class.vim#L21-L22

Considering that you've opened a PR and already know at least something about working with Vimscript, perhaps you could investigate on your side by e.g. adding `unsilent echo "DBG" a:main` above, and perhaps also checking the value of the function arguments before passing them to the `#new` function?

tyckesak:
> Does this mean that you apply the minimal configuration as your regular .vimrc file?

Not at all; I use a more extensive one.

> [...] then use `mvim -u test.vim test.tex` [...]

> perhaps you could investigate on your side by e.g. adding `unsilent echo "DBG" a:main` above

Good pointers, thanks! I thought for sure no one could possibly be debugging Vimscript like I have for this; I'll look into it. If you don't hear by Sunday evening CET, I might have forgot or put this off.

lervag:
Sounds good! I would not be very surprised if the problem is related to your configuration somehow, but I'm not sure. It may be a bug, and I'll be happy to help figure things out!

tyckesak:
I have put fixing this off for a week; turned out that your code reference was exactly what needed fixing! The old “bugfix” is no more, and replaced with another. This time however, we don't need to worry about breakage anymore.

autoload/vimtex/state/class.vim
```diff
@@ -18,8 +18,8 @@ function! vimtex#state#class#new(main, main_parser, preserve_root) abort " {{{1
   endif
 
   let l:ext = fnamemodify(a:main, ':e')
-  let l:new.tex = index(['tex', 'dtx', 'tikz', 'ins'], l:ext) >= 0
-        \ ? a:main : ''
+  " Recognise (La)TeX-related file extensions
+  let l:new.tex = l:ext =~? '^\(\(la\)\=tex\|dtx\|tikz\)$' ? a:main : ''
```
lervag, line 22: I don't want the comment here (for me, it is very clear that this is the intent). Also, you missed the `.ins` extension. I would suggest:

```vim
let l:new.tex = l:ext =~? '\v^%(la)?tex|dtx|tikz|ins)$'
```

lervag: The `%` is used to indicate that we don't need to remember the group content. It should (VERY slightly) speed up the regex. The `\v` turns on very magic mode, which makes it possible to avoid the escaping.

lervag:
Thanks, I agree, this seems to be a relevant fix. I proposed another minor fix and adjustment and will merge when that's applied.

lervag (changes requested):

autoload/vimtex/state/class.vim
tyckesak: You got it.

_Jesus Christ, how much I grew to hate Vimscript regexes._

lervag:
Thanks! Looks fine now! I took the liberty of squashing your commits on top of master to keep a cleaner Git history.

PR #2586, opened 2022-11-28
Added shortintertext as text context.

senesc:
Currently the content of \shortintertext{} is consideted math context by the plugin. With this PR I added shortintertext to the "Text Inside Math regions" patterns.

### DIFF
--- a/autoload/vimtex/syntax/p/mathtools.vim
+++ b/autoload/vimtex/syntax/p/mathtools.vim
@@ -23,6 +23,14 @@ function! vimtex#syntax#p#mathtools#load(cfg) abort " {{{1
   highlight def link texMathToolsOptPos1  texOpt
   highlight def link texMathToolsOptPos2  texOpt
   highlight def link texMathToolsOptWidth texOpt
+
+  " Support for shortintertext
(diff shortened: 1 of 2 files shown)


lervag:
Can you provide a minimal example document that is relevant here?

senesc:
Sure!

```
\documentclass{article}
\usepackage{amsmath}
\usepackage{mathtools}

\begin{document}
\begin{gather*}
	e^{i \pi} + 1 = 0
	\shortintertext{This is an example of shortintertext}
	e^{i \pi} + 1 = 0
\end{gather*}
\end{document}
```

lervag:
Cool, thanks. Now, `\shortintertext` is a command from the `mathtools` package, so we should not define this in `core.vim`. Instead, we need to update this file:

https://github.com/lervag/vimtex/blob/master/autoload/vimtex/syntax/p/mathtools.vim

In addition, I would kindly ask that you update the corresponding test file with a minimal example (just add an equation with the relevant command):

https://github.com/lervag/vimtex/blob/master/test/test-syntax/test-mathtools.tex

senesc:
Thanks for your input! I updated the correct file and added an example to the test folder. Hope everything is okay now

lervag:
Thanks; I've merged (and then applied some minor adjustments).

PR #2612, opened 2022-12-31
Allow creating starred command

Aster89:
I know that `tsc` exists, but would this change be out of the world?

### DIFF
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -126,7 +126,7 @@ endfunction
 function! vimtex#cmd#create_insert() abort " {{{1
   if mode() !=# 'i' | return | endif
 
-  let l:re = '\v%(^|\A)\zs\a+\ze%(\A|$)'
+  let l:re = '\v%(^|\A)\zs\a+(\*=)@>\a*\ze%(\A|$)'
   let l:c0 = col('.') - 1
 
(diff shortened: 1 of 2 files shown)


lervag:
No, this does not seem bad. Perhaps you could include a test for this, e.g. in [this file](https://github.com/Aster89/vimtex/blob/master/test/test-commands/test-insert-cmd.vim)?

Aster89:
At the moment if I run ` MYVIM="vim -T dumb --not-a-term --noplugin -n" make` in a clean repo, it fails. Can't even tell how, as the terminal gets rewritten so I'm not sure what come first and what last. I see this (it's a video):
[asciicast](https://asciinema.org/a/j0ugyWQKpS9AIpinegNCrAfpU)


Besides, I guess we also want that hitting <kbd>F7</kbd> when the cursor is at `|` in insert mode in `foo*|bar` causes the result to be `\foo*{bar|`, right? If so, my change doesn't work this way yet. I have to improve it :D

Aster89:
> Besides, I guess we also want that hitting F7 when the cursor is at `|` in insert mode in `foo*|bar` causes the result to be `\foo*{bar|`, right? If so, my change doesn't work this way yet. I have to improve it :D

Done because why not, but I couldn't run the tests.

lervag:
Thanks! I fixed the failing tests before merging.

PR #2627, opened 2023-02-05
fix the problem that open viewer in windows/neovim will show a cmd pr…

archerc:
In the case that runing on a Windows / neovim installation, the vimtex general viewer will open the viewer with a command prompt shown instead of open it silently.  I found that this is because the job run a command with a prefix `"cmd /s /c "` or  in a list with head `["sh", "-c" ]`. This may be fine on other platforms but not in my case. I want to prefix the command with `"start "` so I need a place to customize it as in this pull request.

### DIFF
--- a/autoload/vimtex/jobs/neovim.vim
+++ b/autoload/vimtex/jobs/neovim.vim
@@ -6,9 +6,18 @@
 
 function! vimtex#jobs#neovim#new(cmd) abort " {{{1
   let l:job = deepcopy(s:job)
-  let l:job.cmd = has('win32')
-        \ ? 'cmd /s /c "' . a:cmd . '"'
-        \ : ['sh', '-c', a:cmd]
+  let l:modifier = get(g:, 'vimtex_job_cmd_modifier', has('win32') ? 'start ' : '')
+  if type(l:modifier) == v:t_func || (type(l:modifier) == v:t_string && exists('*' . l:modifier)) 
+    let l:job.cmd = call(l:modifier, a:cmd)
+  else
+    if empty(l:modifier)
+      let l:job.cmd = has('win32')
+            \ ? 'cmd /s /c "' . a:cmd . '"'
+            \ : ['sh', '-c', a:cmd]
+    else
+      let l:job.cmd = l:modifier . a:cmd
+    endif
+  endif
   return l:job
 endfunction
 


lervag:
> In the case that runing on a Windows / neovim installation, the vimtex general viewer will open the viewer with a command prompt shown instead of open it silently. I found that this is because the job run a command with a prefix `"cmd /s /c "` or in a list with head `["sh", "-c" ]`. This may be fine on other platforms but not in my case. I want to prefix the command with `"start "` so I need a place to customize it as in this pull request.

Thanks, I appreciate the effort of looking into this in depth and suggesting an update!

You propose this (I've made a few minor tweaks for readability):

```vim
function! vimtex#jobs#neovim#new(cmd) abort " {{{1
  let l:job = deepcopy(s:job)

  let l:modifier = get(g:, 'vimtex_job_cmd_modifier',
        \ has('win32') ? 'start ' : '')
  if empty(l:modifier)
    let l:job.cmd = has('win32')
          \ ? 'cmd /s /c "' . a:cmd . '"'
          \ : ['sh', '-c', a:cmd]
  elseif type(l:modifier) == v:t_func
        \ || (type(l:modifier) == v:t_string && exists('*' . l:modifier))
    let l:job.cmd = call(l:modifier, a:cmd)
  else
    let l:job.cmd = l:modifier . a:cmd
  endif

  return l:job
endfunction
```

So:

* You are changing the current default behaviour on Windows. I don't actually mind that, but it does need some thorough testing! I don't use Windows, so it means you (and possibly others) need to do this testing. E.g., does this break any current behaviour that uses the jobs api?
* You are making this change only for neovim, but we should also consider the Vim version: https://github.com/lervag/vimtex/blob/master/autoload/vimtex/jobs/vim.vim#L7-L13.
* Do we really _need_ the new option? I would be much more happy if we can instead find something that "just works"™ for everyone. If `start` is a better way to run system commands on Windows than `cmd /s /c`, then I won't mind just changing this at the core without any options.
* If we want the option: It needs to be properly documented.

VimTeX uses `vimtex#jobs#neovim#new` through `vimtex#jobs#start`, which us currently used for starting the generic viewer, Sioyek (viewer), and for opening a browser or similar for handling viewing documentation. There are some other system call api functions that works slightly differently, but that also uses `cmd /s /c` on Windows. Does `start` work well as a substitute for `cmd /s /c` for _every_ type of system call?

lervag:
I'm curious if #2717 resolves this issue?

PR #2633, opened 2023-02-18
Add wayland viewer workaround option g:vimtex_wayland

### LINKED ISSUE #2046: Request: Wayland Support

rrueger:
Attempts to fix #2046

More information at https://github.com/lervag/vimtex/issues/2046#issuecomment-1435778939

I have not updated the documentation yet to reflect this change. @lervag I would need some pointers to do this (Where in the documentation hierarchy this should go, etc.)

Adding a `g:vimtex_wayland` option is somewhat conservative, since it ensures backward compatibility

### DIFF
--- a/autoload/vimtex/view/_template.vim
+++ b/autoload/vimtex/view/_template.vim
@@ -203,6 +203,10 @@ endfunction
 
 " }}}1
 function! s:viewer.xdo_find_win_id_by_pid() dict abort " {{{1
+  g:vimtex_wayland = get(g:, 'vimtex_wayland', 0)
+  if g:vimtex_wayland
+    return 1
+  endif
   " Attempt to find the viewer's X window ID by the viewer's process ID.
   " Returns the viewer's window ID if one is found or 0. If more than one ID is
   " found, return the first.


lervag:
Thanks for looking into this and for suggesting a way forward!

> Adding a `g:vimtex_wayland` option is somewhat conservative, since it ensures backward compatibility

Sorry, but this is not the direction I want to go. I've pushed a branch with an experimental alternative and will explain the idea in #2046.

Some comments:

> I spend quite some time trying to understand the existing code and writing a fix. I tried using `swaymsg` to get window settings directly from sway (the Wayland Compositor in my case), using `lsof` and `pgrep`.

It is not surprising that this was not so trivial. I still believe it may be possible, but perhaps it is not necessary.

> However, for some reason the replacing the body of `s:viewer.xdo_find_win_id_by_pid()` with `return 1` fixes most of the functionality using **Zathura** under **Sway**.

This solution is _not_ a solution, it is simply a hack that basically disables the `xdotool` stuff. However, it seems you are pointing out that the existing Zathura implementation in VimTeX works quite well for Wayland if we simply ignore the `xdotool` things that improve some quality of life. That is indeed interesting.

> What doesn't work is programmatically changing focus from Vim to Zathura within Vim (after `:VimtexView`, focus stays on Vim)

And I don't know any good ways to achieve this without `xdotool`, sorry.

lervag:
Please see #2639 for the alternative approach from me.

PR #2639, opened 2023-02-19
Add simple Zathura viewer alternative

lervag:
This PR adds a `zathura_simple` viewer variant that does not depend on `xdotool`. This means it should work well on e.g. Wayland (see e.g. #2046).

### DIFF
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -44,7 +44,7 @@ function! vimtex#view#zathura#cmdline(outfile, synctex, start) abort " {{{1
     endif
   endif
 
-  if a:synctex && (!(a:start == 1) || g:vimtex_view_forward_search_on_start)
+  if a:synctex && (a:start != 1 || g:vimtex_view_forward_search_on_start)
     let l:cmd .= printf(
           \ ' --synctex-forward %d:%d:%s',
           \ line('.'), col('.'),
--- a/autoload/vimtex/view/zathura.vim
+++ b/autoload/vimtex/view/zathura.vim
@@ -9,15 +9,7 @@ function! vimtex#view#zathura#new() abort " {{{1
 endfunction
 
 " }}}1
-
-
-let s:viewer = vimtex#view#_template#new({
-      \ 'name': 'Zathura',
-      \ 'xwin_id': 0,
-      \ 'has_synctex': 1,
-      \})
-
-function! s:viewer._check() dict abort " {{{1
+function! vimtex#view#zathura#check(viewer) abort " {{{1
   " Check if Zathura is executable
   if !executable('zathura')
     call vimtex#log#error('Zathura is not executable!')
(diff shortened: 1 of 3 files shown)


lervag:
I would be glad to have some users test this on e.g. Wayland. To test:

* Checkout the `fest/zathura-wayland` branch of VimTeX.
* Use the `zathura_simple` viewer, i.e.:

    ```vim
    let g:vimtex_view_method = 'zathura_simple'
    ```

Things to test:

* Does it generally work?
* Does synctex work (forward and backward search)?
* Do you experience interference if you open multiple vim/neovim + VimTeX sessions with multiple Zathura instances?
    * If so, do you experience this even if you have different file names for the source files (e.g. `main.tex` vs `test.tex`)?

gssfaw:
Hello, I want to mention that the default version of plugin was already just working on wayland in my case, i.e. I haven't encountered multiple zathura windows. 
I've tried this branch, it also works well:
It opens one windows after compilation, after recompiling that one window updates

lervag:
> Hello, I want to mention that the default version of plugin was already just working on wayland in my case, i.e. I haven't encountered multiple zathura windows.

Ok! Still, the `xdotool` stuff will make the standard Zathura version less than optimal when `xdotool` is not available, and I think this simple variant should work better in these cases.

> I've tried this branch, it also works well: …

Thanks for testing! I'll wait some more time before I merge this to see if more people will test and provide feedback.

rrueger:
Thank you for continuing to work on this (especially since you don't use Wayland yourself!)

Unfortunately, this doesn't work for me. I am having multiple viewers being opened and search doesn't work (new viewers are opened in the wrong place).

I don't want to speculate too much on *why* this is the case, but I found that appending 

    function! s:viewer._exists() dict abort " {{{1
      return v:true
    endfunction
    
    " }}}1
    function! s:viewer._forward_search(outfile) dict abort " {{{1
      if !self.has_synctex | return | endif
    
      let l:synctex_file = fnamemodify(a:outfile, ':r') . '.synctex.gz'
      if !filereadable(l:synctex_file) | return | endif
    
      let self.cmd_forward_search
            \ = vimtex#view#zathura#cmdline(a:outfile, self.has_synctex, 0)
    
      call vimtex#jobs#run(self.cmd_forward_search)
    endfunction
    
    " }}}1

to `zathura_simple.vim` fixes both issues. It looks to be that `s:viewer_exists()` is not defined in `zathura_simple` and therefore a new viewer is opened every time. @lervag  even mentioned this in a [comment](https://github.com/lervag/vimtex/issues/2046#issuecomment-1221508105).

> When I changed the zathura._exists() function to always return false, then the above will open multiple viewers.

Surprisingly (to me), telling VimTex that a viewer is always open (`return v:true`) yields the correct behaviour.

This has been tested like my [last comment](https://github.com/lervag/vimtex/issues/2046#issuecomment-1435778939). Opening windows, opening different pdfs etc.

lervag:
> Unfortunately, this doesn't work for me. I am having multiple viewers being opened and search doesn't work (new viewers are opened in the wrong place).

Strange.

> I don't want to speculate too much on _why_ this is the case, but I found that appending …

That shouldn't be necessary. The idea is to never use the `forward_search` method. But I notice I made a mistake, so I'm pushing a new version that I believe should work. Please update and test again.

> It looks to be that `s:viewer_exists()` is not defined in `zathura_simple`

Yes, the idea is to _always_ include the forward and inverse search command-line options. Then we don't need anything except the `_start()` method.

rrueger:
Brilliant, now it works perfectly for me!

Thank you so much for your fast work!

lervag:
Cool, thanks for testing and for the useful feedback!

I'll wait for a few more days before I merge this; and I'll add proper documentation of course.

PR #2669, opened 2023-03-16
fix(view): update skim check to macOS changes

clason:
Problem: macOS 13.4 adds a deprecation warning to the introspection method Skim uses for its scripting. This breaks the check whether Skim is installed, leading to an error message because the looked-for output is now on a second line.

Solution: Concatenate all output lines instead of only looking at the first line.


lervag:
Thanks!

Issue #2534, opened 2022-10-28
VimTex can't find log file


tyalie:
### Description

VimTex cannot find the log file when latexmk is used together with `$emulate_aux=1`. When using `:VimtexErrors` (or similar) in Vim, the message `VimTeX: No log file found` appears.

### Steps to reproduce

1. Create `.latexmk` with:
    ```perl
    $emulate_aux = 1
    $out_dir = "gen"
    $aux_dir = ".aux"
    ```
2. open vim and compile
3. try to use `:VimtexErrors`

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
System info:
  OS: Ubuntu 22.04.1 LTS
  Vim version: NVIM v0.7.2
  Has clientserver: true
  Servername: /tmp/nvimo5jleA/0

VimTeX project: main
  base: main.tex
  root: /project
  tex: /project/main.tex
  main parser: current file verified
  document class: IEEEtran
  packages: base
  source files:
    main.tex
    chapters/01_introduction.tex
  compiler: latexmk
    engine: -lualatex
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    build_dir: gen
    callback: 1
    continuous: 1
    executable: latexmk
    job: 
      jobid: 19
      output: /tmp/nvimo5jleA/2
      cmd: max_print_line=2000 latexmk -verbose -file-line-error -synctex=1 -interaction=nonstopmode -lualatex -outdir=gen -pvc -view=none -e '$compiling_cmd = ($compiling_cmd ? $compiling_cmd . " ; " : "") . "echo vimtex_compiler_callback_compiling"' -e '$success_cmd = ($success_cmd ? $success_cmd . " ; " : "") . "echo vimtex_compiler_callback_success"' -e '$failure_cmd = ($failure_cmd ? $failure_cmd . " ; " : "") . "echo vimtex_compiler_callback_failure"' 'main.tex'
      pid: 1460577
  viewer: General
  qf method: LaTeX logfilel
```

lervag:
> VimTex cannot find the log file when latexmk is used together with `$emulate_aux=1`. When using `:VimtexErrors` (or similar) in Vim, the message `VimTeX: No log file found` appears.

> Create `.latexmk` with:
> 
> ```perl
> $emulate_aux = 1
> $out_dir = "gen"
> $aux_dir = ".aux"
> ```

[...]

PR #2677, opened 2023-03-28
Add support for aux_dir for latexmk

lervag:
See discussion in #2534.

This adds a breaking change: `build_dir` is changed to `out_dir` throughout. Thus, anyone who changes the `build_dir` key of e.g. `g:vimtex_compiler_latexmk` will need to update their configuration!

The term "`build_dir`" is not really used in any of the backend compiler tools, so it was probably a mistake to add a new term here. Therefore, as a step towards adding support for `aux_dir` for the latexmk backend, we change `build_dir` to `out_dir`. `out_dir` (or similar variants) is a specific "concept" in both latexmk, latexrun and tectonic.


doronbehar:
Hey, I am also experiencing perhaps related issues with `build_dir`. I was about to suggest the following PR:

```patch
commit 823c8f4152fb86e549355f401e44ae356b1ded49
Author: Doron Behar <doron.behar@gmail.com>
Date:   Sat May 20 17:48:37 2023 +0300

    Document the real default build_dir compiler option

diff --git a/doc/vimtex.txt b/doc/vimtex.txt
index 0a6bff4e..7fb6b01a 100644
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -962,7 +962,7 @@ OPTIONS                                                        *vimtex-options*
   Default value: >vim
 
     let g:vimtex_compiler_latexmk = {
-        \ 'build_dir' : '',
+        \ 'build_dir' : {-> expand("%:p:h")},
         \ 'callback' : 1,
         \ 'continuous' : 1,
         \ 'executable' : 'latexmk',
```

I took the empty `build_dir` option when I set my `g:vimtex_compiler_tectonic` suggested by the docs and I noticed that `VimtexClean` doesn't work for me because it attempts to clean files in my root directory `/`. Hence the minimum is for the documentation to be correct :).

lervag:
> I took the empty `build_dir` option when I set my `g:vimtex_compiler_tectonic` suggested by the docs and I noticed that `VimtexClean` doesn't work for me because it attempts to clean files in my root directory `/`. Hence the minimum is for the documentation to be correct :).

I agree we should update the docs if they are wrong. However, I don't understand why you claim the docs are wrong. I may just be a tad slow, but as far as I understand, the `'build_dir'` value is an empty string unless it is changed.

Although I see why you find this related, I still find it to be a separate issue. I would therefore prefer if we can continue this discussion in a new thread if you should still think the docs are wrong.

tertsdiepraam:
I'm just passing by, but I ran into a similar problem as mentioned in the issue and wanted to confirm that it was indeed fixed by using this branch. Thank you!

lervag:
I've updated the branch with a force push. Let me know if you need help checking out the updated branch.

lervag:
I'm merging this now. It is a breaking change and I'll make a release to explain necessary steps for those who may be impacted.

tyalie:
Hi sorry for not answering. Needed to recover medically and I found myself in the situation that I wasn't to do what I initially planned to do.

But I've been able to successfully test this PR on the paper that inspired the original issue #2534. Thank you so much for this and investing the time for it

lervag:
> Hi sorry for not answering. Needed to recover medically and I found myself in the situation that I wasn't to do what I initially planned to do.

I'm sorry to hear that and I hope you are doing well, or at least that the outlook is good! <3

> But I've been able to successfully test this PR on the paper that inspired the original issue #2534. Thank you so much for this and investing the time for it

Thank you for the feedback and thank you for pushing me to do this. I believe the changes you've inspired are good and are very sensible.

All the best!

PR #2689, opened 2023-04-20
fix: use non-deprecated alternatives of checkhealth functions

dundargoc:


lervag:
This would imply that we need to update the minimum requirement of VimTeX to neovim 0.9, right? The requirements are stated [here](https://github.com/lervag/vimtex#requirements). Currently, VimTeX should work well for neovim all the way back to 0.4.3. I actually don't mind to make changes that will bump this requirement, but this particular update does not seem to be important enough to bump all they way from 0.4.3 to 0.9.

As far as I know, the deprecated function will work until 0.11 - and when the deprecated functions no longer work I would not mind making the update.

lervag:
By the way, I just had a discussion with @clason about this very thing on Gitter/Element in the chat room: #vimtex-chat_community:gitter.im.

dundargoc:
> but this particular update does not seem to be important enough to bump all they way from 0.4.3 to 0.9.

Yes, that is fair enough. Thank you for your time.

lervag:
No problem, and thank you for noticing and suggesting this with a PR; I appreciate the effort! Even if I in this case decided against merging it.

Issue #2683, opened 2023-04-11
`latexmk` engine `context (pdftex)` fails to generate log file when file name includes a space


ehudbehar:
### Description

Suppose I have the file `my file.tex` which I want to compile with `latexmk` and `pdftex` (NOT `pdflatex`!).

I use `:let g:vimtex_compiler_latexmk_engines._ = "-pdf -pdflatex=texexec"` to change the `latexmk` engine.

When the file has no space in its file name, it works fine - by examining the log file. Once a space is found in the file name, as in `my file.tex`,  compilation fails to produce a log file.

### Steps to reproduce

1. Create file `my file.tex`
2. Write the ex command `:let g:vimtex_compiler_latexmk_engines._ = "-pdf -pdflatex=texexec"`
3. Compile with `<leader>ll`
4. Look at the failure with `<leader>lo`

### Expected behavior

vim compiles my file with pdftex and a log file is created

### Actual behavior

I see the following errors when pressing `<leader>lo`:

```
Rc files read:
  NONE
Latexmk: This is Latexmk, John Collins, 20 November 2021, version: 4.76.
======= Need to update make_preview_continuous for target files
Not using a previewer
------------
Running 'echo vimtex_compiler_callback_compiling'
------------
vimtex_compiler_callback_compiling
Rule 'pdflatex': File changes, etc:
   Changed files, or newly in use since previous run(s):
      'new one.tex'
------------
Run number 1 of rule 'pdflatex'
------------
Latexmk: applying rule 'pdflatex'...
------------
Running 'texexec  -file-line-error -synctex=1 -interaction=nonstopmode -recorder  "new one.tex"'
------------

Latexmk: fls file doesn't appear to have been made.
Latexmk: Errors, so I did not complete making targets
Latexmk: *LaTeX didn't generate the expected log file 'new one.log'
    ==> You will need to change a source file before I do another run <==
------------
Running 'echo vimtex_compiler_callback_failure'
------------
vimtex_compiler_callback_failure
Collected error summary (may duplicate other messages):
  pdflatex: gave an error
```

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info:
  OS: macOS 12.1 (21C52)
  Vim version: VIM 9.0 (1-1276)
  Has clientserver: true
  Servername: VIM

VimTeX project: new one
  base: new one.tex
  root: /Users/udi/Desktop
  tex: /Users/udi/Desktop/new one.tex
  main parser: current file verified
  document class: article
[...]

PR #2705, opened 2023-05-13
Add more engine names as options to latexmk

ehudbehar:
Continuing [issue #2683](https://github.com/lervag/vimtex/issues/2683), I added 3 more lines to the documentation that describe how to compile a plain tex document with either `pdftex`, `xetex` or `luatex`.

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1069,6 +1069,9 @@ OPTIONS                                                        *vimtex-options*
         \ 'luatex'           : '-lualatex',
         \ 'lualatex'         : '-lualatex',
         \ 'xelatex'          : '-xelatex',
+        \ 'pdftex'           : '-pdf -pdflatex=pdftex'
+        \ 'xetex'            : '-pdf -pdflatex=xetex'
+        \ 'luatex'           : '-pdf -pdflatex=luatex'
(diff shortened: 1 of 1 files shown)


lervag:
Thanks!

lervag:
Sorry, I realize I made an error here. The documentation we changed did not list an example; instead, it displays the default value of the option. We should either 1) revert this change, or 2) add the added lines to the default value as well. What do you think?

user202729:
Just noticed this issue.

I think adding to the default value is a fine option.

Unrelated remark, I think that latexmk does not "officially" support running anything other than LaTeX
https://www.overleaf.com/learn/latex/Questions/Can_I_run_plain_TeX_on_Overleaf%3F

so "telling" latexmk that "to call the pdflatex compiler call pdftex instead" is a workaround.

That having said, there may be some performance benefit to "fake" xetex for xelatex instead of for pdflatex: https://stackoverflow.com/a/57734531/5267751 -- testing is needed.

lervag:
Thanks for the interesting remarks, @user202729. I think I want to revert the change instead of adding to the default list. My main point is that the option already allows to change these things for those who want/need them. This group of people is clearly small (as I have never had any similar requests before this one).

PR #2713, opened 2023-05-20
Add .xdv extension to compiler.clean function

doronbehar:
At least 1 compiler (Tectonic) produces these intermediate files.


lervag:
Seems OK. However, I think we could probably improve this even more by adding more well-known auxiliary file extensions and/or make an option for this. I don't like adding options, but adding an option would also bring flexibility to those who want it.

PR #2741, opened 2023-06-22
add imaps

JingMatrix:
Hi, lervag!

May I add some defaults shortcuts to vimtex?
I found them a bit useful in my work.


lervag:
Thanks! I see no reason not to accept this.

PR #2756, opened 2023-07-12
Feature: Add g:vimtex_syntax_custom_envs

### LINKED ISSUE #2745: Provide a way to add/identify custom listings environments
**Is your feature request related to a problem? Please describe it.**
I'm using the [listings](https://ctan.org/pkg/listings?lang=en) package with custom environments:

`commons.sty`
```latex
% ...
\RequirePackage{listings}
\lstset{aboveskip=\baselineskip, backgroundcolor=\color{themegray}, tabsize=4, ...}
\lstnewenvironment{code}[1][]{\lstset{#1}}{}
% ...
```
`test.tex`
```latex
% ...
\begin{code}[language=python]
def test(argument: str) -> str:
    """
    Test function.
    """
    return argument
\end{code}
% ...
```

Sadly, in `test.tex` the code is not highlighted as Python, and neither is the `code` environment suggested when typing `\begin{`. The `lstnewenvironment` command is in a separate `sty` file (installed under `~/texmf/tex/latex/...`)  but I also tried inlining it in `test.tex` and it didn't work. Commands are otherwise properly identified and autocompleted from `commons.sty`, so I suspect the issue is specifically related to `lstnewenvironment`. I've compiled the tex source and cleared the cache, did not help. Using VimTeX 2.13.

Note that if I change `\begin{code}` to `\begin{lstlisting}` the syntax highlighting and autocompletion works as expected.

Optimally the `lstnewenvironment` environments would be automatically picked up, but I would also be fine with a config option where I can set aliases for `lstlisting` environments.

lervag:
This should resolve #2745.

### DIFF
--- a/autoload/vimtex/options.vim
+++ b/autoload/vimtex/options.vim
@@ -349,6 +349,7 @@ function! vimtex#options#init() abort " {{{1
   call s:init_option('vimtex_syntax_conceal_disable', 0)
   call s:init_option('vimtex_syntax_custom_cmds', [])
   call s:init_option('vimtex_syntax_custom_cmds_with_concealed_delims', [])
+  call s:init_option('vimtex_syntax_custom_envs', [])
   call s:init_option('vimtex_syntax_nested', {
         \ 'aliases' : {
         \   'C' : 'c',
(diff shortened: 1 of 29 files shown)


lervag:
Thanks, I fixed both of those now. Note, though, that the latter issue is really due to you forgetting to pass the `region` key. But I realize that VimTeX can just forgive that and create a region name based on the "name" key.

lervag:
Thanks; I've made a minor adjustment to the docs regarding the region name. And I merged this, since it seems good enough for now :)

lervag:
Glad to hear it :)

PR #2759, opened 2023-07-19
Move mouse pointer to activated window

ax3ghazy:
In case it's needed, this prevents some conflicts with "focus follows pointer" features in some WMs.


lervag:
Hmm. Why not. I doubt this will be harmful to anyone, while I agree it may prevent issues for some people.

PR #2766, opened 2023-08-06
Add option to prevent Skim from selecting the text after performing forward search

tsung-ju:


lervag:
This is probably all good, but since I don't use Macs and have no way of testing it: @clason would you care to help me confirm that this may be a  useful addition?

clason:
I'm afk at the moment, but will check it out when I am back.

lervag:
@clason friendly reminder (you may still be afk, but in case you just forgot)

clason:
Oh, sorry, that completely slipped by me. Looks good to me!

PR #2772, opened 2023-08-18
fixed latexmk broken link

gruelingpine185:
Just to be clear, the [URL that points to several Latexmk references](http://personal.psu.edu/jcc8//software/latexmk-jcc/) is broken. As of "July 28, 2023, the Personal web server was retired as a service". I don't know if [this](https://www.cantab.net/users/johncollins/latexmk/index.html) is the exact resource you intend to link, but after stepping through a few directories in the site referred to by the error message, that Is what I managed to find.


lervag:
Thanks!

PR #2798, opened 2023-10-01
fix: detect treesitter in the correct buffer

tomtomjhj:
Problem:
Wrong treesitter detection error message when the user enters another buffer right after opening a tex file.

Solution:
Detect in the buffer in which the ftplugin is loaded.


lervag:
Thanks!

PR #2804, opened 2023-10-03
fix: Don't do timeout in pvc mode after inactivity

wookayin:
By default the `-pvc` mode will stop working after 30 mins of inactivity, where vimtex compiler and the latexmk process can fall into an inconsistent state (i.e., no longer able to build the latex project while vimtex still thinks the continuous mode is still running).

To disable this, we disable inactivity timeout in the pvc mode. Requires latexmk v4.55 (Jan 2018) or higher.

From https://texdoc.org/serve/latexmk/0

```
-pvctimeout
    Do timeout in pvc mode after period of inactivity, which is 30 min.
    by default. Inactivity means a period when latexmk has detected no
    file changes and hence has not taken any actions like compiling the
    document.

-pvctimeout-
    Don’t do timeout in pvc mode after inactivity.
```

### DIFF
--- a/autoload/vimtex/compiler/latexmk.vim
+++ b/autoload/vimtex/compiler/latexmk.vim
@@ -141,7 +141,7 @@ function! s:compiler.__build_cmd() abort dict " {{{1
   endif
 
   if self.continuous
-    let l:cmd .= ' -pvc -view=none'
+    let l:cmd .= ' -pvc -pvctimeout- -view=none'
 
     if self.callback
       for [l:opt, l:val] in [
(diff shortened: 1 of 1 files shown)


wookayin:
Note: This would require latexmk 4.55 or higher; I'm not sure if this would be a concern. If you think we need to update the doc regarding the version requirement, or having some version-dependent feature flag, please let me know.

lervag:
Thanks! First, I agree this makes sense to include. January 2018 is soon 6 years past, right? It could make sense to claim this is a hard requirement. If anyone should experience a problem with it and open an issue, then it may spark a discussion that concludes we should be less rigid.

Still, I'm not sure - perhaps it is better to add a conditional that prevents this option from being added for old versions of latexmk?

wookayin:
Since we didn't have a logic for detecting and comparing version, I'm fine with having it as a strict requirement. Latexmk 4.55 would be shipped with many recent TeX distributions (since 2019), so it wouldn't have a big problem. As you said in any case we can relax version requirements later.

lervag:
Sorry for the late response!

Yes, I agree, let's assume this works and then handle potential issues if it turns out users are on old versions of latexmk. (It may annoy some users, but they should probably be made aware that they are using old versions of software.)

PR #2814, opened 2023-10-19
Fix: Dictionary error in amsmath.vim

DrJupiter:
I was encountering an error, when loading the plugin. To my understanding, it was in the declaration of the dictionary. Changing it to valid(?) syntax fixed my error.

### DIFF
--- a/autoload/vimtex/syntax/p/amsmath.vim
+++ b/autoload/vimtex/syntax/p/amsmath.vim
@@ -16,10 +16,10 @@ function! vimtex#syntax#p#amsmath#load(cfg) abort " {{{1
         \ 'multline',
         \ 'xalignat',
         \]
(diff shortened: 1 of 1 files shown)


lervag:
I'm terribly sorry that I've completely forgotten to respond to this PR! I don't understand how that happened!

So, your PR is necessary because the syntax I've used depends on the literal dict syntax which was introduced in Vim 8.1.1876. At the time you opened the PR, I should have accepted it since I specifically depended only on version 8.0.1453. However, the version requirement has since been bumped up to 8.2.3995. So, I won't accept this PR now, because I will be using more modern syntaxes and features.

However, if you are restricted to use an older Vim version, then I might consider to patch the v2.15 tag. Let me know if this is necessary for you.

Issue #2786, opened 2023-09-06
Nvim-cmp's keyword_pattern and Vimtex's Omnifunc


micangl:
### Description

Vimtex support in nvim-cmp is an issue which has been **extensively** discussed. These are just some of the related issues:
- https://github.com/lervag/vimtex/issues/2215
- https://github.com/hrsh7th/nvim-cmp/issues/833
- https://github.com/hrsh7th/nvim-cmp/issues/1170
- https://github.com/hrsh7th/nvim-cmp/issues/652
- https://github.com/hrsh7th/cmp-omni/issues/1
- https://github.com/hrsh7th/nvim-cmp/issues/258

The problem explained in those issues dealt mainly with the Vimtex-provided neocomplete pattern (https://github.com/lervag/vimtex/blob/master/autoload/vimtex/re.vim#L35), why it didn't work, and why cite-completion wasn't triggered automatically (i.e. the completion menu appeared) as soon as a `\cite{` statement was typed.

I may have found some workarounds and/or hints towards a possible solution. I'll explain them in order:

1. Cite-completion isn't triggered automatically as soon as a `\cite{` statement is issued. Differently from other engines, with nvim-cmp the user first has to type a character to have the menu pop up and show bilbiographic references. This is obviously annoying, as the user may not always remember, at the top of his head, exactly which reference to insert (or the key name). A workaround which I don't think was ever discussed is to use a `trigger_character` in nvim-cmp's configuration, as such:
    ```lua
    { name = "omni", trigger_characters = { "{" } },
    ```
2. Correctness of the Vimtex-provided Neocomplete pattern. When the discussions in the previously linked issues were held, the validity of said pattern was questioned and, if I recall correctly, noone ever managed to make it work with nvim-cmp. The pattern, though, is indeed correct. This can be tested by opening the following file in Neovim and sourcing (`:so`):
    ```lua
    vim.g.regex =
             [=[\%(]=]
          .. [=[\v\\%(]=]
          .. [=[%(\a*cite|Cite)\a*\*?%(\s*\[[^]]*\]){0,2}\s*\{[^}]*]=]
          .. [=[|%(\a*cites|Cites)%(\s*\([^)]*\)){0,2}]=]
          ..    [=[%(%(\s*\[[^]]*\]){0,2}\s*\{[^}]*\})*]=]
          ..    [=[%(\s*\[[^]]*\]){0,2}\s*\{[^}]*]=]
[...]

PR #2816, opened 2023-10-22
Add a new bib parser written in Lua

lervag:
Adding a lua parser for bibtex files.

See #2786.

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "https://raw.githubusercontent.com/LuaLS/vscode-lua/master/setting/schema.json",
+  "runtime": {
+    "version": "LuaJIT"
+  },
+  "workspace": {
+    "library": [
+      "$VIMRUNTIME",
+      "${3rd}/luv/library"
+    ],
+    "checkThirdParty": false
+  }
+}
--- a/.stylua.toml
+++ b/.stylua.toml
@@ -1,0 +1,5 @@
+column_width = 80
+indent_type = "Spaces"
+indent_width = 2
+quote_style = "AutoPreferDouble"
+call_parentheses = "None"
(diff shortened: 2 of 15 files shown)


micangl:
I think there may be an issue. When testing this commit, I've noticed that the time it took for the context-menu to show up wasn't significantly lower. I've tried setting the `g:vimtex_parser_bib_backend` option to alternative with values, but to no avail: the time efficiency was always the same.

When debugging I've seen that, in practically all calls to the `vimtex#parse#bib` function, the argument `{ 'backend': 'vim'}` is provided, effectively voiding the influence of the global backend option.

Also, in the `lua/vimtex/bibparser.lua` file, `M.parse` function, line 230, the `format_entry` is called on every entry; the goal of said function is to format the results in markdown and improve readability (which I don't think could be called a desired effect here?).

lervag:
Yes, sorry, I didn't update the parser for the context menu yet - I've fixed this locally but not pushed it. Thanks for checking, though! Your too fast for me ;)

lervag:
There, I think these are fixed now.

lervag:
> When debugging I've seen that, in practically all calls to the `vimtex#parse#bib` function, the argument `{ 'backend': 'vim'}` is provided, effectively voiding the influence of the global backend option.

Yes; the reason for this is that the context menu _needs_ the source file and line number entries that we do not get with the bibtex backend. The Vimscript parser was written mostly to support the context menu feature.

micangl:
Ok, I've tested the new commit and everything seems to working well.

lervag:
I've improved the code now. It may not be much faster, but at least it is much easier to understand (IMHO).

lervag:
I think that's about as far as I'll go with this. I'll leave the PR open a few days before merging in case anyone wants to test. But I think it should be fine and that it should provide a nice speedup for some people.

lervag:
Ok, I've merged this now.

Notice that I've made some quite significant changes both to the Lua and the Vimscript parsers. I believe I've fixed some bugs in the process, and you should consider to either directly rely on the new VimTeX parser in cmp-vimtex or copy and adapt it as you see fit.

PR #2839, opened 2023-11-27
feat(syntax): add keywords to texCommentTodo

jiangyinzuo:
I believe that the words 'ISSUE' and 'NOTE' are often highlighted in comments as well, and therefore, so I modified the syntax `texCommentTodo`.

### DIFF
--- a/autoload/vimtex/syntax/core.vim
+++ b/autoload/vimtex/syntax/core.vim
@@ -430,6 +430,8 @@ function! vimtex#syntax#core#init_rules() abort " {{{1
   syntax keyword texCommentTodo combak fixme todo xxx
         \ containedin=texComment contained
   syntax case match
+  syntax keyword texCommentTodo ISSUE NOTE
+        \ containedin=texComment contained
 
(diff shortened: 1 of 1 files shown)


lervag:
> I believe that the words 'ISSUE' and 'NOTE' are often highlighted in comments as well, and therefore, so I modified the syntax `texCommentTodo`.

You _may_ be right that these keywords are highlighted in some settings, but I'm not sure I agree they should be highlighted. Both of these words are quite common in regular text, and it can be confusing to have them highlighted in the comments. Notice that we are matching case insensitive; I would be more inclined to accept this if we only matched the upper case variants.

lervag:
I.e., instead of adding to the keywords list, we might instead add a new match rule just below with something like this:

```
  syntax match texCommentTodo "ISSUE\|NOTE"
        \ containedin=texComment contained
```

jiangyinzuo:
I agree with your point of view. I have changed the rule.

lervag:
Thanks!

PR #2886, opened 2024-02-21
feat(conceal): support more superscripts

MithicSpirit:
This adds conceals for some superscripts that weren't supported yet (maybe
they're new? Iosevka seems to have support for them already, at least).

I didn't bother re-aligning the table since I didn't want to mess with the
git-blame history, but I can do so if you want.


lervag:
Thanks! I made a minor adjustment to make it work properly (see the extra commit).

MithicSpirit:
Ah, thank you. Maybe I should've tested it myself before I submitted the PR lol.

lervag:
Always good to test stuff; but better push a nice patch half finished than _not_ push the patch at all! At least, I still appreciate your contribution and it was quick work to fix the missing part!

Issue #2579, opened 2022-11-24
Sioyek: synctex inverse search does not put focus on terminal with vim


3ddiddE:
Like the title already says, the inverse search of sioyek at least seems to be broken for me.
While the cursor is put to the correct row and col in the terminal window that runs vim, the focus stays in the sioyek window.

I do not have these problems with zathura, so at least I can eliminate a problem with my overall system setup I guess.

Do you have any ideas, why the focus is not transferred over to the vim-terminal?

System: Arch
Sioyek Versions Tested: sioyek-appimage, sioyek-git from the AUR

lervag:
If you have `xdotool` installed, then you can use a slightly "hacky" solution: Add this to your `.vimrc` file (or similar; whereever you configure your global autocommands and/or your VimTeX settings):

```vim
augroup init_vimtex
  autocmd!
  autocmd User VimtexEventViewReverse call b:vimtex.viewer.xdo_focus_vim()
augroup END
```

3ddiddE:
Thank you for your solution!

I'd like to understand what the cause for this behavior is.
Is there any particular reason that reverse search stopped working as expected?
Unfortunately, I did not notice which commit has introduced this behavior

lervag:
Yes. There was a change recently due to #2569.

lervag:
Closing due to inactivity.

pyfb78:
Is there a way to do this without xdotool?

lervag:
> Is there a way to do this without xdotool?

Do _what_ exactly? Please, feel free to open a new issue and fully describe your question.

pyfb78:
I'm sorry, I meant for the focus to be transferred back on over to the neovim-terminal after doing an inverse search in sioyek.

pyfb78:
Also, I'm super sorry but I just figured it out and all I had to do was switch vim to neovim.

lervag:
Ah, I'm glad to hear that you got it working.

DaitiDay:
I'm sorry I'm commenting under a closed issue, but is this "workaround" still working? I've tried implementing it in neovim with no success ,-, Looking at the help pages I can't find any reference to the `xdo_focus_vim` function.

lervag:
> I'm sorry I'm commenting under a closed issue, but is this "workaround" still working? I've tried implementing it in neovim with no success ,-, Looking at the help pages I can't find any reference to the `xdo_focus_vim` function.

Yes, it should still work. It is not really documented, though - sorry.

[...]

PR #2894, opened 2024-03-07
Switch back focus to vim/neovim after inverse search from Sioyek

HuidaeCho:
This PR addresses https://github.com/lervag/vimtex/issues/2579.

`xwin_id` was missing in `b:vimtex.viewer`.

### DIFF
--- a/autoload/vimtex/view/sioyek.vim
+++ b/autoload/vimtex/view/sioyek.vim
@@ -13,6 +13,7 @@ endfunction
 
 let s:viewer = vimtex#view#_template#new({
       \ 'name': 'sioyek',
+      \ 'xwin_id': 0,
       \})
 
 function! s:viewer._check() dict abort " {{{1


lervag:
🤔 I'm not exactly sure what you want here. Can you be more explicit about _what_ was not working and _why_ this fixes it?

HuidaeCho:
> 🤔 I'm not exactly sure what you want here. Can you be more explicit about _what_ was not working and _why_ this fixes it?

Sorry for not being clear about it. When I set `let g:vimtex_view_method='sioyek'` and hit `\lv` in neovim, vimtex does a forward search and switches the focus to sioyek. So far so good. Now, from sioyek, `<F4>` (syntex mode) and `right-click` does an inverse search, but the focus would stay on sioyek without this PR because `s:viewer.xdo_check()` returned `false` even with `xdotool` installed.

https://github.com/lervag/vimtex/blob/a77d14b2e4391c0d3e860708bb1fd56161ab68eb/autoload/vimtex/view/_template.vim#L116-L118

The `xwin_id` key was missing in `b:vimtex.viewer` from your `test.vim` in #2579.

lervag:
Did you test that it works for you with this change?

lervag:
Notice that it is _much_ better to include _all_ of the context in a single thread. You are referring to a full other thread. I believe you are thinking about my suggestion here: https://github.com/lervag/vimtex/issues/2579#issuecomment-1326823298, where I suggest adding this:

```vim
augroup init_vimtex
  autocmd!
  autocmd User VimtexEventViewReverse call b:vimtex.viewer.xdo_focus_vim()
augroup END
```

But this particular function does _not_ look for `xwin_id`, so adding it is not necessary.

lervag:
This is the function we are using:

https://github.com/lervag/vimtex/blob/a77d14b2e4391c0d3e860708bb1fd56161ab68eb/autoload/vimtex/view/_template.vim#L220-L222

The dependencies are only that you have `xdotool` and `pstree`; you don't need `xwin_id` in the sioyek viewer for this to work.

HuidaeCho:
> Notice that it is _much_ better to include _all_ of the context in a single thread. You are referring to a full other thread. I believe you are thinking about my suggestion here: [#2579 (comment)](https://github.com/lervag/vimtex/issues/2579#issuecomment-1326823298), where I suggest adding this:
> 
> ```viml
> augroup init_vimtex
>   autocmd!
>   autocmd User VimtexEventViewReverse call b:vimtex.viewer.xdo_focus_vim()
> augroup END
> ```
> 
> But this particular function does _not_ look for `xwin_id`, so adding it is not necessary.

Are you planning to support this focus back for sioyek in vimtex, like for zathura? It would be great not to have to add your suggested setting in `.vimrc`.

lervag:
The feature was never really requested (until now), so I've not considered it.

I'm a little bit skeptical, because it will only work for people with `xdotool` and `pstree`, and perhaps not everyone of those. In such cases, I usually prefer that people instead use the flexibility provided to configure it themselves. It's only 4 lines of code.

HuidaeCho:
> The feature was never really requested (until now), so I've not considered it.
> 
> I'm a little bit skeptical, because it will only work for people with `xdotool` and `pstree`, and perhaps not everyone of those. In such cases, I usually prefer that people instead use the flexibility provided to configure it themselves. It's only 4 lines of code.

Closed the PR. I think my last question is.. Will your setting work for multiple instances of vimtex?

lervag:
Yes, I believe it will work for multiple instances. You should test it, though, as I'm not 100% sure.

HuidaeCho:
> Yes, I believe it will work for multiple instances. You should test it, though, as I'm not 100% sure.

Just FYI, multiple instances don't work. Only one neovim instance receives the inverse search and focus.

lervag:
Ok, thanks for the info!

Issue #2027, opened 2021-04-08
Support for SageTeX


antimike:
## Feature request ("what")
I'd like to suggest that support for the `SageTeX` package be "officially" added to vimtex.
[This has been suggested before](https://github.com/lervag/vimtex/issues/1641), but as the issue was closed because of apparent lack of interest, I'm opening a new one.

### Disclaimer
Although I've been using vim a lot recently, I'm not an expert by any means, so it's possible that there's an obvious solution to the problems I list below which doesn't involve any extra features.  In which case, I apologize.

### The Good
First of all, I'm a huge fan of vimtex---I use it nearly every day for note-taking, homework, and scratch work (I'm a grad student in physics).

I'm also a huge fan of using [Sage](https://www.sagemath.org/) for computations, and it is incredibly convenient to be able to embed simple calculations directly into the markup of a mathematical document via [SageTeX](https://ctan.org/pkg/sagetex?lang=en).  Recently I've become comfortable enough with Sage that almost all my LaTeX documents include some embedded Sage code---for instance, I routinely use Sage (which is really just Python with extra stuff) to construct tedious bits of markup like tables.  (See below for an example of this kind of document structure.)

### The Bad
Unfortunately, editing Sage code, e.g. embedded in a `\begin{sagesilent} ... \end{sagesilent}` environment, is not very functional at the moment.  This is because
* Python is whitespace-delimited, which makes automatic application of TeX indentation rules *very* inconvenient;
* Python textobjects defined by other plugins aren't available in such a context;
* Python linting (via Jedi, e.g.) is either not working or working unpredictably;
* Various other filetype-based odds and ends (e.g., code-commenting keybindings from [NerdCommenter](https://github.com/preservim/nerdcommenter)) aren't working optimally, or at all.  (Another one I use frequently is [vimcmdline](https://github.com/jalvesaq/vimcmdline), which makes running quick tests extremely convenient.  It would be nice to be able to use this plugin, or something similar, to send Sage / Python code to a REPL from within a TeX document.)

### The Ugly
My current workaround is via the [syntax-range](https://github.com/inkarkat/vim-SyntaxRange) plugin, and looks like this:
```vim
autocmd BufReadPost *.tex call SyntaxRange#Include('"""begin=sage"""', '"""end=sage"""', 'sage', 'NonText')
autocmd BufReadPost *.tex call SyntaxRange#Include('%%%begin=sage%%%', '%%%end=sage%%%', 'sage', 'NonText')
```
[...]

PR #2926, opened 2024-04-17
Support for more SageTeX environments

shivangp76:
Builds off of the initial support added in #2027. I didn't think it was necessary to update the tests, but if you feel otherwise, let me know and I can add it.


lervag:
Thanks!

PR #2927, opened 2024-04-18
doc: update the link for the recommended tutorial

ofseed:
A minor fix for beginners. The link would redirect to the homepage of the blog, which contains things not only for vimtex


lervag:
Thanks!

PR #2929, opened 2024-04-19
Complete syntax support for SageTeX

shivangp76:
I read through the [SageTeX documentation](https://github.com/sagemath/sagetex/blob/master/example.tex) to make sure no commands were missed.

- A new in_sage() function was added.
- Bug fix: The texSagetexArg environment is now not considered a mathzone. This comes from the fact that `$2+2=\sage{2+2}$` is a valid use of `\sage` (see [SageTeX documentation](https://github.com/sagemath/sagetex/blob/f9f1ba62c9aa5b3a22f7ac2b5c583df749594cd3/example.tex#L35C20-L35C36)). The contents of its argument should not be considered math, rather Python code.
- The `\sage` and `\sagestr` commands can be used in a mathzone and not in a mathzone, since their purpose is to just convert their Sage input into LaTeX output. Added support for this.

### DIFF
--- a/autoload/vimtex/syntax.vim
+++ b/autoload/vimtex/syntax.vim
@@ -25,18 +25,13 @@ function! vimtex#syntax#in_comment(...) abort " {{{1
 endfunction
 
 " }}}1
-function! vimtex#syntax#in_sage(...) abort " {{{1
-  return call('vimtex#syntax#in', ['texSagetexArg'] + a:000) || call('vimtex#syntax#in', ['texSagetexZone'] + a:000)
-endfunction
-
-" }}}1
 function! vimtex#syntax#in_mathzone(...) abort " {{{1
   " The following checks if we are inside a texMathZone environment. The
-  " arguments to \label{...}, the texRefArg group, the texSagetexArg group,
-  " and \text{...} like commands, the texMathTextArg group, are actively
-  " ignored as these should not be considered to be math environments.
+  " arguments to \label{...}, the texRefArg group, and \text{...} like
+  " commands, the texMathTextArg group, are actively ignored as these should
+  " not be considered to be math environments.
   let l:groups = reverse(call('vimtex#syntax#stack', a:000))
-  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg|SagetexArg)')
+  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg)')
   return l:group =~# '^texMathZone'
 endfunction
 
--- a/autoload/vimtex/syntax.vim
+++ b/autoload/vimtex/syntax.vim
@@ -31,7 +31,7 @@ function! vimtex#syntax#in_mathzone(...) abort " {{{1
   " commands, the texMathTextArg group, are actively ignored as these should
   " not be considered to be math environments.
   let l:groups = reverse(call('vimtex#syntax#stack', a:000))
-  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg)')
+  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg|SagetexArg)')
   return l:group =~# '^texMathZone'
 endfunction
 
--- a/autoload/vimtex/syntax.vim
+++ b/autoload/vimtex/syntax.vim
@@ -25,11 +25,16 @@ function! vimtex#syntax#in_comment(...) abort " {{{1
 endfunction
 
 " }}}1
+function! vimtex#syntax#in_sage(...) abort " {{{1
+  return call('vimtex#syntax#in', ['texSagetexArg'] + a:000) || call('vimtex#syntax#in', ['texSagetexZone'] + a:000)
+endfunction
+
+" }}}1
 function! vimtex#syntax#in_mathzone(...) abort " {{{1
   " The following checks if we are inside a texMathZone environment. The
-  " arguments to \label{...}, the texRefArg group, and \text{...} like
-  " commands, the texMathTextArg group, are actively ignored as these should
-  " not be considered to be math environments.
+  " arguments to \label{...}, the texRefArg group, the texSagetexArg group,
+  " and \text{...} like commands, the texMathTextArg group, are actively
+  " ignored as these should not be considered to be math environments.
   let l:groups = reverse(call('vimtex#syntax#stack', a:000))
   let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg|SagetexArg)')
   return l:group =~# '^texMathZone'
--- a/autoload/vimtex/syntax/p/sagetex.vim
+++ b/autoload/vimtex/syntax/p/sagetex.vim
@@ -16,7 +16,7 @@ function! vimtex#syntax#p#sagetex#load(cfg) abort " {{{1
         \})
   call vimtex#syntax#core#new_opt('texSagetexOpt', {'next': 'texSagetexArg'})
 
-  for l:env in [
+  for l:env_name in [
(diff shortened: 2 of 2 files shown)


autoload/vimtex/syntax.vim
```diff
@@ -24,14 +24,19 @@ function! vimtex#syntax#in_comment(...) abort " {{{1
   return call('vimtex#syntax#in', ['texComment'] + a:000)
 endfunction
 
+" }}}1
+function! vimtex#syntax#in_sage(...) abort " {{{1
```
lervag, line 28: Do we really need this? I don't want to add "extension" code here. Could we instead have the `in_sage` function defined under `vimtex#syntax#p#sagetex#in_sage`?

autoload/vimtex/syntax.vim
```diff
@@ -24,14 +24,19 @@ function! vimtex#syntax#in_comment(...) abort " {{{1
...
+  return call('vimtex#syntax#in', ['texSagetexArg'] + a:000) || call('vimtex#syntax#in', ['texSagetexZone'] + a:000)
```
lervag, line 29: I would much prefer to write this as:

```vim
  return call('vimtex#syntax#in', ['texSagetex\(Arg\|Zone\)'] + a:000)
```

autoload/vimtex/syntax.vim
```diff
@@ -24,14 +24,19 @@ function! vimtex#syntax#in_comment(...) abort " {{{1
...
   " The following checks if we are inside a texMathZone environment. The
-  " arguments to \label{...}, the texRefArg group, and \text{...} like
-  " commands, the texMathTextArg group, are actively ignored as these should
-  " not be considered to be math environments.
+  " arguments to \label{...}, the texRefArg group, the texSagetexArg group,
+  " and \text{...} like commands, the texMathTextArg group, are actively
+  " ignored as these should not be considered to be math environments.
   let l:groups = reverse(call('vimtex#syntax#stack', a:000))
-  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg)')
+  let l:group = matchstr(l:groups, '\v^tex%(Math%(Zone|Text|Tag)|RefArg|SagetexArg)')
```
lervag, line 39: I don't like this. We might need an adjustment here, but this is not it.

autoload/vimtex/syntax/p/sagetex.vim
```diff
@@ -32,6 +30,19 @@ function! vimtex#syntax#p#sagetex#load(cfg) abort " {{{1
           \})
   endfor
 
+  for l:env in [
```
lervag, line 33: `l:env` seems a bad name here. It should be `l:cmd` or `l:cmdname`.

autoload/vimtex/syntax/p/sagetex.vim
```diff
@@ -32,6 +30,19 @@ function! vimtex#syntax#p#sagetex#load(cfg) abort " {{{1
...
+        \ 'sage',
+        \ 'sagestr'
+        \]
+    execute 'syntax match texCmdSagetex /\\' . l:env .
+          \ '\>/ nextgroup=texSagetexArg skipwhite skipnl'
+    call vimtex#syntax#core#new_cmd({
```
lervag, line 39: Why use both the above `syntax match` _and_ the `vimtex#syntax#core#new_cmd`? That will create two match rules. Is that what we want?

autoload/vimtex/syntax/p/sagetex.vim
```diff
@@ -32,6 +30,19 @@ function! vimtex#syntax#p#sagetex#load(cfg) abort " {{{1
...
+          \ 'name': l:env,
+          \ 'mathmode': 1,
```
lervag, line 41: I prefer to use `v:true` instead of 1 for boolean values.

lervag (commented):
Thanks! I believe we should go forward with this, but I have a few comments.

autoload/vimtex/syntax.vim
lervag: When we use the `#new_cmd` function with mathmode, we create syntax groups with these names:

https://github.com/shivangp76/vimtex/blob/9f4be3262f65ab2c357ac4f0a6dc83b73cf9935a/autoload/vimtex/syntax/core.vim#L1006-L1010

Thus, I think we instead should have a pattern like `texMathC\w*Arg`, where `C` stands for "custom" and `\w*` will match the name part. Also, I think this can be added as a different patch, because it will be more like a bugfix.

autoload/vimtex/syntax.vim
shivangp76: I moved it to vimtex/syntax/p/sagetex.vim. The reason I added it was to make snippet integration easier.

autoload/vimtex/syntax/p/sagetex.vim
shivangp76: Yes, I think so. This comes from the fact that these commands are valid both inside and outside math zones. I made some changes to make this clearer.

autoload/vimtex/syntax.vim
shivangp76: This sounds good. I backed this out for now.

autoload/vimtex/syntax.vim
lervag: Huh, I'm curious, do we really need a function for this? Why not just use   `vimtex#syntax#in('texSagetex\(Arg\|Zone\)')`? Considering that this was never requested as a feature before, I would think it is not something many people would end up using, and by using `vimtex#syntax#in` you only need a few more characters.

autoload/vimtex/syntax/p/sagetex.vim
lervag: Great, I think the update is much better!

autoload/vimtex/syntax.vim
shivangp76: That's fair, but can we document this somewhere? If someone is trying to use snippets with sage, I wouldn't expect them to know which environments to check for.

lervag: I've made an attempt to document it now under `:help vimtex#syntax#in` - do you think it is sufficient?

autoload/vimtex/syntax.vim
lervag: Now you can back in again: I've added a function that you can call:

```vim
call vimtex#syntax#add_to_mathzone_ignore('texSagetexArg')
```

Notice, of course, you will need to rebase your work on top of the latest master to have this work. It should be straightforward, something like this:

```sh
git checkout master
git pull
git checkout -
git rebase master
```

Then you would need to do `git push --force` afterwards.

autoload/vimtex/syntax.vim
shivangp76: That's better, but I was more so talking about the specific syntax groups to check (`texSagetexArg` and `texSagetexZone`).

lervag: > That's better, but I was more so talking about the specific syntax groups to check (texSagetexArg and texSagetexZone).

Hmm, ok; how and were do you propose we document that? Feel free to make a follow-up PR for it! I'll be glad to include it, but it does need some consideration. The docs are very long and thorough, anything new we add must be placed at a logical place.

lervag: I'll merge this PR now, so please, if you think the docs should be updated, please open an issue or a PR about that specifically.

lervag:
Thanks; I've merged this now.

autoload/vimtex/syntax.vim
shivangp76: Sounds good, I'll work on this if I have time. Thank you for the help on this PR and for creating this amazing plugin!

lervag: Great; looking forward to seeing further contributions! ;)

And thanks for the kind words!

PR #2947, opened 2024-05-05
Use `U+2216` to conceal `\setminus`

handcart2034:
Hi, currently `\setminus` is concealed by `U+29F5`, the Reverse Solidus Operator, while Unicode provides the genuine Set Minus `U+2216`. For more discussion see this decade long [issue](https://github.com/latex3/unicode-math/issues/181).


lervag:
Thanks!

PR #2948, opened 2024-05-05
New function to toggle `\\`

handcart2034:
Hi, this is a new function that toggles `\\` at line end. In environments like `align`, sometimes the formulas are changed thus the line break is not optimal. One needs `$` or `A` to move to end of line, then either `xx` or `\\`, finally go back to the original position, which can be quite tiring for more than a couple lines.

So wrote this function and map `tsb` to it. If everything looks good will add tests and docs to finalize this request.

### DIFF
--- a/autoload/vimtex/cmd.vim
+++ b/autoload/vimtex/cmd.vim
@@ -28,6 +28,9 @@ function! vimtex#cmd#init_buffer() abort " {{{1
 
   xnoremap <silent><buffer> <plug>(vimtex-cmd-toggle-frac)
         \ :<c-u>call vimtex#cmd#toggle_frac_visual()<cr>
+
+  nnoremap <silent><buffer> <plug>(vimtex-cmd-toggle-break)
+        \ :<c-u>call <sid>operator_setup('toggle_break')<bar>normal! g@l<cr>
 endfunction
 
 " }}}1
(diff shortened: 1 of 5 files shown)


autoload/vimtex/cmd.vim
```diff
@@ -267,6 +270,25 @@ function! vimtex#cmd#toggle_frac_visual() abort " {{{1
...
+  let l:len = col('$') - 1
+
+  if l:len >= 3 && strpart(l:line, l:len - 3) == ' \\'
+    call setline(l:lnum, 
+          \ strpart(l:line, 0, l:len - 3))
+  elseif l:len >= 2 && strpart(l:line, l:len - 2) == '\\'
+    call setline(l:lnum, 
+          \ strpart(l:line, 0, l:len - 2))
+  else
+    call setline(l:lnum,
```
lervag, line 286: In lines 280, 283 and 286 you've included an unnecessary end-of-line space. I prefer to avoid that.

lervag: Also, I think you've unnecessarily broken these statements into multiple lines. I would write it like this:

```vim
  if l:len >= 3 && strpart(l:line, l:len - 3) == ' \\'
    call setline(l:lnum, strpart(l:line, 0, l:len - 3))
  elseif l:len >= 2 && strpart(l:line, l:len - 2) == '\\'
    call setline(l:lnum, strpart(l:line, 0, l:len - 2))
  else
    call setline(l:lnum, l:line . ' \\')
  endif
```

lervag: In fact, I think I would solve this with regexes. Something like this:

```vim
function! vimtex#cmd#toggle_break() abort " {{{1
  let l:lnum = line('.')
  let l:line = getline(l:lnum)

  let l:replace = l:line =~# '\s*\\\\\s*$'
        \ ? substitute(l:line, '\s*\\\\\s*$', '', '')
        \ : substitute(l:line, '\s*$', ' \\\\', '')

  call setline(l:lnum, l:replace)
endfunction

" }}}1
```

lervag:
> Hi, this is a new function that toggles `\\` at line end. In environments like `align`, sometimes the formulas are changed thus the line break is not optimal. One needs `$` or `A` to move to end of line, then either `xx` or `\\`, finally go back to the original position, which can be quite tiring for more than a couple lines.

I don't mind adding this, even though I don't think I'll end up using it. It falls nicely into the current patterns and shouldn't be "in the way" of anything that I can think of. I had a comment the implementation, see above.

> So wrote this function and map `tsb` to it. If everything looks good will add tests and docs to finalize this request.

I think it looks good, although I would prefer if you made the suggested change. And, of course, that you add a couple of tests and docs.

And thanks! I appreciate the effort you've made here in actually implementing the feature you want added!

lervag:
Thanks! I made a minor additional change and then merged this now. The change was basically to change "Toggle line end `\\`" to "Toggle line-break macro `\\`" and to use this name consistently.

lervag:
Note, though: I rebased and merged locally; for some reason, Github doesn't recognize this.

clason:
Neat! Small follow-up suggestion: support for visual mode ;)

lervag:
Heh, yes. @handcart2034 are you up for the challenge? One way to implement it would be to reuse the current implementation and just apply it to every line in the visual selection.

PR #2985, opened 2024-08-08
fix: use sh on path

stephen-huan:
This reverts a change in 2dc2a5435809c13f6f4aa07964be4ef19a7864f5 that I believe is tangential to the issue that the commit addresses. This makes it more consistent with the default value of the `'shell'` option and the rest of the file. Personally, I have a different `sh` on `$PATH` than `/bin/sh` (actually, I do not have `/bin/sh` at all), which is how I noticed the change.

https://github.com/lervag/vimtex/blob/5ac62e0315c6f54f53a7d6da7c485cf8e9cf7240/autoload/vimtex/jobs/neovim.vim#L155-L162

https://github.com/lervag/vimtex/blob/5ac62e0315c6f54f53a7d6da7c485cf8e9cf7240/autoload/vimtex/jobs/vim.vim#L182-L185


lervag:
Thanks, I agree that this seems an unintended side effect.

PR #2990, opened 2024-08-20
just pick any window for inverse search if there are several

ces42:
For some reason I don't understand my `BufReadCmd` makes my files lose syntax highlighting when calling `edit %`. This is triggered by inverse search when I have two windows of the same file open. This avoids it.

I don't 100% understand the code in `vimtex#view#inverse_search` (in particular I don't know what could cause an error in the `try` block), so there might be a downside to doing this.

(sorry, I used the wrong branch in the previous PR)


lervag:
Thanks, that's a good catch!

Issue #3044, opened 2024-11-19
Removing \includeonly from g:vimtex_include_indicators defaults


AsbjornPreuss:
### Description

As suggested in comments of #2519 ```\includeonly``` is does not actually include any files in the document, but only defines which files of actual ```\include```s, that are to be included in the compilation.
Having ```includeonly``` in `g:vimtex_include_indicators` results in files and TODOs to be shown twice in the vimtex-toc.
![image](https://github.com/user-attachments/assets/556f5ea8-6332-4886-9d69-2b0e863dcd01)

### Steps to reproduce

1. Change minimal.vim to point to installation of vimtex
2. `nvim -u minimal.vim minimal.tex`
3. Open vimtex toc (\<leader\>lt)
[minimal.zip](https://github.com/user-attachments/files/17820230/minimal.zip)

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info:
  OS: Ubuntu 20.04.6 LTS
  Vim version: NVIM v0.10.0-dev
  Has clientserver: true
  Servername: /run/user/1001/nvim.9934.0

VimTeX project: minimal
  base: minimal.tex
  root: /tmp/issue
  tex: /tmp/issue/minimal.tex
  main parser: current file verified
  document class: article
  source files:
    minimal.tex
    b.tex
    c.tex
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 1
    executable: latexmk
  viewer: General
  qf method: LaTeX logfile
```

lervag:
This is solved by #3044.

PR #3045, opened 2024-11-19
Fix:  Removed includeonly from g:vimtex_include_indicators defaults

AsbjornPreuss:
Removed includeonly from g:vimtex_include_indicators defaults as a fix to #3044 .


lervag:
Thanks! Very much appreciated!

Issue #1515, opened 2019-10-13
vimtex folding marker problem


slonik-az:
I am using the latest vimtex commit on master (1db32d3) and it seems that vimtex does not recognize marker folds with text on the same line as the opening fold marker.

## Here is the MWE:
I am using custom markers. Here is the relevant `.vimrc` lines:

```
let g:vimtex_fold_enabled = 1
let g:vimtex_fold_types = {
      \ 'markers': {
      \  'open': '<<:',
      \  'close': ':>>',
      \ },
      \}
```

Below is the `foo.tex` file that has two folding regions -- one that works and one which is broken

```
\documentclass{article}
\begin{document}

Broken fold  %<<:
this fold is NOT recognize
by vimtex
%:>>

%<<:
this fold is recognized OK
by vimtex
%:>>

\end{document}
```

When an opening fold marker is on the line by itself preceded only by a comment character `%`  vimtex recognizes fold OK. But fold is broken if other text precedes the opening marker.

slonik-az:
I forgot to mention that I am using vim-8.1 on macosx (MacVim).

lervag:
Thanks. This should work now.

slonik-az:
Thanks for a prompt fix. Works now as expected.

Issue #1502, opened 2019-09-27
How to change fold markers in g:vimtex_fold_types ?


slonik-az:
I find default choice of fold markers `{{{` ... `}}}` quite bad for TeX/LaTeX.  Oftentimes nested commands end in `}}}` like `\emph{\textbf{\textsc{EXAMPLE }}}`. I routinely comment out such lines causing incorrect matching of fold markers.

I found `<<:` ... `:>>` to be a better choice of a fold marker pair for LaTeX. But I do not know how to set it in vimtex. Documentation does not provide any examples for `g:vimtex_fold_types` marker setting.

Any help is greatly appreciated.
Thanks,
--Leo

lervag:
This should be possible now. Please see updated docs for explaining how to configure, if you still need to ask then I must do a better job of documenting...

slonik-az:
Thank you a lot for quickly fixing the issue and updating documentation! I followed the instructions and It works perfectly.

lervag:
No problem, happy to help.

PR #3052, opened 2024-12-08
Check for numbers after markers when setting fold level

rhelder:
# The Problem

Vim's built-in capability for folding code using markers (at least if one is using the default markers, as the Vim documentation recommends) is unsatisfactory when editing LaTeX documents, because it is not so unlikely that a line of LaTeX code will contain three closing braces and be misconstrued by Vim as the end of a fold. VimTeX's folding mechanism improves on this situation by only recognizing markers in comments (although if one uses the default markers, even this can sometimes have unexpected results, as dicussed in e.g. #1502 and #1515). However, VimTeX's folding mechanism falls short compared to Vim's in that VimTeX does not recognize numbers after markers, which are used in Vim to manually specify the fold level.

This difference between VimTeX's folding and Vim's native folding can have not only unexpected but also unfortunate consequences. I've uploaded an [example file](https://github.com/user-attachments/files/18050374/example.sty.txt) with various nested folds (It is supposed to imitate a personal `sty` file, although please note I had to upload it as a `txt` file, since GitHub wouldn't accept a `tex` file. Also, disclaimer: I don't necessarily endorse the folds in the example file with respect to style/organization; I just wanted to make sure there was some good nesting in there. :)) Here's how the file looks at first glance with Vim's built-in folding:
default_fold
But here's how it looks with VimTeX's folding:
vimtex_current_fold
Because Vim allows you to manually specify the fold level, not every opening marker needs to be followed by a closing marker (which is nice, because it's less messy). My general policy, for example, is that I only add a closing marker when I want there to be white space visible after the fold. The result, in this file, is that no opening marker of fold level 1 is followed by a closing marker (except at the end of the document; that way, you can add things to the end of the document outside of the last fold), but every opening marker of fold level greater than 1 is followed by a corresponding closing fold marker. (I think you follow this sort of style in many, though not all, of your VimTeX source files). What happens, then, because VimTeX doesn't recognize the numbers following the fold markers, is that the fold level increases for each `{{{1`, without there being a closing marker to decrease the fold level again. So the fold level gets bigger and bigger, and everything gets subsumed under 'Options'.

# A Solution

Essentially, my code just looks for a number after the marker specified by the user (or the default marker) by appending a capture group to the existing regexes and using `matchlist` to get the number. If there is a number, the the `level` function returns e.g. '>2' if the marker is an opening marker, and e.g. '<2' if the marker is a closing marker. So we get the expected, Vim-like behavior, and VimTeX users can specify the `g:vimtex_fold_types.markers` option in the way that they always have. Here's how it looks at first glance:
vimtex_pr_fold

# The `foldtext`

I personally feel dissatisfied with VimTeX's `foldtext` for markers. The markers themselves don't mean anything, so they shouldn't be part of the text, and there's also no way to see what level each fold is. I considered including an alternative `foldtext` in my PR, but then I decided I was getting myself involved in matters of taste... Others, including you, may well like the text returned by the current `foldtext`. And people like me, if they want e.g. the default Vim `foldtext`, can very easily get it like so:
```
" Make sure the g:vimtex_fold_types.markers dict is defined!
let g:vimtex_fold_types = {'markers': {}}

function! g:vimtex_fold_types.markers.text(line, level) abort dict
  return foldtext()
endfunction
```
The only recommendation I *might* make would be to inform people in the documentation that they can specify their own `foldtext` by adding a funcref to the appropriate dictionary in their `g:vimtex_fold_types` configuration, as above. Someone could only figure this out by reading the code (or reading this post, I suppose). *If* you want to add that tip to the documentation (I can see reasons not to), I would be happy to write something and submit it as a separate PR, if that would save you some time.

I appreciate all of your work on VimTeX, and I hope this code can be helpful. Thank you!

### DIFF
--- a/autoload/vimtex/fold/markers.vim
+++ b/autoload/vimtex/fold/markers.vim
@@ -34,21 +34,17 @@ endfunction
 
 " }}}1
 function! s:folder.level(line, lnum) abort dict " {{{1
-  let l:start = matchlist(a:line, self.re.start . '\(\d\{0,2}\)')
-  let l:end = matchlist(a:line, self.re.end . '\(\d\{0,2}\)')
+  let l:start = matchlist(a:line, self.re.start .. '\(\d\?\)')
+  let l:end = matchlist(a:line, self.re.end .. '\(\d\?\)')
   if !empty(l:start)
-    if !empty(l:start[1])
-      return '>' . l:start[1]
-    else
-      return 'a1'
-    endif
+    return empty(l:start[1])
+          \ ? 'a1'
+          \ : '>' .. l:start[1]
     let self.opened = 1
   elseif !empty(l:end)
-    if !empty(l:end[1])
-      return '<' . l:end[1]
-    else
-      return 's1'
-    endif
+    return empty(l:end[1])
+          \ ? 's1'
+          \ : '<' .. l:end[1]
     let self.opened = 0
   endif
 endfunction
--- a/autoload/vimtex/fold/markers.vim
+++ b/autoload/vimtex/fold/markers.vim
@@ -34,12 +34,22 @@ endfunction
 
 " }}}1
 function! s:folder.level(line, lnum) abort dict " {{{1
-  if a:line =~# self.re.start
+  let l:start = matchlist(a:line, self.re.start . '\(\d\{0,2}\)')
+  let l:end = matchlist(a:line, self.re.end . '\(\d\{0,2}\)')
+  if !empty(l:start)
+    if !empty(l:start[1])
+      return '>' . l:start[1]
+    else
+      return 'a1'
+    endif
     let self.opened = 1
-    return 'a1'
-  elseif a:line =~# self.re.end
+  elseif !empty(l:end)
+    if !empty(l:end[1])
+      return '<' . l:end[1]
+    else
+      return 's1'
+    endif
     let self.opened = 0
-    return 's1'
   endif
 endfunction
 


autoload/vimtex/fold/markers.vim
```diff
@@ -34,12 +34,22 @@ endfunction
 
 " }}}1
 function! s:folder.level(line, lnum) abort dict " {{{1
-  if a:line =~# self.re.start
+  let l:start = matchlist(a:line, self.re.start . '\(\d\{0,2}\)')
```
lervag, line 37: Do we ever need more than _one_ digit here? I would think it suffices to use the regex `\(\d\?\)` here.

autoload/vimtex/fold/markers.vim
```diff
@@ -34,12 +34,22 @@ endfunction
...
+  let l:end = matchlist(a:line, self.re.end . '\(\d\{0,2}\)')
+  if !empty(l:start)
+    if !empty(l:start[1])
+      return '>' . l:start[1]
+    else
+      return 'a1'
+    endif
```
lervag, line 44: I've personally found it better to use the terse ternary variant here:

```vim
    return empty(l:start[1])
          \ ? 'a1'
          \ : '>' .. l:start[1]
```

Also note: I've started to prefer the "modern" `..` syntax for string concatenation.

autoload/vimtex/fold/markers.vim
rhelder: In practice, no – no one would ever need more than one digit. Vim's maximum nesting level for folds is 20, so I was conforming to that. So yes, I agree that `\(\d\?\)` would suffice.

lervag:
Thanks, this seems like a valid and good contribution! I had a couple of minor comments to the code.

Also, a couple of comments to your text:

> However, VimTeX's folding mechanism falls short compared to Vim's in that VimTeX does not recognize numbers after markers, which are used in Vim to manually specify the fold level.

True.

> Also, disclaimer: I don't necessarily endorse the folds in the example file with respect to style/organization; I just wanted to make sure there was some good nesting in there. :)

Haha, good to hear!

> (I think you follow this sort of style in many, though not all, of your VimTeX source files).

Yes; although I've started to shift towards syntax based folding with tree-sitter (only possible with neovim, though).

I find marker based folding to be sort of a "poor man's" solution when a filetype does not have any very sensible default. With treesitter, it seems we are getting closer to good folding for many more syntaxes, which is something I like quite a lot.

> # The `foldtext`
> 
> I personally feel dissatisfied with VimTeX's `foldtext` for markers. The markers themselves don't mean anything, so they shouldn't be part of the text, and there's also no way to see what level each fold is. I considered including an alternative `foldtext` in my PR, but then I decided I was getting myself involved in matters of taste... Others, including you, may well like the text returned by the current `foldtext`. And people like me, if they want e.g. the default Vim `foldtext`, can very easily get it like so:

Thanks for recognizing the matter of taste here. Personally, I like having the markers in the foldtext to make it clear that this is indeed a marker based fold. I recognize that this is clearly a subjective choice, which is one of the reasons I've tried to make it relatively easy to customize for anyone who wants to.

Notice, I think you can make the customization even simpler:

```vim
let g:vimtex_fold_types = #{markers: #{
      \ text: { line, level -> foldtext() }
      \}}
```

> The only recommendation I _might_ make would be to inform people in the documentation that they can specify their own `foldtext` by adding a funcref to the appropriate dictionary in their `g:vimtex_fold_types` configuration, as above. Someone could only figure this out by reading the code (or reading this post, I suppose).

Good point.

> _If_ you want to add that tip to the documentation (I can see reasons not to), I would be happy to write something and submit it as a separate PR, if that would save you some time.

Yes, thank you, I would very much appreciate that!

I think the best place to add this is under the section "FOLDING" and/or under the docs for `g:vimtex_fold_types`.

> I appreciate all of your work on VimTeX, and I hope this code can be helpful. Thank you!

Thanks! Glad you find VimTeX useful and I'm happy to see your contribution!

autoload/vimtex/fold/markers.vim
lervag: Heh, yes, I guess it is _sort_ of a bug to not allow two digits, then. But it's a bug I'm comfortable with having 😅

autoload/vimtex/fold/markers.vim
rhelder: Great, I usually use `..` as well, but was conforming to the rest of the file. I also like the ternary variant (no need to repeat `return`) – I wouldn't have thought of that.

This is my first PR, so I don't know the next step: do I make these changes and resubmit somehow, or do you make them directly?

lervag: If you would care to update your PR with the suggested changes, then I could simply merge directly. If you prefer I make the changes before merging, that's also fine.

And if this is your first PR, then concrats on it being a good one! 👏🏻

rhelder:
> Yes; although I've started to shift towards syntax based folding with tree-sitter (only possible with neovim, though).
>
> I find marker based folding to be sort of a "poor man's" solution when a filetype does not have any very sensible default. With treesitter, it seems we are getting closer to good folding for many more syntaxes, which is something I like quite a lot.

That's helpful to hear your thoughts on marker-based folding – although I submit an example of a `sty` file with this PR, when you start doing package development with e.g. `dtx` files (where user comments become part of the syntax) marker folding becomes more and more problematic, so it's nice to hear your thoughts on alternatives.

> Notice, I think you can make the customization even simpler:

Again, I wouldn't have thought of this – glad to learn something!

> Yes, thank you, I would very much appreciate that!
> 
> I think the best place to add this is under the section "FOLDING" and/or under the docs for g:vimtex_fold_types.

Great, I will get to work on that. Thanks again!

autoload/vimtex/fold/markers.vim
rhelder: Thank you! Ok, I just made the changes.

lervag:
> That's helpful to hear your thoughts on marker-based folding – although I submit an example of a `sty` file with this PR, when you start doing package development with e.g. `dtx` files (where user comments become part of the syntax) marker folding becomes more and more problematic, so it's nice to hear your thoughts on alternatives.

Glad to share my thoughts :)

> > Notice, I think you can make the customization even simpler:
> 
> Again, I wouldn't have thought of this – glad to learn something!

My pleasure!

> > Yes, thank you, I would very much appreciate that!
> > I think the best place to add this is under the section "FOLDING" and/or under the docs for g:vimtex_fold_types.
> 
> Great, I will get to work on that. Thanks again!

Looking forward to seeing it. And thanks to you for this; I've merged it now with a minor fix on top.

PR #3056, opened 2024-12-13
Update docs to explain customization of `foldtext`

rhelder:
As promised in #3052, here is a draft explaining how to customize the text displayed when folds are closed. For readers who are less familiar with dictionary functions, I included both the more verbose and the more concise variants we discussed (with appropriate references to Vim's documentation). I expect there are changes you'll want to make, but I hope this is a good starting point!

### DIFF
--- a/doc/vimtex.txt
+++ b/doc/vimtex.txt
@@ -1810,6 +1810,32 @@ OPTIONS                                                        *vimtex-options*
   can be configured with a list of patterns or similar, the patterns assume
   that one uses very magic regexes (see |\v|).
 
+  One may also customize the text that is displayed in a closed fold of a
+  given type by setting the `text` key to a |Funcref| (see |fold-foldtext| and
+  |Dictionary-function|). The function that the |Funcref| refers to takes two
+  arguments, the text of the current line and the fold level of the current
+  line, and it returns the text that will be displayed in a closed fold. For
+  example, the following configuration will set the text for a closed marker
+  fold to Vim's default: >vim
+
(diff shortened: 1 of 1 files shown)


doc/vimtex.txt
```diff
@@ -1810,6 +1810,32 @@ OPTIONS                                                        *vimtex-options*
...
+  (Note that the |g:vimtex_fold_types| `marker` dictionary must be
+  initialized, even if you are not setting custom markers!)
+
+  Using a |literal-Dict| and a |lambda|, one could specify the same
+  configuration more concisely: >vim
+
+    let g:vimtex_fold_types = #{
+          \ markers : #{text : {line, level -> foldtext()}}
+          \}
+<
```
lervag, line 1838: Perhaps we should also add a Lua example? The following text would work:

```
  And for completeness, the following would work with a Lua based config: >lua

    vim.g.vimtex_fold_types = {
      markers = {
        text = function(line, level)
          return vim.fn.foldtext()
        end
      }
    }
<
```

lervag (changes requested):
I think this is more or less perfect. Thanks! I suggested a minor update to add an example for Lua based configs.

doc/vimtex.txt
rhelder: Sounds good to me! I just pushed the changes.

lervag:
Thanks!

rhelder:
My pleasure! Thanks as always for your work.

Issue #3072, opened 2025-01-10
Toggle between itemize and enumerate


firubat:
I often change my mind as to whether I want a numbered or unnumbered list. It would be nice to have a keymap to toggle between `itemize` and `enumerate` environments, so I won't have to type the entire environment name in the "change environment" dialogue box.

lervag:
Good idea. I'm adding this now. I'll push to a branch and PR first to get some feedback before I merge.

lervag:
Could you checkout #3075 and see if it works for you and if it is what you were thinking?

Perhaps there are other natural environment maps that we could add here?

lervag:
I'll close this issue; let's continue the discussion in the PR.

firubat:
Works for me, thanks a lot! Also commented in the PR.

PR #3075, opened 2025-01-13
Add <plug>(vimtex-env-toggle) for toggling environments

lervag:
This implements a solution for #3072


firubat:
This works for me as I expect :) Thanks a lot!

By the way, I personally remap the toggle commands because the default interferes with the `t` movement. I remap them to `cx...`. But I guess that might be too big a change from what people are already used to.

Regarding your question about other toggles, I can't really think of something I use so often. Maybe toggling different `align` environments? Not sure if it's standard enough to be in the plugin. I'll update if I think of something else.

lervag:
> This works for me as I expect :) Thanks a lot!

Great, thanks for testing and for the feedback!

> By the way, I personally remap the toggle commands because the default interferes with the `t` movement. I remap them to `cx...`. But I guess that might be too big a change from what people are already used to.

Ah, yes. I've found `ts` to be quite safe here, but some people use `s` for motion plugins as well. In any case, it's easy to remap so.

> Regarding your question about other toggles, I can't really think of something I use so often. Maybe toggling different `align` environments? Not sure if it's standard enough to be in the plugin. I'll update if I think of something else.

The point is that it is _trivial_ to add toggles here. But then, it is also trivial to add more as a user. So I guess I could merge now and possibly add more default maps later.

PR #3151, opened 2025-04-13
fix: allow lambdas and Lua functions as doc handlers

gwuen:
When `g:vimtex_doc_handlers` holds a lambda expression or bare Lua function, the iterator itself will become a Funcref. To avoid E704, this Funcref variable name must start with a capital letter.

In practice, this means one can directly pass in a Lua function:

```lua
vim.g.vimtex_doc_handlers = { MyVimtexDocHandler }
```

Instead of having to create a Vim wrapper function, which can be `call()`ed directly by its name:

```vim
function! MyVimtexDocHandler(context)
  return v:lua.MyVimtexDocHandler(a:context)
endfunction

let g:vimtex_doc_handlers = ['MyVimtexDocHandler']
```

### DIFF
--- a/autoload/vimtex/doc.vim
+++ b/autoload/vimtex/doc.vim
@@ -32,9 +32,9 @@ function! vimtex#doc#package(word) abort " {{{1
   let l:context = vimtex#doc#get_context(a:word)
   if empty(l:context) | return | endif
 
-  for l:handler in g:vimtex_doc_handlers
+  for l:Handler in g:vimtex_doc_handlers
     try
-      if call(l:handler, [l:context]) | return | endif
+      if call(l:Handler, [l:context]) | return | endif
     catch /E117/
     endtry
   endfor


gwuen:
By the way, I'd be really happy if someone can explain to me why `:h lua-call` clearly says:

> `v:lua` without a call is not allowed in a Vimscript expression: Funcrefs cannot represent Lua functions.

While the error message suggests that the Lua function I passed is now held **as a Funcref**:

> E704: Funcref variable name must start with a capital: l:handler

Is this because the bare Lua function is treated the same as a lambda expression when handled inside Vim script? This is at least the behavior I observed, but I couldn't find any conclusive information in the Neovim docs or online.

So if my reasoning in the commit text is wrong, feel free to correct me :)

lervag:
Sorry for the slow response!

> When `g:vimtex_doc_handlers` holds a lambda expression or bare Lua function, the iterator itself will become a Funcref. To avoid E704, this Funcref variable name must start with a capital letter.

Yes, that's right, and I think your change is both safe and useful! Thanks!

> By the way, I'd be really happy if someone can explain to me why `:h lua-call` clearly says: …

I'm sorry, I can't explain that.

> And a note for anyone considering using a Lua function: From inside Lua, there's absolutely no way to modify the `context` dictionary passed to the handler outside its scope, see `:h lua-vimscript`: …

Thanks, that's useful to know. And yes, I think you are right that there is no way to change the vimscript reference of the `context` dictionary here if you are moving over to Lua.

----

In any case, thanks for the contribution!

gwuen:
> Sorry for the slow response!

No worries at all! I appreciate your feedback. I'm genuinely impressed that you've been actively maintaining this plugin for so many years – huge thanks for your dedication!

lervag:
> No worries at all! I appreciate your feedback. I'm genuinely impressed that you've been actively maintaining this plugin for so many years – huge thanks for your dedication!

Great, and thanks for the kind words! It's been fun to keep working on this. 😊

PR #3175, opened 2025-05-31
RFC: Add completion file for tkz-euclide

kiryph:
I have several questions:

1. The completions should only be visible within tikz environments. Is this possible?

2. [tkz-euclide](https://ctan.org/pkg/tkz-euclide) offers a package option `[mini]`. In this case only a subset of the completion candidates should be loaded. Here the user wants to use the package [tkz-elements](https://ctan.org/pkg/tkz-elements) and write the coordinate calculations and definitions in lua code.
Is it possible to have variants of completion files for the same package dependent on a package option?

3. When using the tool [robust-externalize](https://ctan.org/pkg/robust-externalize), it is possible to NOT load tikz and the tkz-* packages in the main document. However, the code is still in the main document. Can a VimTeX user specify manually to load certain packages for completion regardless of `.fls` and preamble parsing? I have seen the option `g:vimtex_syntax_packages` with the `load` value of 2. But it is not clear if this option is also considered for omni-completion. 

**Note**
The list of commands is generated from the texmf tree of TeXLive 2024 and tkz-euclide version: v5.10c (April 2024)
 with following ripgrep command applied on the path `/usr/local/texlive/2024/texmf-dist/doc/latex/tkz-euclide`:

```sh
rg -I -N -o "\\\\tkz[a-zA-Z]*" | sort | uniq > $HOME/tkz-euclide
```


lervag:
> 1. The completions should only be visible within tikz environments. Is this possible?

Sort of. The way these completion packages work is that they are loaded if the package with a corresponding name is recognized.

Do you have a small compilable sample `.tex` file that uses one or more of the commands in `tkz-euclide`? That would be nice to include in the PR, as it helps for testing purposes.

> 2. [tkz-euclide](https://ctan.org/pkg/tkz-euclide) offers a package option `[mini]`. In this case only a subset of the completion candidates should be loaded. Here the user wants to use the package [tkz-elements](https://ctan.org/pkg/tkz-elements) and write the coordinate calculations and definitions in lua code.
> Is it possible to have variants of completion files for the same package dependent on a package option?

I don't think that's possible, and I don't see an immediate simple solution for it, sorry. But again, it would be easier to consider it if I saw an example of how this looks.

> 3. When using the tool [robust-externalize](https://ctan.org/pkg/robust-externalize), it is possible to NOT load tikz and the tkz-* packages in the main document. However, the code is still in the main document. Can a VimTeX user specify manually to load certain packages for completion regardless of `.fls` and preamble parsing? I have seen the option `g:vimtex_syntax_packages` with the `load` value of 2. But it is not clear if this option is also considered for omni-completion.

Good question! It is not possible through any features, but you can force it. E.g., if you open some minimal latex file - doing `\li` will show you the recognized packages. Now you can easily add another package by doing `:let b:vimtex.packages["tkz-euclide"] = {}`.

I don't think I want to add a specific feature for it, but it should be very easy to make a simple custom feature here, e.g. a custom mapping or custom command that simply loads a package for you. Something like this, for instance:

```vim
command -nargs=1 AddPackage :let b:vimtex.packages[<q-args>] = {}
```

lervag:
> > Sort of. The way these completion packages work is that they are loaded if the package with a corresponding name is recognized.
> 
> If I get this correctly, the answer is closer to no. They are not shown if the package is not loaded, but it does not depend on the current surrounding environment. …
>
> However, I am not sure how to create different sources for the LaTeX preamble, default LaTeX context, math environments, TikZ environments, and others.

I'm sorry, but I didn't quite understand your thoughts here.

VimTeX relies on omni completion and uses the line context to determine which type of completion will be performed. The `tkz-euclide` completion is a command-completion and is triggered by the slash. The relevant code is here, but I'll admit it is probably not trivial to understand at a glance:

https://github.com/lervag/vimtex/blob/main/autoload/vimtex/complete.vim#L292

> > Do you have a small compilable sample .tex file that uses one or more of the commands in tkz-euclide? That would be nice to include in the PR, as it helps for testing purposes.
> 
> I have added one.

Thanks! I'll adjust it so that the tests work and merge this, then.

> I actually would do it the following way using one of `vimtex-events` …

Or even more Lua:

```lua
local au_group = vim.api.nvim_create_augroup("vimtex_events", {})
vim.api.nvim_create_autocmd('User', {
    pattern = 'VimtexEventInitPost',
    group = au_group,
    callback = function()
      vim.b.vimtex.packages['tkz-euclide'] = {}
    end
  })
```

> First I was surprised, the buffer-local variable `b:vimtex.packages` is apparently removed afterwards: I cannot `:echo b:vimtex.packages`.

That's strange - this does not happen on my end. The variable is present as expected whenever I test it.

kiryph:
**Update**

I checked old issues and apparently we had already a command completion discussion in 2017:

- https://github.com/lervag/vimtex/issues/650
- https://github.com/lervag/vimtex/pull/770

I guess I was asking what you have mentioned back then that this was left to implement:

> What is lacking now is mode specific completion, but I think this should be fine for now.
> 
> If someone would be willing to learn the format of the cwl files from texstudio and explain how to parse the mode specifiers, then I could try to parse that as well and add mode specific command completion. https://github.com/lervag/vimtex/pull/770#issuecomment-294277140

lervag:
> Consider the lines 299 and 302 in `autoload/vimtex/complete.vim`:
> 
> https://github.com/lervag/vimtex/blob/22d9ae5265df069cb68310449a7218b836875c5b/autoload/vimtex/complete.vim#L297-L305
> 
> They apparently check if a user tries to complete a command in a math zone and filter candidates accordingly.

Ah, yes. I've comletely forgotten about that!

> So I miss something which filters based on `tikz` environment (and others). The code could look like: …

Yes, I guess this was something I was thinking about back then.

> Is it clearer now what I am talking about?

Yes, definitely.

> Since so much work has been done for the `.cwl` files in TeXstudio, I am wondering whether vimtex omnicompletion could be extended to support cwl files.

Well, to be honest, this feels like a big task, and one where I'm not immediately sure where to start. And I'm sorry to say I don't think I have the time to work on this.

As you noticed, the completion works, but it is not context dependent. So, clearly, it could be improved, but it is also better than nothing.

I would not mind considering PR's, of course, but I don't think I'll pursue this personally.

kiryph:
> Well, to be honest, this feels like a big task, and one where I'm not immediately sure where to start. And I'm sorry to say I don't think I have the time to work on this.

I agree that this is probably a big task.

Right now I think delegating this to a snippet plugin like LuaSnip is probably the best option without putting too much burden on VimTeX.

LuaSnip has already a wiki entry how to define snippets using conditions based on information from vimtex: `in_mathzone`, `tikz`, `in_beamer`, `in_siunitx`:

["Context-aware" Snippets]( https://github.com/L3MON4D3/LuaSnip/wiki/Cool-Snippets#context-aware-snippets)

So defining the completion candidates can be left to the user.

However, I am not sure how to check if a package was loaded. I will raise a separate issue/discussion for this.

Nevertheless in the long run I see value in collections of snippets or support existing cwl files since most people do the same thing.

A few references, if someone wants to take on the cwl path, e.g. a converter from cwl format to LuaSnip format:

1. [CWL Specification](https://texstudio-org.github.io/background.html#description-of-the-cwl-format)
2. [CWL Files in TeXstudio](https://github.com/texstudio-org/texstudio/tree/master/completion)
3. The [LaTeX Workshop Plugin](https://github.com/James-Yu/LaTeX-Workshop) for VS Code repurposes `.cwl` files for its own completion:
  > If you write your own package along with the corresponding .cwl file, you can use the Typescript script [parse-cwl.ts](https://github.com/James-Yu/LaTeX-Workshop/blob/master/dev/parse-cwl.ts). For details on how to run this script, please read https://github.com/James-Yu/LaTeX-Workshop/tree/master/dev#parse-cwlts . Place the generated <mypackage>.json file in a folder defined in [latex-workshop.intellisense.package.dirs](https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopintellisensepackagedirs). Note it only works when [latex-workshop.intellisense.package.enabled](https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#latex-workshopintellisensepackageenabled) is set to true and you have imported this package in LaTeX, i.e., \usepackage{mypackage}.
  https://github.com/James-Yu/LaTeX-Workshop/wiki/Intellisense#commands-starting-with-  

4. The [LaTeXTools](https://github.com/SublimeText/LaTeXTools) Plugin for Sublime Text also uses cwl files: https://github.com/SublimeText/LaTeXTools/tree/master/cwl

Alternatively, I think a `luasnip-tikz-snippets` collection as a sensible project similar to the ones for math by [iurimateus](https://github.com/iurimateus/luasnip-latex-snippets.nvim) or by [evesdropper](https://github.com/evesdropper/luasnip-latex-snippets.nvim).

Issue #3180, opened 2025-06-10
Feature Request: b:vimtex with package and documentclass options at load-time


kiryph:
**Is your feature request related to a problem? Please describe it.**
I want to create a LuaSnip snippet which takes into account package and documentclass options as given on load-time.

This is similar to the functionality of the completion in TeXstudio using `.cwl` files with `#ifOption:<option>` as described here: https://texstudio-org.github.io/background.html#description-of-the-cwl-format

**Describe the solution you'd like**
The buffer-local variable `b:vimtex` should also have documentclass and package options of the current document.  So the document parser and related methods like `VimTeXInfo` must be extended.

Afaik, there are two types of options, `key-only` or `key=value` and they can be written on a single line or on multiple lines. But they are always enclosed between square brackets between documentclass/usepackage/requirepackage and `{name}`.

Example file
```tex
\RequirePackage[debrief]{silence}
\documentclass[draft]{article}

\usepackage[rgb]{graphicx}
\usepackage[bibencoding=utf8]{biblatex} 

\begin{document}

\end{document}
```

For the packages `:lua vim.print(vim.b.vimtex.packages)` could give something like
```lua
{
  biblatex = { bibencoding = 'utf8' },
  graphicx = { rgb = nil },
  keyval = vim.empty_dict(),
}
```

I know that many packages have additional setup commands which are often equivalent to package options. Examples are 
```tex
\hypersetup{options} % package hyperref
\mmzset{options} % package memoize
% ...
```

I consider this outside the scope of this feature request.
Many package and documentclass options are given directly and sometimes are even required on load time (see e.g. biblatex). So this is already helpful. But this can mean the option could be altered after loading the package and hence the state depends on the line in the tex file.

**Describe alternatives you've considered**
None that I am aware of.

lervag:
Thanks so much for raising the issue _and_ fixing this yourself! <3

lervag:
I consider this issue resolved by #3181 which was just merged.

PR #3181, opened 2025-06-11
Parse options for document class and packages

kiryph:
This fixes also an issue for following multi-line notation

```tex
\usepackage{
  amsmath,
  tikz
}
```

Related issues, discussions and PRs:

- https://github.com/lervag/vimtex/issues/3180
- https://github.com/lervag/vimtex/pull/3175 (refers to question 2 in the first comment)
- https://github.com/lervag/vimtex/discussions/3178 (now you can take options into account)

### DIFF
--- a/autoload/vimtex/state/class.vim
+++ b/autoload/vimtex/state/class.vim
@@ -201,16 +201,44 @@ endfunction
 
 " }}}1
 function! s:parse_packages(preamble) abort " {{{1
-  let l:usepackages = filter(copy(a:preamble),
-        \ 'v:val =~# ''\v%(usep|RequireP)ackage''')
-  let l:pat = g:vimtex#re#not_comment . g:vimtex#re#not_bslash
-      \ . '\v\\%(usep|RequireP)ackage\s*%(\[[^[\]]*\])?\s*\{\s*\zs%([^{}]+)\ze\s*\}'
-  call map(l:usepackages, {_, x -> split(matchstr(x, l:pat), '\s*,\s*')})
+  " Remove EOL comments and then join
+  let l:preamble_joined = join(map(copy(a:preamble),
+        \ {_, x -> split(x..' ', '%')[0]}), '')
 
+  let l:pat = g:vimtex#re#not_comment . g:vimtex#re#not_bslash
+      \ . '\v\\%(usep|RequireP)ackage\s*%(\[([^[\]]*)\])?\s*\{\s*\zs%([^{}]+\S)\ze\s*\}'
+  " Regex:
+  " - Match contains package name(s)
+  " - First submatch contains package options
   let l:parsed = {}
-  for l:packages in l:usepackages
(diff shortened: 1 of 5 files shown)


lervag:
> Thanks for merging this addition.

My pleasure - I was very glad to see the contribution as I was not immediately sure how to solve #3180. I think your solution looks good and should work in most cases!

> I actually like your adjustments and had similarly changes considered: introduction of a shared function to parse options (documentclass or packages) and also to setup a `preamble_joined` string before doing all the preamble parsing.
> 
> I often try to narrow down the modifications for PRs. But I think a sensible refactoring in the context of a PR which introduces more changes can be better.

Glad you like the adjustments. I have a pragmatic view on things like this. I could have merged and then applied the changes, and I could have made comments and asked you to make those changes. Instead, I just made them myself on top before merging because I felt like it. It also helps me to understand your contribution better, which I think is a useful thing in itself.

kiryph:
@ChillyWall Can you post the preamble of the document you are opening?

kiryph:
The line in your preamble causing the error is:

```tex
\usepackage[font={stretch=1.2}, textfont={color=caption-color}, position=top, skip=4mm, labelfont=bf, singlelinecheck=false, justification=raggedright]{caption}
```

The current package option parser does not understand nested options in curly braces as possible for the LaTeX package *caption*.

Another example with two nested options:
```tex
\usepackage[font={small,stretch=0.80}]{caption}
```

I do not know if there are LaTeX packages which allow even more deeply nested options or if it is sufficient to extend the parser to understand one level of nested options as for *caption*.

ChillyWall:
@kiryph Thanks. After removing the options, vimtex works again.

lervag:
> This commit breaks my vimtex. When opening a tex file, I would get this error. It seems that the error is caused by the new parse_packages function.

Thanks for reporting this! I just pushed a minor fix; it doesn't actually fix the parsing problem, but it avoids the error and instead silently ignores the problem.

@kiryph This is a classical example of how regexes are not really ideal for this kind of parsing. Now, we _could_ work on a better fix here with more complex regexes, but I think perhaps it might be better to consider a more robust approach.

I believe we might want to do something like what I'm doing here:

https://github.com/lervag/vimtex/blob/8f8a5176b33856cf4dba9ae59f827b8deecc44c9/autoload/vimtex/parser/tex.vim#L140-L146

Now, it's not that easy, because we need this more robust parsing both to find the full option string, but also to parse _within_ the string. Thus, the outer `split(..., ",")` is not really good anymore - instead, we need to seek through the string allow braced values that themselves may contain commas.

I wouldn't say it is very _hard_, but it is somewhat meticulous.

And I think the right way forward is to start generalizing a set of api functions for parsing things from an input string. There are already quite a few such functions, but it's a little bit messy. :\

lervag:
The `vimtex#util#texsplit` function is clearly relevant here, though.

PR #3182, opened 2025-06-12
Fix test-tkz-euclide.vim

kiryph:


test/test-completion-commands/test-tkz-euclide.vim
```diff
@@ -1,6 +1,5 @@
 set nocompatible
-set runtimepath^=~/.local/plugged/vimtex
-set runtimepath+=~/.local/plugged/vimtex/after
+let &rtp = '../..,' . &rtp
```
lervag, line 2: I much prefer the `set runtimepath^=` syntax, including the long-form option name. Thus I propose instead to change it to this:

```vim
set runtimepath^=../..
```

kiryph: I have updated the PR accordingly with a force-push.

PR #3184, opened 2025-06-15
fix: E688 with option with empty value

tomtomjhj:
Problem:
If a file contains a line like this
```
\usepackage[key=]{package}
```
option parsing raises E688: More targets than List items.

Solution:
Use {keepempty} argument.


lervag:
Thanks!

PR #3194, opened 2025-07-27
Fix indentation for comments

patnr:
As indicated by this [comment](https://github.com/lervag/vimtex/blob/7f2633027c8f496a85284de0c11aa32f1e07e049/indent/tex.vim#L54) the intent is to apply the indent of the preceding line, not the current.

Without this patch, the code
```tex
        \begin{equation}
                  a = b \\
                  % comment
                  c = d
        \end{equation}
```
Gets indented (for example with `vie=`) to
```tex
        \begin{equation}
          a = b \\
                  % comment
          c = d
        \end{equation}
```

With the patch I get

```tex
        \begin{equation}
          a = b \\
          % comment
          c = d
        \end{equation}
```

### DIFF
--- a/indent/tex.vim
+++ b/indent/tex.vim
@@ -53,7 +53,7 @@ function! VimtexIndent(lnum) abort " {{{1
 
   " Use previous indentation for comments
   if l:line =~# '^\s*%'
-    return indent(a:lnum)
+    return indent(l:prev_lnum)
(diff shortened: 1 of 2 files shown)


lervag:
Sorry about the delay - I've been on vacations.

This looks like a very good catch, thanks!

Issue #365, opened 2016-02-15
Make a/ie linewise


Konfekt:
The Environment Text object should be linewise as Operator map, That is, :normal Vae<Cr>. For example, When  yanking , yae, this allows for convenient pasting afterwards.

lervag:
Actually, I would claim that it is not this easy. There are several scenarious in which I can see `begin/end` on the same line, for instance:

```tex
\[ a = \begin{bmatrix} 1 \\ 2 \end{bmatrix} \]
```

Here one would not want the operator to be linewise. Also, when we do `cie`, it is much better to preserver the indentation, which is impossible with linewise visual mode.

I agree that in very many cases we write the `begin/end` blocks very structured and on separate lines, and that in very many such cases, the linewise operator would be useful. I think an optimal solution is to have the text object be aware of its operator, but I don't know how.

lervag:
FYI, I've opened a [question on vi.stackexchange.com](http://vi.stackexchange.com/q/6500/21).

Konfekt:
Yeah true, I'd like to withdraw this issue, and in any case only let it apply to yanking. 

If people put `\begin{...} ... \end{...}` environments inline, then `vimtex` should check for this case leave when yanking the selection characterwise. Here, environments are always linewise, and then `yie` and pasting ideally should also be.

Konfekt:
I answered your question by the way. There are a couple of maps using this construction in `.vimrc`. Others:
```vim
onoremap        <SID>(esc) <esc>
omap     <expr> C          '<SID>(esc)' . v:operator . v:count1 . (v:operator ==# 'd' ? '<Plug>(textobj-comment-big-a)' : '<Plug>(textobj-comment-i)')

omap     <expr> <CR>       '<SID>(esc)' . v:operator . v:count1 . (v:operator ==# 'y' ? '<Plug>(textobj-entire-i)' : '<Plug>(textobj-entire-a)')
```

Konfekt:
Thinking about: it, by the operator mapping in http://vi.stackexchange.com/q/6500/21, check if the operator is `y` and then, if `\begin{...}` and `\end{...}` stand alone on separate lines, set the yank register to linewise by 
```vim
  call setreg(a:regname, getreg(a:regname), "l")
```

lervag:
I've begun working on this. I've pushed a commit that actually breaks things now (in that it now uses linewise almost randomly), but it is the first step in fixing things. I'll try to get things sorted out in not too long.

[...]

PR #3195, opened 2025-07-27
Fix: 'a'-kind textobjs never operated linewise

patnr:
The intention in
https://github.com/lervag/vimtex/issues/365#issuecomment-193966866 is that non-inline regions should be linewise.
So I don't know why `l:is_inline` is a requirement on top of `l:linewise`


patnr:
Closing, as it seems that I'm just confused. But I would comment that (without this PR) I never seem to get linewise operator behaviour, even though the environments are cleanly separated into lines, and `g:vimtex_text_obj_linewise_operators = ['d', 'y']`

Issue #2835, opened 2023-11-23
Some issues with TeXShop previewer


griffinwxk:
### Description

I've been trying out various previewers recently due to dissatisfaction of them (unrelated to vimtex itself):
-  Skim has to be launched by neovim for inverse searching to work, but since I use it as my main pdf viewer for papers it's not convenient.
- zathura is great overall as a previewer, but trackpad scrolling and keyboard scrolling have vastly different speed so no matter what it's always janky.
- sioyek is not polished enough, specifically its font rendering is buggy for multiple displays with different dpi

So it leads to me trying out configuring TeXShop.  As it turns out, vimtex's setup seems not working for me. I dug into the source and extracted the AppleScript code and tested it. Maybe it's due to OS version (I'm on Sonoma), I'm very surprised to find out that it has syntax errors.

I modified the code and finally got it working and it seems good enough. During my testing I also found out some shortcomings of the original code:

- Notably it will only do forward searching for the "front most document" opened by TeXShop. So if I have multiple projects open, only one of them at a time can have forward searching. I fixed it on my machine by checking the corresponding pdf everytime instead of using the front most one.
- When synctex is set, the current code also opens tex file in TeXShop and launches TeXShop's compiler for some reason, which seems completely unnecessary? I deleted the line and it seemed perfectly fine.
- Related to above, when synctex is set, unlike other viewing method like Skim, which simply appends some command to include synctex, the code for TeXShop overwrites it with new command. So the behavior is weird: no synctex => launch previewer normally; yes synctex => launch tex file and compile. I don't know why it was coded this way.

The reason I didn't create a PR is that I don't know if the original author had other intentions and if there is compatibility issues across macos versions. And I'm also not quite familiar with PRs (only done a very tiny one before). If anyone can confirm this issue I'd be happy to submit a PR later.

### Steps to reproduce

Set `g:vimtex_view_method` to `"texshop"`, and launch viewer.

### Expected behavior

TeXShop opens with the pdf only, and forward/inverse searching working.

### Actual behavior

As is, nothing shows up. I described several other issues after testing the code in the description. 

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
It's irrelevant as I was able to pin the error to the AppleScript code in texshop.vim
```

lervag:
[...]

PR #3200, opened 2025-08-10
implement #2835

leo-leesco:
I implemented the modifications suggested by [griffinwxk](https://github.com/griffinwxk)
Nothing much, but it seems that this was what was broken.

There is still some work to do (at least on my computer, forward search is a bit strange, I have two previews open on startup and SyncTex only seems to work on one of the previews)

Thought I should also mention that, inside TexShop, `Settings > PDF Preview > External editor > Automatic preview update` should be toggled on

Thanks for the great plugin !

### DIFF
--- a/autoload/vimtex/view/texshop.vim
+++ b/autoload/vimtex/view/texshop.vim
@@ -52,7 +52,7 @@ function! s:make_cmd_view(outfile, open, sync) abort " {{{1
         \ '-e ''tell application "TeXShop"''',
         \ '-e ''try''',
         \ '-e ''set theDocs to get documents whose path is thePath''',
-        \ '-e ''if (count of theDocs) > 0 then revert theDocs''',
+        \ '-e ''if (count of theDocs) > 0 then revert(theDocs)''',
         \ '-e ''end try''',
         \ a:open ? '-e ''open theFile''' : '',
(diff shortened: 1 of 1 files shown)


lervag:
> I implemented the modifications suggested by [griffinwxk](https://github.com/griffinwxk). Nothing much, but it seems that this was what was broken.
> 
> There is still some work to do (at least on my computer, forward search is a bit strange, I have two previews open on startup and SyncTex only seems to work on one of the previews). Thought I should also mention that, inside TexShop, `Settings > PDF Preview > External editor > Automatic preview update` should be toggled on.

Great, thanks! As I don't use neither MacOS nor TeXShop I will have to take your word for it that this indeed works as expected. I have no means of testing it myself.

So, let me know when you want me to consider it for merging.

> Thanks for the great plugin!

Glad you like it!

> Maybe another comment, I don't really know where to put that, but a "correct" configuration for using TexShop could be: …

Well, I _really_ don't like a suggestion that asks users to run commands with sudo. I think instead we should ask people to create a local script e.g. under `$HOME/.local/bin/othereditor` and to make it executable with `chmod +x`. We could write about this in the docs under `:help vimtex-view-texshop`. There we could also suggest that people do the `defaults write ...` commands.

Issue #3224, opened 2025-11-02
mkdir tries to create current directory when TeX file in parent and .latexmkrc specifies "build"


ordinary-slim:
### Description

With this project structure

```
.
├── metadata.tex
└── subfolder
    ├── .latexmkrc
    └── main.tex
```

and .latexmkrc specifying

```perl
$out_dir = 'build';
$aux_dir = 'build';
```

VimtexCompile errors in its first run, trying to create the current folder when `build` is still absent:

```
VimTeX: Creating directorie(s):
        * [...]/subfolder/build/..
Error detected while processing function vimtex#compiler#compile[7]..vimtex#compiler#start[18]..38[3]..56[1]..32:
line   29:
E739: Cannot create directory [...]/subfolder/build/..: file already exists
```

This boils down to `call mkdir("./build/..", "p")`, and `build` has not been created yet. This errors despite the "p" flag. After the first run, the project structure is:

```
.
├── metadata.tex
└── subfolder
    ├── build
    ├── .latexmkrc
    └── main.tex
```

and consequent runs don't error.

### Steps to reproduce

1. [https://github.com/ordinary-slim/vimtex-mkdir-bug](https://github.com/ordinary-slim/vimtex-mkdir-bug)
2. I believe this will error regardless of the config. Open `main.tex` in `subfolder` and run `VimtexCompile

### Expected behavior

Vimtex should not try to mkdir the current folder, and not like this: `./build/..`, where `build` is still absent.

### Actual behavior

Vimtex runs `mkdir("./build/..", "p")`, where `build` is still absent, and errors in its first run.

### Do you use a latexmkrc file?

Yes

### VimtexInfo

```yaml
System info:
  OS: Linux 6.17.5-arch1-1 (I use arch btw)
  LaTeX version: pdfTeX 3.141592653-2.6-1.40.28 (TeX Live 2025)
  Vim version: NVIM v0.11.4
  Has clientserver: true
  Servername: /run/user/1000/nvim.134803.0
  $PATH:
    - /bin
    - /home/mehdi/.local/bin
    - /home/mehdi/.local/share/nvim/mason/bin
    - /sbin
    - /usr/bin
    - /usr/bin/core_perl
    - /usr/bin/site_perl
    - /usr/bin/vendor_perl
    - /usr/local/bin
    - /usr/local/sbin
    - /usr/local/texlive/2025/bin/x86_64-linux
    - /usr/sbin

VimTeX project: main
  base: main.tex
[...]

PR #3225, opened 2025-11-02
dont mkdir any path that starts with '..' (instead of '../' previously)

ordinary-slim:
Addresses #3224 . Tests (ran with `make` in `tests` folder) run.


lervag:
Thanks, I appreciate the proposed fix. I'll probably accept it, but please consider my questions/comments in #3224 first.

ordinary-slim:
Your solution is more robust, thanks for letting me commit it.

lervag:
No problem, and thanks again for your contribution! I'm always happy when people (like you) both detect a problem and attempt to fix it themselves, and I'm glad to assist and let you complete the final stages. :)

PR #3243, opened 2025-12-17
Add macrocode as special case for syntax highlighting

Skillmon:
Inside a DTX file the `macrocode` environment has a very special meaning as it starts and ends the portions of the file which will be added to the exported file.

For that reason this PR adds a highlighting group for `%    \begin{macrocode}` (and `\end`) and links it to `Special`.


lervag:
Thanks! Can you give an example `.dtx` file for this as well?

lervag:
Ah, sorry, I need it for testing. I like to have a simple test file for more or less any feature. I can add the file myself, I just need to have some example of a relevant dtx file that uses the macrocode environment.

Issue #3262, opened 2026-04-07
[Feature Request] Texpresso


keikun555:
**Is your feature request related to a problem? Please describe it.**

When compiling super large documents (50+ page latex documents, long beamer presentations, etc), the compilation time gets super long (1+ minutes).

There is already a PR that adds texpresso (see https://github.com/lervag/vimtex/pull/2975 and https://github.com/lervag/vimtex/pull/2975#issuecomment-2212503633 for why this can help with compilation time), but it was closed due to inactivity. It seems close to completion, however, and this Issue aims to track getting this PR to merge.

**Describe the solution you'd like**

It would be nice if we could reopen https://github.com/lervag/vimtex/pull/2975 and get this merged. I unfortunately don't know enough vim script to get this to completion, but I feel like this would be an awesome addition to vimtex.

**Describe alternatives you've considered**

There's also https://github.com/frabjous/knap , but it seems to be orthogonal since it's packaged with a viewer itself, which seems to encroach on what vimtex already does.

keikun555:
I also looked up incremental compilation of tex files. The idea is that if we only change a few lines in a document, we would only recompile the changed bits and not need to recompile the entire document. There doesn't seem to be a tex compiler that supports this, only fully-packaged editors (e.g. [texifier](https://www.texifier.com/)).

lervag:
> When compiling super large documents (50+ page latex documents, long beamer presentations, etc), the compilation time gets super long (1+ minutes).

Yes, long compilation times for large documents is a known problem.

> There is already a PR that adds texpresso (see [#2975](https://github.com/lervag/vimtex/pull/2975) and [#2975 (comment)](https://github.com/lervag/vimtex/pull/2975#issuecomment-2212503633) for why this can help with compilation time), but it was closed due to inactivity.

> It seems close to completion, however, and this Issue aims to track getting this PR to merge.

Well, yes, but the person who was contributing no longer had time to work on it. But I agree, it did seem quite close to completion.

> It would be nice if we could reopen [#2975](https://github.com/lervag/vimtex/pull/2975) and get this merged. I unfortunately don't know enough vim script to get this to completion, but I feel like this would be an awesome addition to vimtex.

Reopening the PR is not really the best approach, because we would need someone to finish the work. I'm sorry to say that I don't have much time to spend on maintenance these days, so I would rather prioritize answering issues and fixing simple things.

That is, I would very much prefer if someone (perhaps you) could do the main work here. I don't mind helping and guiding!

[...]

PR #3267, opened 2026-05-14
feat(compiler): add TeXpresso integration

SeniorMars:
Add a VimTeX compiler backend for TeXpresso, a live-rendering LaTeX tool that updates the preview incrementally as the buffer changes. The backend communicates with TeXpresso through its JSON protocol over stdio and supports SyncTeX forward/inverse search, quickfix integration, and live buffer synchronization. Neovim buffer synchronization is handled through lua/vimtex/compiler/texpresso.lua.

This revives the work from PR #2975 and addresses the main review comments from that discussion:

- Move the Neovim buffer-attach Lua code to lua/vimtex/compiler/texpresso.lua, with type annotations and a defensive check for compiler state during shutdown or reload.
- Add documentation in doc/vimtex.txt modeled on the existing compiler backend sections, including a Requirements block and option docs.
- Internalize the required -json and -lines flags so users cannot accidentally break the protocol by editing g:vimtex_compiler_texpresso.
- Normalize passed_options spacing in __build_cmd with trim() so the function is robust regardless of whether the caller includes a leading space.
- Add a stdin_pipe flag to the compiler template so that only TeXpresso enables stdin piping for its job, rather than changing the default for all continuous compiler backends.
- Register texpresso in the valid compiler methods list in compiler.vim.
- Add tests (test/test-compiler-texpresso/):
  - Unit tests (`make` / `make vim`) verify stdin_pipe, continuous mode, __build_cmd output, and passed_options spacing without requiring TeXpresso to be installed. These run under both Neovim and Vim.
  - Integration smoke test (`make integration`) starts a real TeXpresso process under Neovim, sends a reload message, and verifies that it stops cleanly.

Note: I'm a bit unfamiliar with the vim codebase, so additional tests may be needed; however, neovim has been tested.

Co-Authored-By: Tsung-Ju Chiang <71379180+tsung-ju@users.noreply.github.com>


doc/vimtex.txt
```diff
@@ -254,7 +255,7 @@ Vim configuration ~
 
 Compiler backend ~
 
-  VimTeX uses `latexmk`, `latexrun`, `tectonic` or `arara` to compile the LaTeX document.
+  VimTeX uses `latexmk`, `latexrun`, `tectonic`, `arara` or `texpresso` to compile the LaTeX document.
```
lervag: Please restrict to 79 characters per row in this file.

autoload/vimtex/compiler/texpresso.vim
```diff
@@ -0,0 +1,226 @@
...
+  call call(a:super, a:000, self)
+
+  augroup vimtex_compiler_texpresso
+    autocmd! * <buffer>
+    autocmd CursorMoved <buffer> call b:vimtex.compiler.texpresso_synctex_forward()
+    autocmd ColorScheme <buffer> call b:vimtex.compiler.texpresso_theme()
+  augroup END
+
+  if has('nvim')
+    let self.nvim_detach = s:nvim_attach()
```
lervag: I'm curious, why don't we inline the `luaeval` here instead of running it in line 44 at module load time?

lervag (commented):
Thanks! This seems to be a good PR, and I only have a couple of minor code comments.

> Note: I'm a bit unfamiliar with the vim codebase, so additional tests may be needed; however, neovim has been tested.

Thanks! I can probably test this with Vim when I get the time. I'm assuming that it should be easy to install texpresso on Arch Linux (which seems to be the case).

Is there anything specific you would advice me to test before accepting this?

lervag:
If the PR is ever merged, it should resolve #3262.

autoload/vimtex/compiler/texpresso.vim
SeniorMars: Ty.

lervag:
Thanks for the updates! You're a person of few words; should I consider the PR ready for merge from your side?

SeniorMars:
Omg, sorry, for some reason I didn't read the previous comment. 

> Is there anything specific you would advice me to test before accepting this?

It would do wonders for me if you can confirm that vim works fine with this compiler. I tested it a bit, but honestly, I feel like you would have more experience on this compiler backend stuff. 

Other than that, I do think it's ready to merge.

lervag:
Great! I'll test it with Vim as soon as I get the time, then merge this after (unless I find a surprising glaring issue). I assume you agree that this resolves #3262?

SeniorMars:
Yes! Though I only tested it on one large file. I think it did it's job!

lervag:
I've tested with Vim, and it seems to work fine. Thanks for your contribution!

PR #3268, opened 2026-06-13
doc: Correct to encoding name utf-8

attjo:
According to [vimhelp](https://vimhelp.org/mbyte.txt.html#encoding-names) the encoding format is `utf-8`.
A suggestion is to only set the encoding when working on tex files.


lervag (approved):

lervag:
Thanks!

Issue #2203, opened 2021-10-11
Issue with window swallow when running `(vimtex-view)`


paniash:
### Description

Previously mentioned in #1768, the recent commit (not sure specifically) made changes to the `<plug>(vimtex-view)` function which now swallows windows by default. This was not the case before. 

I am currently running vimtex from commit 45d59be8017efabd7fe0249bc3d666e71f38ef6c, which does not have this issue.

### Expected behavior

Running `<plug>(vimtex-view)` spawns a new window for the pdf viewer.

### Actual behavior

The pdf-viewer gets swallowed by the terminal when running `<plug>(vimtex-view)`.

### Do you use a latexmkrc file?

No

### VimtexInfo

```yaml
System info
  OS: Linux 5.10.70-1-lts
  Vim version: NVIM v0.6.0-dev+449-gf4359b5db
  Has clientserver: true
  Servername: /tmp/nvimT2RZgx/0

VimTeX project: main
  base: main.tex
  root: /home/ashish/documents/sixth-sem/project/bscthesis
  tex: /home/ashish/documents/sixth-sem/project/bscthesis/main.tex
  main parser: current file verified
  document class: thesis
  packages: amsmath biblatex circuitikz cleveref csquotes fontenc graphicx hyperref hyphenat inputenc microtype pgf tgschola tikz xcolor
  source files:
    main.tex
    abstract.tex
    acknowledgements.tex
    chapter1.tex
    chapter2.tex
    chapter3.tex
    chapter4.tex
    chapter5.tex
    nonlocal-with-field.pgf
    chapter6.tex
  compiler: latexmk
    engine: -pdf
    options:
      -verbose
      -file-line-error
      -synctex=1
      -interaction=nonstopmode
    callback: 1
    continuous: 0
    executable: latexmk
  viewer: Zathura
    xwin id: 0
  qf method: LaTeX logfile
```

paniash:
The commit 7d8d92ac295e816fdecccd23cb55766162ec5b22 introduced this swallowing change.

lervag:
Sorry, I meant `jobstart`, not `job_start` (the latter is the Vim equivalent, the former is the neovim version).

lervag:
No, because there is a good reason to make those changes. Reverting it only for you is not necessarily a good thing. So, it is more complicated.

Can you test if `jobstart('... &')` works?

lervag:
My pleasure :)

PR #3303, opened 2026-09-08
fix(view): run zathura with `vimtex#jobs#start`

rhelder:
VimTeX runs  zathura in the background by calling (ultimately) `system(<cmd> .. '&')`. Neovim documentation for `system()` discourages running commands in the background with `system()`, because `system()` cannot write to or read from backgrounded shell commands. The docs recommend using `jobstart()` instead. The same considerations probably apply to Vim as well.

In practice, this has caused me some scattered – and I suspect system-dependent – problems while using the `zahtura_simple` viewer, so I may not have an issue caused by this that you can reproduce. On my previous system, zathura would crash with a core dump whenever VimTeX tried to start it; back then, the code in this PR solved that problem. On my current system, I no longer had that problem, so I thought maybe I'd set this code aside. Then I found that if I set an autocommand to run `:VimtexView` after a successful compilation, such as the following,

```lua
vim.api.nvim_create_autocmd(
  'User', {
    group = augroup,
    pattern = 'VimtexEventCompileSuccess',
    callback = function () vim.cmd.VimtexView() end,
  }
)
```
that VimTeX would open two zathura instances. Now, this is expected behavior with `zathura_simple`, but interestingly, with this PR, I don't get multiple zathura instances.

I can't think of anything intrinsic to this code that would prevent multiple zathura instances from being opened, so I'm not suggesting that this is a fix for that (expected) problem. There just happens to be something about this code that interacts with this version of zathura, on my window manager, etc. etc. that's resulting in behavior I like better. To be honest, I wouldn't pretend that this code is a solution for any particular problem! All I have to say in the code's favor is that it's in line with what the Neovim documentation recommends (although I assume the same considerations apply to Vim), and that the code has cleared up some (sometimes substantial) system-dependent problems for me at different times.

I include the same explanatory comment as you have in the Sioyek viewer code, and I set `detached` to `true` as you do in the Sioyek viewer code. I do this for consistency, not because I favor setting `detached` to `true` myself – I actually would prefer not to, but it's your call!

This partially reverts 97cb7fa52412940c8696a9d00425a2c87b04b7e8, which I know is in response to #2203. I don't find the issues raised by the author of that issue to be particularly valid. As best I can understand it, if you've enabled window swallowing for your terminal emulator, you are essentially using your terminal as an application launcher. At that point, I don't think it's particularly fair to complain that your application launcher doesn't stick around to edit documents after it's launched the application you asked it to launch. What you want for VimTeX is precisely a terminal emulator with window swallowing disabled (whether at the application level or WM/DE level), because you want both the terminal and the PDF viewer to be open... So that user was just using the wrong tool for the job. So I don't think their issue is a strong objection against reverting back to using `vimtex#jobs#start()`.

I'm not sure if I've offered strong reasons to merge this, and I can imagine some reasons not to merge it, but I've been sitting on this code for a while one way or another, so you can have it if you want it! 😅

Thanks very much for your work!


lervag:
Hi! First, I'm always glad to consider PR's, so thanks!

> Thanks very much for your work!

Thanks, glad to hear it! <3

> VimTeX runs zathura in the background by calling (ultimately) `system(<cmd> .. '&')`.

Yes, that's right.

> Neovim documentation for `system()` discourages running commands in the background with `system()`, because `system()` cannot write to or read from backgrounded shell commands.

Yes, but in this case, we don't need or use that at all.

> In practice, this has caused me some scattered – and I suspect system-dependent – problems while using the `zahtura_simple` viewer, so I may not have an issue caused by this that you can reproduce.

Hard to say, you never opened an issue report 😅

> I can't think of anything intrinsic to this code that would prevent multiple zathura instances from being opened, so I'm not suggesting that this is a fix for that (expected) problem.

I think I can explain this one, and I agree that it is not intrinsic.

For `zathura_simple` we always pass `--synctex-forward` when starting. Per `man zathura`, that option looks for a running instance with the given document open and forwards the position to it; a new instance is only launched if none is found. That D-Bus lookup is what replaces `xdotool` for the simple variant. So a second `_start` only gives you a second window if it happens _before_ the first instance has registered itself on the bus.

That's a race, and the patch changes its timing. `#run` with `&` returns in ~2 ms, while `#start` sleeps 100 ms before returning in order to ensure the job was properly started. On my machine, Zathura registers on D-Bus after ~120 ms. Starting two instances with a controlled gap between them:

```
gap   2 ms  ->  2, 2, 2, 2 windows
gap 100 ms  ->  1, 1, 2, 2 windows
gap 300 ms  ->  1, 1, 1, 1 windows
```

So the patch doesn't prevent the second instance; it just widens the gap enough that you usually win the race. That's why it helps you, why it is machine dependent, and why you couldn't find anything intrinsic. It's an accidental fix, which is not really a fix at all.

By the way: `g:vimtex_view_automatic` is enabled by default, so your `VimtexEventCompileSuccess` autocommand duplicates what VimTeX already does after a successful compilation. That is likely what produces two `_start` calls in quick succession in the first place. If you drop the autocommand, the duplicate instances should go away regardless of this patch.

> This partially reverts [97cb7fa](https://github.com/lervag/vimtex/commit/97cb7fa52412940c8696a9d00425a2c87b04b7e8), which I know is in response to #2203.

Right, so it seems to me that you are asking me to accept a patch to fix an issue that is not quite reproducible in a way that _reverts_ a patch that fixed a reproducible issue. That doesn't feel quite right.

> I don't find the issues raised by the author of that issue to be particularly valid. As best I can understand it, if you've enabled window swallowing for your terminal emulator, you are essentially using your terminal as an application launcher. At that point, I don't think it's particularly fair to complain that your application launcher doesn't stick around to edit documents after it's launched the application you asked it to launch. What you want for VimTeX is precisely a terminal emulator with window swallowing disabled (whether at the application level or WM/DE level), because you want both the terminal and the PDF viewer to be open... So that user was just using the wrong tool for the job. So I don't think their issue is a strong objection against reverting back to using `vimtex#jobs#start()`.

I'd like to separate two things here: whether that workflow is sensible, and whether this patch actually avoids it.

On the first point: you may well be right that window swallowing plus VimTeX is an odd combination. But I don't think it's my place to tell people their WM configuration is wrong. And I don't have to decide, because `&` works for both groups of users while `#start` works for only one of them. It's not really a trade-off between your preference and theirs; one option is simply more compatible.

For the second point: the patch doesn't actually avoid the swallowing, even with `detached` set. `detached` does not detach the process from the process _tree_. In Neovim it calls `setsid()`, and in Vim it only clears `stoponexit` so that we don't kill the job when quitting. Either way the viewer survives Vim exiting, but neither reparents it, so Vim stays its parent for the whole session. Using `sleep` in place of zathura:

```
nvim: jobs#run + '&'        viewer ppid=1       <- orphaned
nvim: jobs#start detached   viewer ppid=<nvim>  <- still our child
vim:  jobs#run + '&'        viewer ppid=1       <- orphaned
vim:  jobs#start detached   viewer ppid=<vim>   <- still our child
```

Window swallowing works by walking up the parent chain from the process that owns the newly mapped window. With `&`, `sh` exits within a couple of milliseconds and the viewer is reparented to init long before zathura maps its window (~120 ms on my machine). With `#start` we are still its parent at that moment, so the swallow triggers.

So this isn't something I could fix by passing better options: to break the parent chain I have to background with `&`, and at that point the job handle refers to an `sh` that has already exited, which gives me nothing. Backgrounding and holding a useful job handle are mutually exclusive here.

> I include the same explanatory comment as you have in the Sioyek viewer code, and I set `detached` to `true` as you do in the Sioyek viewer code.

Fair. The inconsistency is real: Sioyek, MuPDF and the general viewer do use `#start`. The difference is only that nobody has reported a #2203-style problem for those. If someone does, I would expect to resolve it in the other direction rather than take the regression here for the sake of symmetry.

To be clear, I'm not dismissing your experience! Segfaulting zathura is not nothing. But if it shows up again, or you manage to pin down the scattered problems, please open an issue with steps to reproduce and I'll be glad to dig into it. I just don't want to merge a change to the process handling on the strength of an accidental 100 ms delay.

Thanks again for taking the time to write this up, and for the clear and honest description of what you did and didn't observe. That made it much easier to work out what is actually going on. As you probably understand, I'm leaning towards not merging this for the reasons given above.

rhelder:
> I think I can explain this one, and I agree that it is not intrinsic.

Thanks for your explanation of this – I found it very educational!

> It's an accidental fix, which is not really a fix at all.

Agreed (and I hope I didn't advertise it as a fix).

> By the way: g:vimtex_view_automatic is enabled by default, so your VimtexEventCompileSuccess autocommand duplicates what VimTeX already does after a successful compilation. That is likely what produces two _start calls in quick succession in the first place. If you drop the autocommand, the duplicate instances should go away regardless of this patch.

Yes, I am aware of this. I actually don't use this autocommand as part of my regular set-up – I only used it as a temporary workaround due to a bug in zathura, where the document position was reset to the top of the document on each re-compilation. This bug was very quickly fixed, so I only used this autocommand for a short time. I happened to notice during that time, when I reapplied this patch, that I no longer was seeing multiple zathura instances on the first compilation. You have very informatively explained why that turned out to be the case 😄.

> But I don't think it's my place to tell people their WM configuration is wrong.

Fair enough.

> For the second point: the patch doesn't actually avoid the swallowing

Right – I'm aware of this. That's the only reason I was offering my opinion about the validity of #2203, because I know that this patch would reintroduce the issues reported there.

> I would expect to resolve it in the other direction rather than take the regression here for the sake of symmetry.

Well, I wasn't really suggesting it for the sake of symmetry. I was suggesting it because it seemed more in line with the practices recommended by the docs (and that the practices recommended by the docs had some good results), and because think the concerns in #2203 are due to improper configuration on the user's side. I think you are very generous to say it's not your place to tell users that their WM configuration is wrong – but to me, it seems like the user wasf using a tool objectively meant for a different job.

> Hard to say, you never opened an issue report 😅

>To be clear, I'm not dismissing your experience! Segfaulting zathura is not nothing. But if it shows up again, or you manage to pin down the scattered problems, please open an issue with steps to reproduce and I'll be glad to dig into it. I just don't want to merge a change to the process handling on the strength of an accidental 100 ms delay.

Fair enough, and I wanted to open an issue, or submit this PR, at that time – but my old system was a beater PC with my first Arch Linux installation, so development moved fast. Once this patch was solving my issue, the issue went on the back-burner. Apologies for not reporting sooner while it was happening in real time.

> Thanks again for taking the time to write this up, and for the clear and honest description of what you did and didn't observe. That made it much easier to work out what is actually going on. As you probably understand, I'm leaning towards not merging this for the reasons given above.

Also fair enough. If you happened to agree that `system(<cmd> .. '&')` is not to be preferred, and that in hindsight #2203 wasn't worth introducing `system(<cmd> .. '&')`, maybe this could have been helpful – but putting myself in your shoes, I knew that it might not be worth it to merge something that is in some ways cosmetic. I will go ahead and close – thanks for your response.

lervag:
Cool; thanks again! Happy to have been able to shine some light on things.

If you do have similar issues in the future, feel free to raise them. And although I do prefer well laid out issues with minimal repros, I also don't mind if you raise an issue before you have all the details prepared.

In the meantime, I hope VimTeX is working well for you, and I'm sorry for the inconvenience of not merging this (even as I stand by the decision).

